PR #31, opened 2018-12-13
All MqttConnection operations return CompletableFutures

justinboswell:


src/main/java/software/amazon/awssdk/crt/mqtt/MqttConnection.java
```diff
@@ -127,130 +115,129 @@ public ConnectionState getState() {
...
+        CompletableFuture<Void> future = new CompletableFuture<>();
+        AsyncCallback connectAck = wrapVoidFuture(future);
         try {
             connectionState = ConnectionState.CONNECTING;
-            mqtt_connect(native_ptr(), clientId, cleanSession, keepAlive, connectAck);
+            mqttConnectionConnect(native_ptr(), clientId, cleanSession, keepAlive, connectAck);
         } catch (CrtRuntimeException ex) {
-            return false;
+            future.completeExceptionally(ex);
         }
```
graebm: I don't really know the right answer here, but this got me thinking...

This error didn't happen asynchronously, so should we re-throw it as an MqttException? Is that better than stuffing it into the Future that this function just created?

I notice that the `publish()` function will throw if there's an immediate error. So whichever approach we take, let's be consistent.

src/main/java/software/amazon/awssdk/crt/mqtt/QoS.java
```diff
@@ -0,0 +1,35 @@
+package software.amazon.awssdk.crt.mqtt;
+
+/**
+ * Quality of Service associated with a publish action or subscription [MQTT-4.3].
+  */
+public enum QoS {
```
graebm: Matthew suggested we name this "Qos" or "QualityOfService"

src/main/java/software/amazon/awssdk/crt/mqtt/MqttConnection.java
```diff
@@ -127,130 +115,129 @@ public ConnectionState getState() {
...
 
-    public boolean setWill(MqttMessage message, QOS qos, boolean retain) throws MqttException {
+    public boolean setWill(MqttMessage message, QoS qos, boolean retain) throws MqttException {
         if (native_ptr() == 0) {
             throw new MqttException("Invalid connection during setWill");
         }
 
         try {
-            return mqtt_set_will(native_ptr(), message.getTopic(), qos.getValue(), retain, message.getPayloadDirect());
+            return mqttConnectionSetWill(native_ptr(), message.getTopic(), qos.getValue(), retain, message.getPayloadDirect());
```
graebm: this just returns True/False for success/failure. It should probably be void/throw

src/main/java/software/amazon/awssdk/crt/mqtt/MqttConnection.java
```diff
@@ -127,130 +115,129 @@ public ConnectionState getState() {
...
         }
 
-        AsyncCallback subAck = wrapAck(ack);
+        CompletableFuture<Integer> future = new CompletableFuture<>();
+        AsyncCallback subAck = wrapPacketFuture(future);
         try {
-            return mqtt_subscribe(native_ptr(), topic, qos.getValue(), new MessageHandler(topic, handler), subAck);
+            int packetId = mqttConnectionSubscribe(native_ptr(), topic, qos.getValue(), new MessageHandler(topic, handler), subAck);
+            // When the future completes, complete the returned future with the packetId
+            return future.thenApply(unused -> packetId);
```
graebm: In the future, there might be more we want to convey. Ex: QOS granted by the server (might be lower than user asked for). And I have no idea how multi-subscribe should be handled, where SOME of the topics might have succeeded.

Anyway, you could change this when we get those changes from the underlying libraries

src/main/java/software/amazon/awssdk/crt/mqtt/QoS.java
justinboswell: Can do.

src/main/java/software/amazon/awssdk/crt/mqtt/MqttConnection.java
justinboswell: That's a good point. I'll make it consistent: If we get the exception in the immediate call, we'll throw. If not, we'll return it with the future.

src/main/java/software/amazon/awssdk/crt/mqtt/MqttConnection.java
justinboswell: ok.

src/main/java/software/amazon/awssdk/crt/mqtt/MqttConnection.java
justinboswell: yeah, in that case, the return type will be more than packetId. Problem for future us :)

src/main/java/software/amazon/awssdk/crt/CRT.java
```diff
@@ -155,5 +135,5 @@ private static void loadLibraryFromJar() {
         }
     }
 
-    private static native void aws_crt_init() throws CrtRuntimeException;
+    private static native void AwsCrtInit() throws CrtRuntimeException;
```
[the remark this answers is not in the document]
justinboswell: Good catch, fixed.

src/main/java/software/amazon/awssdk/crt/mqtt/MqttConnection.java
millems: It's actually more common (and expected) that a method returning CompletableFuture would never throw an exception, and exceptions would always be reported through the future.

src/test/java/software/amazon/awssdk/crt/test/IotServiceTest.java
```diff
@@ -18,14 +18,17 @@
 import org.junit.Test;
 import static org.junit.Assert.*;
 import software.amazon.awssdk.crt.*;
```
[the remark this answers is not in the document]
justinboswell: Done

src/main/java/software/amazon/awssdk/crt/mqtt/MqttConnection.java
dagnir: +1. Having more than one error path is confusing for the customer; see https://github.com/aws/aws-sdk-java-v2/issues/436

justinboswell: Ok. Easy fix.

PR #34, opened 2019-01-10
API changes required to match new MQTT

justinboswell:

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/mqtt/MqttConnection.java
+++ b/src/main/java/software/amazon/awssdk/crt/mqtt/MqttConnection.java
@@ -15,10 +15,10 @@
  */
 package software.amazon.awssdk.crt.mqtt;
 
-import software.amazon.awssdk.crt.io.TlsContext;
 import software.amazon.awssdk.crt.CrtRuntimeException;
 import software.amazon.awssdk.crt.CrtResource;
 import software.amazon.awssdk.crt.io.SocketOptions;
+import software.amazon.awssdk.crt.io.TlsContext;
 
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Consumer;
@@ -32,15 +32,16 @@ import java.io.Closeable;
  * MqttConnection represents a single connection from one MqttClient to an MQTT
  * service endpoint
  */
-public class MqttConnection extends CrtResource implements Closeable {
+public class MqttConnection extends CrtResource implements Closeable, MqttConnectionEvents {
     private final MqttClient client;
     private volatile ConnectionState connectionState = ConnectionState.DISCONNECTED;
+    private MqttConnectionEvents userConnectionCallbacks;
+    private AsyncCallback connectAck;
+    private AsyncCallback disconnectAck;
 
     public enum ConnectionState {
         DISCONNECTED, CONNECTING, CONNECTED, DISCONNECTING,
     }
-
-
     
     private class MessageHandler {
         String topic;
@@ -51,6 +52,7 @@ public class MqttConnection extends CrtResource implements Closeable {
             this.topic = topic;
         }
 
+        /* called from native when a message is delivered */
         void deliver(byte[] payload) {
             callback.accept(new MqttMessage(topic, ByteBuffer.wrap(payload)));
         }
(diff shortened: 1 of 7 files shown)


src/main/java/software/amazon/awssdk/crt/mqtt/MqttConnection.java
```diff
@@ -144,29 +119,78 @@ public void onSuccess() {
             }
 
             @Override
-            public void onFailure(String reason) {
-                Throwable cause = new MqttException(reason);
-                future.completeExceptionally(cause);
+            public void onFailure(Throwable reason) {
+                future.completeExceptionally(reason);
```
ColdenCullen: I DID REAL GOOD

this method name is funny, that's my joke

src/native/mqtt_connection.c
```diff
@@ -46,23 +46,26 @@ void s_cache_async_callback(JNIEnv *env) {
     assert(cls);
     s_async_callback.on_success = (*env)->GetMethodID(env, cls, "onSuccess", "()V");
     assert(s_async_callback.on_success);
-    s_async_callback.on_failure = (*env)->GetMethodID(env, cls, "onFailure", "(Ljava/lang/String;)V");
+    s_async_callback.on_failure = (*env)->GetMethodID(env, cls, "onFailure", "(Ljava/lang/Throwable;)V");
     assert(s_async_callback.on_failure);
 }
 
-/* methods of MqttConnection.ClientCallbacks */
+/* methods of MqttCConnectionEvents */
```
ColdenCullen: `MqttCCCCConnectionEvents`

src/main/java/software/amazon/awssdk/crt/mqtt/MqttConnection.java
```diff
@@ -144,29 +119,78 @@ public void onSuccess() {
...
+    @Override
+    public void onConnectionResumed(boolean sessionPresent) {
+        connectionState = ConnectionState.CONNECTED;
+        if (userConnectionCallbacks != null) {
+            userConnectionCallbacks.onConnectionResumed(sessionPresent);
+        }
     }
 
-    public CompletableFuture<Void> connect(String clientId, boolean cleanSession, int keepAliveMs) {
+    public CompletableFuture<Void> connect(String clientId, String endpointUri, int port) {
```
graebm: Shouldn't the CompletableFuture have `boolean sessionPresent` in it?

src/main/java/software/amazon/awssdk/crt/mqtt/MqttConnectionEvents.java
```diff
@@ -0,0 +1,38 @@
...
+
+/** 
+ * Interface used to receive connection events from the CRT 
+ */
+public interface MqttConnectionEvents {
+    /**
+     * called when a connection is resolved, one way or the other. Non-zero error
+     * code indicates connection never completed or was disconnected
+     */
+    void onConnectionComplete(int errorCode, boolean sessionPresent);
```
graebm: In python, we're just exposing connect-completion via the Future returned from connect().

Is there a compelling reason to offer the user 2 ways to learn of connect-completion (via this callback and also via the CompletableFuture)?

Also, if there's a compelling reason for connect-completion, why isn't there a disconnect-completion?

src/main/java/software/amazon/awssdk/crt/mqtt/MqttConnection.java
```diff
@@ -180,9 +204,9 @@ public void onFailure(String reason) {
             future.complete(null);
             return future;
         }
-        AsyncCallback disconnectAck = wrapVoidFuture(future);
+        disconnectAck = wrapVoidFuture(future);
         connectionState = ConnectionState.DISCONNECTING;
```
graebm: Doesn't need to be fixed in this CR, but I notice that this class doesn't protect any of its internal state with locks. So multiple threads calling `disconnect()` are going to scramble things like disconnectAck

src/main/java/software/amazon/awssdk/crt/mqtt/MqttConnection.java
justinboswell: It could indeed.

src/main/java/software/amazon/awssdk/crt/mqtt/MqttConnectionEvents.java
justinboswell: There isn't, I guess I could re-work this to pull that out.

src/main/java/software/amazon/awssdk/crt/mqtt/MqttConnection.java
justinboswell: Luckily, connectionState is volatile, and it basically doesn't matter atm, but you're right, I've been actively ignoring this for a while, needs a pass soon.

src/main/java/software/amazon/awssdk/crt/mqtt/MqttConnection.java
justinboswell: Yeah, welcome to Java!

PR #38, opened 2019-02-07
Continuous Deployment via CodeBuild/Jenkins

justinboswell:


CMakeLists.txt
```diff
@@ -16,7 +16,8 @@ if (BUILD_DEPS)
     set(AWS_DEPS_BUILD_DIR ${CMAKE_CURRENT_BINARY_DIR}/deps)
     set(AWS_DEPS_INSTALL_DIR ${CMAKE_CURRENT_BINARY_DIR}/deps/install)
 
-    list(APPEND CMAKE_MODULE_PATH "${AWS_DEPS_INSTALL_DIR}/${CMAKE_INSTALL_LIBDIR}/cmake")
+    list(APPEND CMAKE_MODULE_PATH "${AWS_DEPS_INSTALL_DIR}/lib/cmake")
+    list(APPEND CMAKE_MODULE_PATH "${AWS_DEPS_INSTALL_DIR}/lib64/cmake")
```
JonathanHenson: If you do my trick with the gnuinstall dirs stuff, then use CMAKE_INSTALL_LIBDIR you can avoid this.

Also probably want to use CMAKE_PREFIX_PATH

pom.xml
```diff
@@ -2,10 +2,42 @@
...
+            <url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
+        </license>
+    </licenses>
+
+    <developers>
+        <developer>
+            <name>AWS SDK Common Runtime Team</name>
+            <email>aws-sdk-common-runtime@amazon.com</email>
+            <organization>Amazon Web Services</organization>
+            <organizationUrl>http://aws.amazon.com</organizationUrl>
```
JonathanHenson: https.....

pom.xml
```diff
@@ -111,12 +137,68 @@
...
+                                <id>cmake-configure</id>
+                                <phase>generate-sources</phase>
+                                <goals>
+                                    <goal>exec</goal>
+                                </goals>
+                                <configuration>
+                                    <executable>${cmake.command}</executable>
+                                    <workingDirectory>mvn-build</workingDirectory>
+                                    <arguments>
+                                        <argument>-G${cmake.generator}</argument>
```
JonathanHenson: where's CMAKE_PREFIX_PATH?

CMakeLists.txt
justinboswell: I tried that, it did not work, I always got lib64 here, and all the other libs got lib, despite using GNUInstallDirs. I didn't troubleshoot it further yet.

pom.xml
justinboswell: Will fix.

pom.xml
justinboswell: Isn't necessary, because the deps build takes care of setting up CMAKE_MODULE_PATH, after which we just load the config scripts and we're good to go. An external user should not have to provide CMAKE_PREFIX_PATH unless they did their own compiling and are not automatically building deps.

PR #49, opened 2019-04-22
Autodetect IoT Certificates in the Credentials Directory

alexw91:
*Issue #, if available:* N/A 

*Description of changes:*
Readme [here](https://github.com/awslabs/aws-crt-java/blob/master/src/test/resources/credentials/README.md) says to place IoT test credentials in that directory, but AFAICT the tests don't actually auto-detect credentials in that directory.

This PR updates the `findCredentials()` method to actually find credentials in that directory. 


justinboswell:
So this would literally undo the change where I switched from a hardcoded directory to maven settings. I think we stick with the maven settings. We should instead fix the docs. I'm happy to do that next Fix-it-Friday

PR #52, opened 2019-04-29
Enforce Native Resource Memory Mangement in Unit Tests

alexw91:
*Issue #, if available:* N/A

*Description of changes:*
Adds checks to all existing Unit Tests that ensures Native Resources are freed correctly. Also fixes a bug uncovered by these new checks in `host_resolver.c` that wasn't cleaning up scheduled host resolver tasks from the Event Loop before freeing the Host Resolver, which could cause Segfaults if the task callback was called.

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/CrtResource.java
+++ b/src/main/java/software/amazon/awssdk/crt/CrtResource.java
@@ -14,11 +14,16 @@
  */
 package software.amazon.awssdk.crt;
 
+import java.util.Collection;
+import java.util.Collections;
+import java.util.concurrent.ConcurrentHashMap;
+
 /**
  * This wraps a native pointer to an AWS Common Runtime resource. It also ensures
  * that the first time a resource is referenced, the CRT will be loaded and bound.
  */
 public class CrtResource {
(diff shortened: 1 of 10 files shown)


src/main/java/software/amazon/awssdk/crt/CrtResource.java
```diff
@@ -14,26 +14,41 @@
...
     static {
         /* This will cause the JNI lib to be loaded the first time a CRT is created */
         new CRT();
     }
 
+    public static int getAllocatedNativeResourceCount() {
+        return NATIVE_RESOURCES.size();
+    }
+
+    public static Collection<String> getAllocatedNativeResources() {
```
justinboswell: Are the strings useful in identifying the outstanding objects? It feels like you'd need the address too, especially if trying to correlate them to logs or something.

alexw91: This function helps find the Java CrtResource Class Names that haven't been released yet, and is useful for finding CrtResource leaks. I'm hesitant to expose the raw pointers through the global public API since then other Threads can use the global API to get pointers that don't belong to them and do things they shouldn't be allowed to do.

PR #53, opened 2019-04-30
Move AsyncCallback to its own Class

alexw91:
*Issue #, if available:* N/A

*Description of changes:*
Move `AsyncCallback` into it's own class in preparation of adding `aws-c-http` integrations.

### DIFF
--- a/src/native/async_callback.c
+++ b/src/native/async_callback.c
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2010-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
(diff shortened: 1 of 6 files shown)


src/native/mqtt_connection.c
```diff
@@ -42,7 +42,7 @@ static struct {
 } s_async_callback = {0};
 
 void s_cache_async_callback(JNIEnv *env) {
-    jclass cls = (*env)->FindClass(env, "software/amazon/awssdk/crt/mqtt/MqttConnection$AsyncCallback");
+    jclass cls = (*env)->FindClass(env, "software/amazon/awssdk/crt/AsyncCallback");
```
justinboswell: The async header struct and this function should probably move into crt.c or something, and have a header which externs it, so you can use it in http and wherever else we need callbacks.

alexw91: Fixed

src/native/async_callback.c
```diff
@@ -0,0 +1,29 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+#include <assert.h>
+#include <jni.h>
+
+#include "async_callback.h"
+
+struct crt_async_callback global_async_callback;
```
justinboswell: g_ is sufficient, and matches our coding standard

alexw91: Fixed

PR #55, opened 2019-05-04
Refactor CrtResource

alexw91:
*Issue #, if available:* N/A

*Description of changes:*
Refactors the `CrtResource` class in preparation of adding Http Support. 

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/CrtResource.java
+++ b/src/main/java/software/amazon/awssdk/crt/CrtResource.java
@@ -16,14 +16,17 @@ package software.amazon.awssdk.crt;
 
 import java.util.Collection;
 import java.util.Collections;
+import java.util.LinkedList;
 import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * This wraps a native pointer to an AWS Common Runtime resource. It also ensures
  * that the first time a resource is referenced, the CRT will be loaded and bound.
  */
-public class CrtResource {
+public class CrtResource implements AutoCloseable {
     private static final ConcurrentHashMap<Long, String> NATIVE_RESOURCES = new ConcurrentHashMap<>();
+    private static final long NULL = 0;
+    private final LinkedList<CrtResource> ownedSubResources = new LinkedList<>();
     private long ptr;
 
     static {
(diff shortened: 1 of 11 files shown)


src/main/java/software/amazon/awssdk/crt/CrtResource.java
```diff
@@ -42,12 +46,42 @@ public static int getAllocatedNativeResourceCount() {
...
+            throw new IllegalStateException("Can't acquire >1 Native Pointer");
+        }
+
+        if (_ptr == NULL) {
+            throw new IllegalStateException("Can't acquire NULL Pointer");
+        }
+
+        String lastValue = NATIVE_RESOURCES.put(_ptr, this.getClass().getCanonicalName());
+
+        if (lastValue != null) {
```
justinboswell: This isn't a valid check, it is entirely possible to be given the same address when requesting memory for an object.

src/main/java/software/amazon/awssdk/crt/CrtResource.java
```diff
@@ -14,16 +14,20 @@
...
 import java.util.Collections;
+import java.util.LinkedList;
 import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * This wraps a native pointer to an AWS Common Runtime resource. It also ensures
  * that the first time a resource is referenced, the CRT will be loaded and bound.
  */
-public class CrtResource {
+public class CrtResource implements Closeable, AutoCloseable {
```
justinboswell: You only need AutoCloseable. it is undesirable to use both, and AutoCloseable supercedes Closeable.

src/main/java/software/amazon/awssdk/crt/CrtResource.java
alexw91: Won't you run into use-after-free if you have two `CrtResource`'s with the same native pointer and you close one of them?

src/main/java/software/amazon/awssdk/crt/CrtResource.java
alexw91: Fixed

src/main/java/software/amazon/awssdk/crt/CrtResource.java
justinboswell: No you're right, this is my mistake. Basically, there can only be 1 resource at a time per address. As long as dead resources are cleared (which they are), all is well. I'm going to chalk that one up to pre-coffee code review :)

PR #56, opened 2019-05-08
 Add JNI Http Connection Open/Close APIs 

alexw91:
## Issue #, if available:
N/A

## Description of changes:
Adds Integration with `aws-c-http` library for Http connection open and close API's, and adds Unit Tests that confirms connections are opened and closed. 

This Pull Request does **NOT** integrate with the Http request/response API yet, that will be added in a later PR.

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/http/HttpConnection.java
+++ b/src/main/java/software/amazon/awssdk/crt/http/HttpConnection.java
@@ -1,0 +1,197 @@
+/*
+ * Copyright 2010-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.crt.http;
+
+import software.amazon.awssdk.crt.AsyncCallback;
+import software.amazon.awssdk.crt.CrtResource;
+import software.amazon.awssdk.crt.CrtRuntimeException;
+import software.amazon.awssdk.crt.io.ClientBootstrap;
+import software.amazon.awssdk.crt.io.SocketOptions;
+import software.amazon.awssdk.crt.io.TlsContext;
+
+import java.net.URI;
+import java.util.concurrent.CompletableFuture;
+
+import static software.amazon.awssdk.crt.CRT.AWS_CRT_SUCCESS;
+
+
+/**
+ * This class wraps aws-c-http to provide the basic HTTP request/response functionality via the AWS Common Runtime.
+ *
+ * HttpConnection represents a single connection to a HTTP service endpoint.
+ *
+ * This class is not thread safe and should not be called from different threads.
+ */
+public class HttpConnection extends CrtResource {
+    private static final String HTTP = "http";
+    private static final String HTTPS = "https";
+    private static final int DEFAULT_HTTP_PORT = 80;
+    private static final int DEFAULT_HTTPS_PORT = 443;
+
+    private final ClientBootstrap clientBootstrap;
+    private final SocketOptions socketOptions;
+    private final TlsContext tlsContext;
+    private final URI uri;
+    private final int port;
+    private final boolean useTls;
+
+    private final CompletableFuture<HttpConnection> connectedFuture;
+    private final CompletableFuture<Void> shutdownFuture;
+
+    /**
+     * Creates a new CompletableFuture for a new HttpConnection.
+     * @param uri Must be non-null and contain a hostname
+     * @param bootstrap The ClientBootstrap to use for the Connection
+     * @param socketOptions The SocketOptions to use for the Connection
+     * @param tlsContext The TlsContext to use for the Connection
+     * @return CompletableFuture indicating when the connection has completed
+     * @throws CrtRuntimeException if Native threw a CrtRuntimeException
+     */
+    public static CompletableFuture<HttpConnection> createConnection(URI uri, ClientBootstrap bootstrap,
+                                                                     SocketOptions socketOptions, TlsContext tlsContext) throws CrtRuntimeException {
+        HttpConnection conn = new HttpConnection(uri, bootstrap, socketOptions, tlsContext);
+        return conn.connect();
+    }
+
+    /**
+     * Constructs a new HttpConnection.
+     * @param uri Must be non-null and contain a hostname
+     * @param bootstrap The ClientBootstrap to use for the Connection
+     * @param socketOptions The SocketOptions to use for the Connection
+     * @param tlsContext The TlsContext to use for the Connection
+     */
+    private HttpConnection(URI uri, ClientBootstrap bootstrap, SocketOptions socketOptions, TlsContext tlsContext) {
+        if (uri == null) {  throw new IllegalArgumentException("URI must not be null"); }
+        if (uri.getScheme() == null) { throw new IllegalArgumentException("URI does not have a Scheme"); }
+        if (!HTTP.equals(uri.getScheme()) && !HTTPS.equals(uri.getScheme())) { throw new IllegalArgumentException("URI has unknown Scheme"); }
+        if (uri.getHost() == null) { throw new IllegalArgumentException("URI does not have a Host name"); }
+        if (bootstrap == null || bootstrap.isNull()) {  throw new IllegalArgumentException("ClientBootstrap must not be null"); }
+        if (socketOptions == null || socketOptions.isNull()) { throw new IllegalArgumentException("SocketOptions must not be null"); }
+        if (HTTPS.equals(uri.getScheme()) && tlsContext == null) { throw new IllegalArgumentException("TlsContext must not be null if https is used"); }
+
+        int port = uri.getPort();
+
+        /* Pick a default port based on the scheme if one wasn't set in the URI */
+        if (port == -1) {
+            if (HTTP.equals(uri.getScheme()))  { port = DEFAULT_HTTP_PORT; }
+            if (HTTPS.equals(uri.getScheme())) { port = DEFAULT_HTTPS_PORT; }
+        }
+
+        if (port <= 0 || 65535 < port) {
+            throw new IllegalArgumentException("Invalid or missing port: " + uri);
+        }
+
+        this.uri = uri;
+        this.port = port;
+        this.useTls = HTTPS.equals(uri.getScheme());
+        this.clientBootstrap = bootstrap;
+        this.socketOptions = socketOptions;
+        this.tlsContext = tlsContext;
+        this.connectedFuture = new CompletableFuture<>();
+        this.shutdownFuture = new CompletableFuture<>();
+    }
+
+    /**
+     * Schedules a connection task on the EventLoop to connect to the Http Endpoint
+     * @return Future indicating when the connection has completed.
+     */
+    private CompletableFuture<HttpConnection> connect() throws CrtRuntimeException {
+        if (!isNull()) {
+            return connectedFuture;
+        }
+
+        acquire(httpConnectionNew(this,
+                clientBootstrap.native_ptr(),
+                socketOptions.native_ptr(),
+                useTls ? tlsContext.native_ptr() : 0,
+                uri.getHost(),
+                port));
+
+        return connectedFuture;
+    }
+
+    /**
+     * Closes and frees this HttpConnection and any native sub-resources associated with this connection
+     */
+    @Override
+    public void close() {
+        if (!isNull()) {
+            httpConnectionRelease(release());
+        }
+
+        super.close();
+    }
+
+    /** Called from Native EventLoop when the connection is established the first time **/
+    private void onConnectionComplete(int errorCode) {
+        if (errorCode == AWS_CRT_SUCCESS) {
+            connectedFuture.complete(this);
+        } else {
+            /* The user can't close this HttpConnection object since they never got a reference to it through the
+                CompletableFuture, so we need to close ourselves on Connection Error. */
+            this.close();
+            connectedFuture.completeExceptionally(new HttpException(errorCode));
+        }
+    }
+
+    /** Called from Native EventLoop when the connection is shutdown. **/
+    private void onConnectionShutdown(int errorCode) {
+        if (errorCode == AWS_CRT_SUCCESS) {
+           shutdownFuture.complete(null);
+        } else {
+            shutdownFuture.completeExceptionally(new HttpException(errorCode));
+        }
+    }
+
+    /**
+     * Returns the Shutdown Future
+     * @return CompletableFuture that will be completed when the connection is shut down.
+     */
+    public CompletableFuture<Void> getShutdownFuture() {
+        return shutdownFuture;
+    }
+
+    /**
+     * Schedules a task on the Native EventLoop to shut down the current connection
+     * @return When this future completes, the shutdown is complete
+     */
+    public CompletableFuture<Void> shutdown() {
+        if (isNull()) {
+           return shutdownFuture;
+        }
+        try {
+            httpConnectionShutdown(native_ptr());
+        } catch (CrtRuntimeException e) {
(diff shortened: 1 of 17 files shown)


src/main/java/software/amazon/awssdk/crt/http/HttpConnection.java
```diff
@@ -0,0 +1,218 @@
...
+                connectAck = null;
+            }
+        }
+        if (userConnectionCallbacks != null) {
+            userConnectionCallbacks.onConnectionComplete(errorCode);
+        }
+    }
+
+    /** Called from native when the connection is shutdown. **/
+    private synchronized void onConnectionShutdown(int errorCode) {
```
JonathanHenson: It appears the only state you're protecting with synchronized is the connectionState. Why not just make that atomic? the underlying c code is thread safe.

src/main/java/software/amazon/awssdk/crt/http/HttpConnection.java
```diff
@@ -0,0 +1,218 @@
...
+     * @throws HttpException If httpClient is null
+     */
+    public HttpConnection(URI uri, ClientBootstrap bootstrap, SocketOptions socketOptions, TlsContext tlsContext,
+                          HttpConnectionEvents callbacks) throws HttpException {
+
+        if (uri == null) {  throw new IllegalArgumentException("URI must not be null"); }
+        if (uri.getHost() == null) { throw new IllegalArgumentException("URI does not have a Host name"); }
+        if (bootstrap == null || bootstrap.isNull()) {  throw new IllegalArgumentException("ClientBootstrap must not be null"); }
+        if (socketOptions == null || socketOptions.isNull()) { throw new IllegalArgumentException("SocketOptions must not be null"); }
+        if ((HTTPS.equals(uri.getScheme()) || uri.getPort() == 443) && tlsContext == null) { throw new IllegalArgumentException("TlsContext must not be null if https is used"); }
```
andrewhop: Is the `uri.getPort() == 443` part a safe assumption? What if some clients have a proxy so for some reason they connect over HTTP on port 443?

src/main/java/software/amazon/awssdk/crt/http/HttpConnection.java
```diff
@@ -0,0 +1,218 @@
...
+        if (!isNull()) {
+            return connectedFuture;
+        }
+
+        String endpoint = uri.getHost();
+        int port = uri.getPort();
+
+        /* Pick a default port if one wasn't set in the URI */
+        if (port == -1) {
+            port = HTTP.equals(uri.getScheme()) ? 80 : 443;
```
andrewhop: NP: default port number constants and up on L84.

src/native/http_connection.c
```diff
@@ -0,0 +1,256 @@
...
+    struct aws_allocator *allocator = aws_jni_get_allocator();
+    struct aws_byte_cursor endpoint = aws_jni_byte_cursor_from_jstring(env, jni_endpoint);
+    uint16_t port = jni_port;
+
+    if (port <= 0 || 65536 <= port) {
+        aws_jni_throw_runtime_exception(env, "HttpConnection.httpConnectionNew: Port must be between 1 and 65535");
+        return (jlong)NULL;
+    }
+
+    int use_tls = !(port == 80 || port == 8080);
```
andrewhop: I also don't think this is a guarantee. Why not rely on the jni_tls_ctx being set or not?

src/main/java/software/amazon/awssdk/crt/http/HttpConnection.java
alexw91: There's potential for race conditions between the EventLoop Thread and the User Thread if `synchronized` is removed.

Here's an example from the `HttpConnection.connect()` method:
```
// The httpConnectionNew call will schedule a task on the EventLoop to call onConnectionComplete() when the connection is completed (and set the connection's internal state to CONNECTED)
acquire(httpConnectionNew(this,
                clientBootstrap.native_ptr(),
                socketOptions.native_ptr(),
                tlsContext != null ? tlsContext.native_ptr() : 0,
                endpoint,
                (short) port));
// If this method was not synchronized, and the EventLoop task on the above line was completed "instantaneously" here, the User Thread would then overwrite the state from CONNECTED to CONNECTING.
connectionState = ConnectionState.CONNECTING;
```

Even if I migrate to `AtomicReference<ConnectionState>` and do compare-and-swap everywhere, it still opens the door for compare-and-swap failures if the EventLoop is "too fast" or UserThread "too slow", which would be extremely rare and hard to reproduce connection failures.

src/main/java/software/amazon/awssdk/crt/http/HttpConnection.java
```diff
@@ -0,0 +1,218 @@
...
+    private final URI uri;
+    private final HttpConnectionEvents userConnectionCallbacks;
+
+    private ConnectionState connectionState = ConnectionState.DISCONNECTED;
+    private CompletableFuture<Boolean> connectedFuture;
+    private AsyncCallback connectAck;
+    private AsyncCallback disconnectAck;
+
+    public enum ConnectionState {
+        DISCONNECTED, CONNECTING, CONNECTED, SHUTTING_DOWN, SHUT_DOWN
```
graebm: Probably don't need to track state here in Java.
If the connection is provided via CompletableFuture you could remove the CONNECTING state.

Also, the C implementation will give appropriate errors if its functions are called while it's shutting down or shut down. You can just blithely call into the C-layer without worrying.

src/main/java/software/amazon/awssdk/crt/http/HttpConnection.java
```diff
@@ -0,0 +1,218 @@
...
+        own(this.tlsContext);
+    }
+
+    /**
+     * Constructs a new HttpConnection. Connections are reusable after being disconnected.
+     * @param uri Must be non-null
+     * @param callbacks Optional handler for connection interruptions/resumptions
+     * @throws HttpException If httpClient is null
+     */
+    public HttpConnection(URI uri, ClientBootstrap bootstrap, SocketOptions socketOptions, TlsContext tlsContext,
```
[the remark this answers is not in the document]
justinboswell: It's worth noting that this is how it's handled in C++, Python, and .NET

There is ```HttpClientConnection.NewConnection()``` and ```HttpClientConnection.MakeRequest()```. Both return futures. ```NewConnection``` yields a ```HttpClientConnection```, and ```MakeRequest``` yields a result for the entire stream operation. I would suggest looking at the .NET bindings, as they will be closest to Java.

src/main/java/software/amazon/awssdk/crt/http/HttpConnection.java
justinboswell: This is probably echoing the state tracking in the MqttConnection, which is also probably wrong. I thought it was a good idea last year when I wrote it, I'm pretty well convinced of removing it before GA.

src/main/java/software/amazon/awssdk/crt/http/HttpConnection.java
justinboswell: As noted below, the most likely solution is not to track this state at all, since callbacks are available for every state change.

src/main/java/software/amazon/awssdk/crt/http/HttpConnection.java
justinboswell: That's a good catch, this will definitely fail when proxies get involved. We don't assume this anywhere other than unit tests at the moment, for that reason. If the user gives a bad port, that's on them, and the connection will fail accordingly.

src/main/java/software/amazon/awssdk/crt/http/HttpConnection.java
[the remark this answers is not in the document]
alexw91: I was following the way it was done for MqttConnection:
 - https://github.com/awslabs/aws-crt-java/blob/master/src/main/java/software/amazon/awssdk/crt/mqtt/MqttConnection.java#L81
 - https://github.com/awslabs/aws-crt-java/blob/master/src/main/java/software/amazon/awssdk/crt/mqtt/MqttConnection.java#L176

src/main/java/software/amazon/awssdk/crt/http/HttpConnection.java
```diff
@@ -0,0 +1,218 @@
...
+        return connectedFuture;
+    }
+
+    /**
+     * Disconnects if necessary, and frees native resources associated with this connection
+     */
+    @Override
+    public synchronized void close() {
+        if (!isNull()) {
+            httpConnectionRelease(release());
```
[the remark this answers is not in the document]
alexw91: `httpConnectionRelease()` will also shutdown the connection if it wasn't already being shutdown.

 - https://github.com/awslabs/aws-c-http/blob/master/include/aws/http/connection.h#L130-L137

src/main/java/software/amazon/awssdk/crt/http/HttpConnection.java
alexw91: Removed `ConnectionState` Enum

src/native/http_connection.c
```diff
@@ -0,0 +1,256 @@
...
+    jlong jni_tls_ctx,
+    jstring jni_endpoint,
+    jshort jni_port) {
+
+    struct aws_client_bootstrap *client_bootstrap = (struct aws_client_bootstrap *)jni_client_bootstrap;
+    struct aws_socket_options *socket_options = (struct aws_socket_options *)jni_socket_options;
+    struct aws_tls_ctx *tls_ctx = (struct aws_tls_ctx *)jni_tls_ctx;
+
+    if (!client_bootstrap) {
+        aws_jni_throw_runtime_exception(env, "HttpConnection.httpConnectionNew: jni_client_bootstrap can't be null");
```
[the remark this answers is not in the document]
alexw91: Cleaned up all calls to `aws_jni_throw_runtime_exception`

src/native/http_connection.c
```diff
@@ -0,0 +1,256 @@
...
+};
+
+static void s_on_http_conn_setup(struct aws_http_connection *connection, int error_code, void *user_data) {
+    struct http_jni_connection *http_jni_conn = (struct http_jni_connection *)user_data;
+    assert(http_jni_conn);
+
+    // Save the native pointer
+    http_jni_conn->native_http_conn = connection;
+
+    AWS_LOGF_DEBUG(AWS_LS_HTTP_CONNECTION, "Connected: http_connection: %p, err: %d", (void *)connection, error_code);
```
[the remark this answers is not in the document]
alexw91: Removed all `AWS_LOGF_DEBUG` calls

src/main/java/software/amazon/awssdk/crt/io/TlsContext.java
```diff
@@ -36,14 +34,23 @@ public TlsContext(TlsContextOptions options) throws CrtRuntimeException {
         acquire(tlsContextNew(options.native_ptr()));
     }
 
+    /**
+     * Creates a new TlsContext. There are significant native resources consumed to create a TlsContext, so most
```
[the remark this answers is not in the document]
alexw91: Updated JavaDoc to use Client TLS Context

src/native/http_connection.c
alexw91: Updated Java side to require `http` or `https` Scheme for all URI's, and use the URI scheme to determine whether to use TLS or not. Java side then uses Scheme to decide whether or not to pass TlsContext/NULL to Native, and Native now always uses `TlsContext` if present.

src/main/java/software/amazon/awssdk/crt/http/HttpConnection.java
alexw91: Fixed

src/main/java/software/amazon/awssdk/crt/http/HttpConnection.java
alexw91: Updated Java side to require `http` or `https` Scheme for all URI's, and use the URI scheme to determine whether to use TLS or not. Java side then uses Scheme to decide whether or not to pass TlsContext/NULL to Native, and Native now always uses `TlsContext` if present.

src/main/java/software/amazon/awssdk/crt/http/HttpConnection.java
alexw91: Removed `ConnectionState` Enum

src/native/http_connection.c
```diff
@@ -0,0 +1,256 @@
...
+}
+
+JNIEXPORT void JNICALL Java_software_amazon_awssdk_crt_http_HttpConnection_httpConnectionShutdown(
+    JNIEnv *env,
+    jclass jni_class,
+    jlong jni_connection) {
+
+    struct http_jni_connection *http_jni_conn = (struct http_jni_connection *)jni_connection;
+
+    http_jni_conn->shutdown_requested = true;
```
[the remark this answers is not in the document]
alexw91: `shutdown_requested` wasn't used for anything, and `httpConnectionRelease()` will shutdown the connection if it's not already shutdown. Removed this variable.

src/main/java/software/amazon/awssdk/crt/http/HttpConnection.java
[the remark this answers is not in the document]
justinboswell: I think you will probably want to match the other languages, for consistency.

src/main/java/software/amazon/awssdk/crt/http/HttpConnection.java
```diff
@@ -0,0 +1,207 @@
...
+        }
+
+        super.close();
+    }
+
+    /** Called from native when the connection is established the first time **/
+    private void onConnectionComplete(int errorCode) {
+        if (userConnectionCallbacks != null) {
+            userConnectionCallbacks.onConnectionComplete(errorCode);
```
justinboswell: There should not be user connection callbacks in addition to the future, the future is enough. Otherwise, you put the user in a position where they get the same information twice.

src/main/java/software/amazon/awssdk/crt/http/HttpConnection.java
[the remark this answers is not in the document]
JonathanHenson: Mqtt is a different API largely because the connection lifetime is managed differently, I'd stick to the way the other languages are done for http.

src/main/java/software/amazon/awssdk/crt/http/HttpConnection.java
JonathanHenson: you can put a callback on the future anyways. Agreed, don't mix futures and callbacks.

alexw91: Fixed

src/main/java/software/amazon/awssdk/crt/http/HttpConnection.java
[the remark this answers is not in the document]
alexw91: Moved to static `createConnection` method.

src/main/java/software/amazon/awssdk/crt/http/HttpException.java
```diff
@@ -0,0 +1,46 @@
...
+ * This exception will be thrown by any exceptional cases encountered within the
+ * JNI bindings to the AWS Common Runtime
+ */
+public class HttpException extends RuntimeException {
+    private static final int DEFAULT_ERROR_CODE = -1;
+    private int errorCode;
+
+    public HttpException(String msg) {
+        super(msg);
+        this.errorCode = DEFAULT_ERROR_CODE;
```
justinboswell: Don't have to do this now, but it wouldn't hurt to resolve the error code to a string to make the exception message helpful. See ```CRT.awsErrorString()```

src/main/java/software/amazon/awssdk/crt/http/HttpConnection.java
```diff
@@ -0,0 +1,189 @@
...
+        }
+    }
+
+    /**
+     * Schedules a task on the Native EventLoop to shut down the current connection
+     * @return When this future completes, the shutdown is complete
+     */
+    public CompletableFuture<Void> shutdown() {
+        if (isNull()) {
+           return shutdownFuture;
```
graebm: Since the `shutdownFuture` is private, users can't be notified of the event unless they trigger a shutdown themselves. Users might want to be notified ASAP of shutdowns that happen unexpectedly

alexw91: Added `getShutdownFuture()` method

src/main/java/software/amazon/awssdk/crt/http/HttpException.java
alexw91: Removed this String constuctor, leaving only one that takes the CRT ErrorCode and resolves the Error string using `CRT.awsErrorString()`.

PR #57, opened 2019-05-21
Add Http Request and Response APIs

alexw91:
## *Issue #, if available:* 
N/A

## *Description of changes:*
Adds Integration with `aws-c-http` library for Http Client request/response API's, and adds Unit Tests that makes Http Requests and validates Http Responses. 

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/http/CrtHttpStreamHandler.java
+++ b/src/main/java/software/amazon/awssdk/crt/http/CrtHttpStreamHandler.java
@@ -1,0 +1,97 @@
+/*
+ * Copyright 2010-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.awssdk.crt.http;
+
+import java.nio.ByteBuffer;
+
+/**
+ * Interface that Native code knows how to call when handling Http Request/Responses
+ *
+ * Maps 1-1 to the Native Http API here: https://github.com/awslabs/aws-c-http/blob/master/include/aws/http/request_response.h
+ */
+public interface CrtHttpStreamHandler {
+
+    /**
+     * Called from Native when new Http Headers have been received.
+     * Note that this function may be called multiple times as HTTP headers are received.
+     *
+     * @param stream The HttpStream object
+     * @param responseStatusCode The HTTP Response Status Code
+     * @param nextHeaders The headers received in the latest IO event.
+     */
+    void onResponseHeaders(HttpStream stream, int responseStatusCode, HttpHeader[] nextHeaders);
+
+    /**
+     * Called from Native once all HTTP Headers are processed. Will not be called if there are no Http Headers in the
+     * response. Guaranteed to be called exactly once if there is at least 1 Header.
+     *
+     * @param stream The HttpStream object
+     * @param hasBody True if the HTTP Response had a Body, false otherwise.
+     */
+    default void onResponseHeadersDone(HttpStream stream, boolean hasBody) {
+        /* Optional Callback, do nothing by default */
+    }
+
+    /**
+     * Called when new Body bytes have been received.
+     * Note that this function may be called multiple times as bodyBytes are received.
+     *
+     * Do NOT keep a reference to this ByteBuffer past the lifetime of this function call. The CommonRuntime reserves
+     * the right to use DirectByteBuffers pointing to memory that only lives as long as the function call.
+     *
+     * Sliding Window:
+     * The Native HttpConnection EventLoop will keep sending data until the end of the sliding Window is reached.
+     * The user application is responsible for setting the initial Window size appropriately when creating the
+     * HttpConnection, and for incrementing the sliding window appropriately throughout the lifetime of the HttpStream.
+     *
+     * For more info, see:
+     *  - https://en.wikipedia.org/wiki/Sliding_window_protocol
+     *
+     * @param bodyBytesIn The HTTP Body Bytes received in the last IO Event. The user MUST either copy all bytes from
+     *                      this Buffer, or copy a reference to this ByteBuffer, since there will not be another chance
+     *                      to read this data.
+     * @return The number of bytes to move the sliding window by. Repeatedly returning zero will eventually cause the
+     *          sliding window to fill up and stop data from flowing until the user slides the window back open.
+     */
+    default int onResponseBody(HttpStream stream, ByteBuffer bodyBytesIn) {
+        /* Optional Callback, ignore incoming response body by default unless user wants to capture it. */
+        return bodyBytesIn.remaining();
+    }
+
+    /**
+     * Called from Native when the Response has completed.
+     * @param stream
+     * @param errorCode
+     */
+    void onResponseComplete(HttpStream stream, int errorCode);
+
+    /**
+     * Called from Native when the Http Request has a Body (Eg PUT/POST requests).
+     * Note that this function may be called many times as Native sends the Request Body.
+     *
+     * Do NOT keep a reference to this ByteBuffer past the lifetime of this function call. The CommonRuntime reserves
+     * the right to use DirectByteBuffers pointing to memory that only lives as long as the function call.
+     *
+     * @param stream The HttpStream for this Request/Response Pair
+     * @param bodyBytesOut The Buffer to write the Request Body Bytes to.
+     * @return True if Request body is complete, false otherwise.
+     */
+    default boolean sendRequestBody(HttpStream stream, ByteBuffer bodyBytesOut) {
+        /* Optional Callback, return empty request body by default unless user wants to return one. */
+        return true;
+    }
+
+}
--- a/src/main/java/software/amazon/awssdk/crt/http/CrtHttpStreamHandler.java
+++ b/src/main/java/software/amazon/awssdk/crt/http/CrtHttpStreamHandler.java
@@ -61,10 +61,9 @@ public interface CrtHttpStreamHandler {
      *  - https://en.wikipedia.org/wiki/Sliding_window_protocol
      *
      * @param bodyBytesIn The HTTP Body Bytes received in the last IO Event. The user MUST either copy all bytes from
-     *                      this Buffer, or copy a reference to this ByteBuffer, since there will not be another chance
-     *                      to read this data.
+     *                      this Buffer, since there will not be another chance to read this data.
      * @return The number of bytes to move the sliding window by. Repeatedly returning zero will eventually cause the
-     *          sliding window to fill up and stop data from flowing until the user slides the window back open.
+     *          sliding window to fill up and data to stop flowing until the user slides the window back open.
      */
     default int onResponseBody(HttpStream stream, ByteBuffer bodyBytesIn) {
         /* Optional Callback, ignore incoming response body by default unless user wants to capture it. */
--- a/src/main/java/software/amazon/awssdk/crt/http/HttpConnection.java
+++ b/src/main/java/software/amazon/awssdk/crt/http/HttpConnection.java
@@ -15,7 +15,6 @@
 
 package software.amazon.awssdk.crt.http;
 
-import software.amazon.awssdk.crt.AsyncCallback;
 import software.amazon.awssdk.crt.CrtResource;
 import software.amazon.awssdk.crt.CrtRuntimeException;
 import software.amazon.awssdk.crt.io.ClientBootstrap;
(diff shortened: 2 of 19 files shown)


alexw91:
Note that I needed to make this workaround to get the Unit Tests to consistently pass: https://github.com/awslabs/aws-crt-java/pull/57/files#diff-73d8acf2b44a3e8400a4c81ec3612cebR179

src/test/java/software/amazon/awssdk/crt/test/HttpRequestResponseTest.java
```diff
@@ -0,0 +1,318 @@
...
+             * The conn.close() call above schedules a shutdown task on the Native EventLoop. That task will shut down
+             * the s2n TLS connection and sends a TLS CloseNotify message to the TLS peer, and needs access to the
+             * TlsContext below.
+             *
+             * The tlsContext.close() call below immediately frees the native TlsContext structures used by s2n when
+             * sending the TLS CloseNotify message.
+             *
+             * Need to sleep here to allow enough time to pass for the task on the EventLoop to complete.
+             */
+            Thread.sleep(100);
```
graebm: You need to wait until the connection finishes shutting down, like: `conn.shutdown().get()`

There might still be a bug where that event fires sliiiightly too early, but that's how the code ought to go.

alexw91: Updated tests to always wait for shutdown before closing. 

Do we want to update the [documentation here?](https://github.com/awslabs/aws-c-http/blob/master/include/aws/http/connection.h#L130-L137)

Something along the lines of this?
```
Calling release before shutdown has completed may cause race conditions and/or
Segfaults if you later free memory that can be used by shutdown tasks for this connection.
You should always wait for shutdown to complete before releasing.
```

src/test/java/software/amazon/awssdk/crt/test/HttpConnectionTest.java
```diff
@@ -18,33 +18,38 @@
...
             actuallyConnected = true;
-            conn.shutdown().get();
-            conn.close();
+
         } catch (Exception e) {
             exceptionThrown = true;
             Assert.assertTrue(e.getMessage(), e.getMessage().contains(exceptionMsg));
         } finally {
+            if (conn != null) {
+                conn.shutdown().get();
```
justinboswell: Why not just have ```close()``` do this? This won't protect users in the wild.

alexw91: If `close()` automatically waited for `shutdown()` to complete then [this line](https://github.com/awslabs/aws-crt-java/pull/57/files#diff-dbee0f3e9747085aa04f4f613340a732R215) would block the EventLoop.

justinboswell: I'm not sure that's true. If the future has already completed, you'll have no wait at all, and it'll throw, right? Am I missing something?

src/main/java/software/amazon/awssdk/crt/http/CrtHttpStreamHandler.java
```diff
@@ -0,0 +1,83 @@
...
+    void onResponseHeaders(HttpStream stream, int responseStatusCode, HttpHeader[] nextHeaders);
+
+    /**
+     * Called from Native once all HTTP Headers are processed. Will not be called if there are no Http Headers in the
+     * response. Guaranteed to be called exactly once if there is at least 1 Header.
+     *
+     * @param stream The HttpStream object
+     * @param hasBody True if the HTTP Response had a Body, false otherwise.
+     */
+    void onResponseHeadersDone(HttpStream stream, boolean hasBody);
```
justinboswell: This should be defaulted

src/main/java/software/amazon/awssdk/crt/http/CrtHttpStreamHandler.java
```diff
@@ -0,0 +1,83 @@
...
+     * For more info, see:
+     *  - https://en.wikipedia.org/wiki/Sliding_window_protocol
+     *
+     * @param bodyBytesIn The HTTP Body Bytes received in the last IO Event. The user MUST either copy all bytes from
+     *                      this Buffer, or copy a reference to this ByteBuffer, since there will not be another chance
+     *                      to read this data.
+     * @return The number of bytes to move the sliding window by. Repeatedly returning zero will eventually cause the
+     *          sliding window to fill up and stop data from flowing until the user slides the window back open.
+     */
+    int onResponseBody(HttpStream stream, ByteBuffer bodyBytesIn);
```
justinboswell: This should be defaulted

src/main/java/software/amazon/awssdk/crt/http/CrtHttpStreamHandler.java
```diff
@@ -0,0 +1,83 @@
...
+
+    /**
+     * Called from Native when the Http Request has a Body (Eg PUT/POST requests).
+     * Note that this function may be called many times as Native sends the Request Body.
+     *
+     * @param stream The HttpStream for this Request/Response Pair
+     * @param bodyBytesOut The Buffer to write the Request Body Bytes to.
+     * @return True if Request body is complete, false otherwise.
+     */
+    boolean sendRequestBody(HttpStream stream, ByteBuffer bodyBytesOut);
```
justinboswell: This should be defaulted.

src/main/java/software/amazon/awssdk/crt/http/HttpConnection.java
```diff
@@ -117,17 +138,61 @@ private HttpConnection(URI uri, ClientBootstrap bootstrap, SocketOptions socketO
...
+    /**
+     * Schedules an HttpRequest on the Native EventLoop for this HttpConnection.
+     *
+     * @param request The Request to make to the Server.
+     * @param streamHandler The Stream Handler to be called from the EventLoop
+     * @throws CrtRuntimeException
+     * @return The HttpStream that represents this Request/Response Pair. It can be closed at any time during the
+     *          request/response, but must be closed by the user thread making this request when it's done.
+     */
+    public HttpStream executeRequest(HttpRequest request, CrtHttpStreamHandler streamHandler) throws CrtRuntimeException {
```
justinboswell: This function is MakeRequest/make_request in other SDKs, this should probably match

src/native/http_request_response.c
```diff
@@ -0,0 +1,481 @@
...
+    jobject jByteBuffer = aws_jni_byte_buffer_copy_from_cursor(env, data);
+    jint window_increment = (*env)->CallIntMethod(
+        env,
+        callback->java_crt_http_callback_handler,
+        s_crt_http_stream_handler.onResponseBody,
+        callback->java_http_stream,
+        jByteBuffer);
+
+    // Verify Window Increment is within bounds
+    assert(0 <= window_increment);
```
justinboswell: These should probably throw an exception or clamp. We throw an exception in .NET.

src/native/crt.c
```diff
@@ -44,6 +44,105 @@ void aws_jni_throw_runtime_exception(JNIEnv *env, const char *msg, ...) {
...
+    (*env)->CallObjectMethod(env, jByteBuffer, s_java_byte_buffer.set_limit, (jint)native_data->len);
+
+    return jByteBuffer;
+}
+
+/**
+ * Converts a Native aws_byte_cursor to a Java DirectByteBuffer
+ */
+jobject aws_jni_direct_byte_buffer_from_byte_buf(JNIEnv *env, const struct aws_byte_buf *dst) {
+    jobject jByteBuf = (*env)->NewDirectByteBuffer(env, (void *)dst->buffer, (jlong)dst->capacity);
```
justinboswell: This can only be used in cases where you guarantee that the ByteBuffer will be copied or dealt with within the scope of the callback you're calling. If any reference is kept to it (for instance, to re-assemble a large file), it will be pointing at dead memory the moment you leave the native callback. 

You probably want to stick to byte[] and do a copy to guarantee safety.

src/native/http_request_response.c
```diff
@@ -0,0 +1,481 @@
...
+
+enum aws_http_outgoing_body_state s_stream_outgoing_body_fn(
+    struct aws_http_stream *stream,
+    struct aws_byte_buf *dst,
+    void *user_data) {
+
+    struct http_stream_callback_data *callback = (struct http_stream_callback_data *)user_data;
+    JNIEnv *env = aws_jni_get_thread_env(callback->connection->jvm);
+
+    jobject jDirectByteBuffer = aws_jni_direct_byte_buffer_from_byte_buf(env, dst);
```
justinboswell: See above about NewDirectByteBuffer, this is going to be a problem if someone keeps a reference to this ByteBuffer, this should probably be a byte[] copy.

justinboswell: Ok, after discussion in chat, forget this, we're going to leave it up to the user.

src/native/crt.c
justinboswell: Actually, on second thought, just note that in the comment in the header for this function, that the byte buffer is an alias, not a copy

src/main/java/software/amazon/awssdk/crt/http/CrtHttpStreamHandler.java
alexw91: I wanted this PR to be more focused on the Java interface than the implementation. I'd prefer to punt the Java implementation to a later PR, or have the implementation live in [aws-sdk-java-v2](https://github.com/aws/aws-sdk-java-v2/tree/master/http-clients).

src/test/java/software/amazon/awssdk/crt/test/HttpConnectionTest.java
alexw91: > If the future has already completed

The shutdown future is [guaranteed to not be complete](https://github.com/awslabs/aws-c-http/blob/92c780a4f5bb20c54b91efa9bbe7b3988aecd6fa/include/aws/http/connection.h#L100-L101) when the connection fails to connect.

I'm trying to avoid calling and waiting for the `on_shutdown()` callback to complete while inside the `on_conn_complete()` callback. Users don't need to wait for shutdown before releasing the connection in that case, but should wait for shutdown in other cases.

alexw91: Addressed both of our issues. `close()` now automatically waits for `shutdown()` so users don't accidentally shoot themselves in the foot, but I also added logic skip shutdown [if we never successfully connected in the first place](https://github.com/awslabs/aws-crt-java/commit/9ee099d83b756a65572a0c1e722003ce48326d0d#diff-dbee0f3e9747085aa04f4f613340a732R181).

src/main/java/software/amazon/awssdk/crt/http/HttpConnection.java
alexw91: Fixed

src/native/crt.c
alexw91: Fixed

src/main/java/software/amazon/awssdk/crt/http/CrtHttpStreamHandler.java
justinboswell: Yes, that's why I'm bringing it up. Java interfaces can have a [default implementation](https://docs.oracle.com/javase/tutorial/java/IandI/defaultmethods.html) (which in all of the cases I've highlighted should do nothing). The important thing here is that an implementer is only required to implement onHeader and onComplete.

src/native/http_request_response.c
justinboswell: can get rid of the asserts then, you'll crash in debug before delivering the exception, or you'll get a native exception (depending on platform).

alexw91: Fixed

src/main/java/software/amazon/awssdk/crt/http/HttpConnection.java
```diff
@@ -164,6 +246,15 @@ private void onConnectionShutdown(int errorCode) {
...
     }
 
+    public boolean didConnectSuccessfully() {
+        return connectedFuture.isDone() && !connectedFuture.isCompletedExceptionally();
+    }
+
+
+    public boolean isShutdownComplete() {
+        return shutdownFuture.isDone();
+    }
```
justinboswell: We've removed public state-exposing functions from the other languages. These could probably be private.

src/native/http_request_response.c
```diff
@@ -239,6 +246,9 @@ static void s_on_incoming_headers_fn(
         resp_status,
         jHeaders);
 
+    // Other threads might edit the callback struct, so ensure that we gain a lock on it
```
SalusaSecondus: Comment does not match following line.

src/native/http_request_response.c
```diff
@@ -258,6 +272,9 @@ static void s_on_incoming_header_block_done_fn(struct aws_http_stream *stream, b
         callback->java_http_stream,
         jHasBody);
 
+    // Other threads might edit the callback struct, so ensure that we gain a lock on it
```
SalusaSecondus: Comment does not match following line.

src/native/http_request_response.c
```diff
@@ -363,6 +397,8 @@ enum aws_http_outgoing_body_state s_stream_outgoing_body_fn(
 
     (*env)->DeleteGlobalRef(env, jByteBuffer);
 
+    aws_mutex_unlock(&callback->lock);
```
SalusaSecondus: Why not combine this unlock with the one in the exception check (line 387) and move it immediately prior to the exception check on line 384?

src/native/http_request_response.c
alexw91: Copy/paste error. Fixed :)

src/main/java/software/amazon/awssdk/crt/http/HttpConnection.java
alexw91: Fixed

PR #59, opened 2019-06-07
Updated dependencies, converted asserts, acquire -> calloc

justinboswell:


src/native/event_loop_group.c
```diff
@@ -36,11 +36,11 @@ jlong JNICALL
     Java_software_amazon_awssdk_crt_io_EventLoopGroup_eventLoopGroupNew(JNIEnv *env, jclass jni_elg, jint num_threads) {
     (void)jni_elg;
     struct aws_allocator *allocator = aws_jni_get_allocator();
-    struct aws_event_loop_group *elg = aws_mem_acquire(allocator, sizeof(struct aws_event_loop_group));
+    struct aws_event_loop_group *elg = aws_mem_calloc(allocator, 1, sizeof(struct aws_event_loop_group));
```
graebm: I don't know that it's a great idea to `x=calloc()` when it's immediately followed by `aws_x_init(*x)`.
If we start doing that everywhere, we're more likely to make mistakes while implementing init() functions, assuming that we're always dealing with pre-zeroed-out memory.

justinboswell: After discussion offline, we compromised: Language CRTs will get calloc treatment, but unit tests and core libs will not. The language CRTs have the least room for error, and we don't want to customer-surface these issues.

justinboswell:
Fixed https://github.com/awslabs/aws-c-http/issues/66

PR #72, opened 2019-07-19
Buffer Http Response Body in Native to reduce Java GC

alexw91:
*Issue #, if available:* N/A

*Description of changes:*
During Performance Stress Testing I found that Native's behavior of creating a Java `byte[]` per invocation of `s_on_incoming_body_fn()` (essentially on each TCP packet received) was creating Java Objects faster than Java's GC could keep up, eventually leading to `OutOfMemory` Exceptions.

This PR updates Native to internally buffer a configurable number of bytes (default set to 1 MB) of data before calling into Java so that many fewer Java `byte[]` Objects are created.

### DIFF
--- a/src/native/http_request_response.c
+++ b/src/native/http_request_response.c
@@ -236,7 +236,6 @@ static jobjectArray s_java_headers_array_from_native(
 
 static void aws_http_close_connection_with_reason(
     struct aws_http_stream *stream,
-    struct http_stream_callback_data *callback,
     char *reason) {
 
     struct aws_http_connection *conn = aws_http_stream_get_connection(stream);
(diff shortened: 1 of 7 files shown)


src/native/http_request_response.c
```diff
@@ -289,36 +387,28 @@ static void s_on_incoming_body_fn(
...
+    size_t total_window_increment = 0;
 
-    jobject jByteBuffer = aws_jni_byte_buffer_copy_from_cursor(env, data);
-    jint window_increment = (*env)->CallIntMethod(
-        env,
-        callback->java_crt_http_callback_handler,
-        s_crt_http_stream_handler.onResponseBody,
-        callback->java_http_stream,
-        jByteBuffer);
+    while (body_in_remaining.len > 0) {
```
justinboswell: If you are transferring less than the buffer size, how will that ever get published to Java?

alexw91: I force a java.onRespBody() publish when native.onStreamComplete() is called even if buffer is not full.

PR #74, opened 2019-07-26
 Add Java TLS Cipher API 

alexw91:
**Note: This PR builds on top of two other PR's, and will fail to build until both of these are merged:**
 - [awslabs/aws-crt-java#73](https://github.com/awslabs/aws-crt-java/pull/73 )
 - [https://github.com/awslabs/aws-c-io/pull/161](https://github.com/awslabs/aws-c-io/pull/161)

*Issue #, if available:* N/A

*Description of changes:*
Adds API's to configure TLS Cipher. 

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/io/TlsCipherPreference.java
+++ b/src/main/java/software/amazon/awssdk/crt/io/TlsCipherPreference.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2010-2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package software.amazon.awssdk.crt.io;
+
+public enum TlsCipherPreference {
+    /**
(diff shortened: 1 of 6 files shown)


src/main/java/software/amazon/awssdk/crt/io/TlsCipherPreference.java
```diff
@@ -0,0 +1,45 @@
...
+
+public enum TlsCipherPreference {
+    /**
+     * Use whatever the System Default Preference is. This is usually the best option, as it will be automatically
+     * updated as the underlying OS or platform changes.
+     */
+    TLS_CIPHER_SYSTEM_DEFAULT(0),
+
+    /**
+     * Contains Draft Hybrid TLS Ciphers: https://tools.ietf.org/html/draft-campagna-tls-bike-sike-hybrid
```
justinboswell: <3 docs

src/test/java/software/amazon/awssdk/crt/test/TlsContextOptionsTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+        options.setMinimumTlsVersion(TlsVersions.TLS_VER_SYS_DEFAULTS);
+
+        for (TlsCipherPreference pref: TlsCipherPreference.values()) {
+            if (TlsContextOptions.isCipherPreferenceSupported(pref)) {
+                options.setCipherPreference(pref);
+            }
+        }
+
+        boolean exceptionThrown = false;
```
justinboswell: You could use junit's built-in [exception testing](https://github.com/junit-team/junit4/wiki/exception-testing)

alexw91: I wanted to catch the Exception so that I can call `options.close();` and not leak memory.

justinboswell: That's fine, you might want to do a try with resources then.

PR #75, opened 2019-08-07
Use ConnectionManager for all Http Connections

alexw91:
*Issue #, if available:* N/A

*Description of changes:*
Use Native Http Connection Manager for all Connections.


src/native/http_connection_manager.c
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    uint16_t port = (uint16_t)jni_port;
+
+    int use_tls = (jni_tls_ctx != 0);
+
+    struct aws_tls_connection_options tls_conn_options = {0};
+
+    if (use_tls) {
+        aws_tls_connection_options_init_from_ctx(&tls_conn_options, tls_ctx);
```
JonathanHenson: tls_ctx isn't cheap and can technically be reused, if mtls isn't being used. Do we need to create a new ctx each time?

src/main/java/software/amazon/awssdk/crt/http/HttpConnectionPoolManager.java
```diff
@@ -0,0 +1,201 @@
...
+import software.amazon.awssdk.crt.io.ClientBootstrap;
+import software.amazon.awssdk.crt.io.SocketOptions;
+import software.amazon.awssdk.crt.io.TlsContext;
+
+/**
+ * Manages a Pool of Http Connections
+ */
+public class HttpConnectionPoolManager extends CrtResource {
+    public static final int DEFAULT_MAX_WINDOW_SIZE = Integer.MAX_VALUE;
+    public static final int DEFAULT_MAX_CONNECTIONS = 10;
```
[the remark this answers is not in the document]
alexw91: Fixed

src/native/http_connection_manager.c
alexw91: I just moved this code over from `http_connection.c` with very few changes. I don't think that change should block this PR

PR #78, opened 2019-08-21
Fixed missing ownership spotted while explaining how ownership works

justinboswell:


src/main/java/software/amazon/awssdk/crt/mqtt/MqttClient.java
```diff
@@ -37,6 +37,7 @@
      */
     public MqttClient() throws CrtRuntimeException {
         this(new ClientBootstrap(1), null);
```
justinboswell: No, sadly, ```own()``` cannot be called until AFTER ```this()```, I can't just do ```this(own(new ClientBootstrap(1)), null)``` as much as I want to :(

PR #89, opened 2019-09-12
Preallocate Request Buffers in Buffer Pool

alexw91:
*Issue #, if available:* N/A

*Description of changes:*
Creates a Pool of ByteBuffers when the ConnectionPool is created, so that Http Requests don't allocate and free a large buffer for every Request.


alexw91:
> Does this get around the GC pressure we were seeing?
Yes, instead of attempting to allocate buffers in Native when creating a Stream, we pre-allocate a Pool of Buffers when the ConnectionPool is created, and reuse those buffers for the lifetime of the ConnectionPool. Since we're reusing the buffers over multiple Streams and Connections, there's no more GC Pressure.

> I thought the guidance was to avoid direct buffers, and just use GC buffers?
That guidance was given since it's possible for Native Threads to allocate Java Buffers faster than JVM GC can keep up, leading to OutOfMemory errors. With this change, we only allocate Java Buffers in the Java Thread when creating the Connection Pool, so using Direct Buffers here should be fine.

PR #90, opened 2019-09-16
Add Sanity Check Test for Memory Leaks

alexw91:
*Issue #, if available:* N/A

*Description of changes:*
Adds a Sanity check that checks for Memory Leaks.


alexw91:
Made the leak detector generic so it could be used by other tests.

PR #92, opened 2019-09-24
Fix from guru review in brazil

justinboswell:

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/CRT.java
+++ b/src/main/java/software/amazon/awssdk/crt/CRT.java
@@ -110,7 +110,6 @@ public final class CRT {
                 // Copy from jar stream to temp file
                 Files.deleteIfExists(libTempPath);
                 Files.copy(in, libTempPath);
-                in.close();
             }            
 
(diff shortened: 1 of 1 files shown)


src/main/java/software/amazon/awssdk/crt/CRT.java
```diff
@@ -103,14 +103,15 @@ private static void loadLibraryFromJar() {
...
-            Files.copy(in, libTempPath);
-            in.close();
+            try (InputStream in = CRT.class.getResourceAsStream(libraryPath)) {
+                if (in == null) {
+                    throw new IOException("Unable to open library in jar for AWS CRT: " + libraryPath);
+                }
+                // Copy from jar stream to temp file
+                Files.deleteIfExists(libTempPath);
+                Files.copy(in, libTempPath);
+                in.close();
```
alexw91: `in.close()` will already be called due to the try with resources call won't it?

justinboswell: You are correct, fixed!

PR #94, opened 2019-09-26
Move CRTResource to a more traditional ref-counting approach

bretambrose:
The most critical changes are within CRTResource and its subclasses.

Additionally, resource creation (especially within the tests) have been converted to use try-with-resources almost exclusively.  Only when a resource needs to "leak" out of current scope do we not wrap it in a try block.  Those cases still require a manual close. 


src/main/java/software/amazon/awssdk/crt/CrtResource.java
```diff
@@ -16,96 +16,291 @@
...
+import software.amazon.awssdk.crt.Log;
 
 /**
  * This wraps a native pointer to an AWS Common Runtime resource. It also ensures
  * that the first time a resource is referenced, the CRT will be loaded and bound.
  */
-public class CrtResource implements AutoCloseable {
-    private static final ConcurrentHashMap<Long, String> NATIVE_RESOURCES = new ConcurrentHashMap<>();
+public abstract class CrtResource implements AutoCloseable {
+    private static final String NATIVE_DEBUG_PROPERTY_NAME = "aws.iot.sdk.debugnative";
```
alexw91: Should this have `iot` in the name?

bretambrose: Yeah, will switch "iot.sdk" -> "crt"

src/main/java/software/amazon/awssdk/crt/CrtResource.java
```diff
@@ -16,96 +16,291 @@
...
+     * Debug method to increment the current native object count.
+     */
+    private static void incrementNativeObjectCount() {
+        if (!debugNativeObjects) {
+            return;
+        }
+
+        lock.lock();
+        try {
+            ++resourceCount;
```
alexw91: Is there a reason you're not using `AtomicInteger`? Seems like it would be simpler and less error prone than manually managing a lock and int.

src/main/java/software/amazon/awssdk/crt/CrtResource.java
```diff
@@ -16,96 +16,291 @@
...
+    /**
+     * Debug/test method to wait for the CRTResource count to drop to zero.  Times out with an exception after
+     * a period of waiting.
+     */
+    public static void waitForNoResources() {
+        if (!debugNativeObjects) {
+            return;
+        }
+
+        lock.lock();
```
alexw91: How is this working?? By locking the `resourceCount` lock here, won't that stops other Threads from being able to decrement the int to zero?

bretambrose: The lock is released during await.

src/main/java/software/amazon/awssdk/crt/CrtResource.java
bretambrose: Not a Java expert, but Conditions need to be associated with a lock and it's a runtime error to signal a condition without the lock being held so AtomicInteger doesn't seem to buy anything.

src/native/event_loop_group.c
```diff
@@ -55,21 +67,57 @@ jlong JNICALL
...
+};
+
+static void s_event_loop_group_cleanup_completion_callback(void *user_data) {
+    struct event_loop_group_cleanup_callback_data *callback_data = user_data;
+
+    // Elg memory can be freed
+    struct aws_event_loop_group *elg = (struct aws_event_loop_group *)callback_data->elg_addr;
+    aws_mem_release(elg->allocator, elg);
+
+    AWS_LOGF_DEBUG(AWS_LS_IO_EVENT_LOOP, "Event Loop Shutdown Complete");
```
justinboswell: Should probably make a Java/JNI log subject instead, otherwise this might get confusing later.

codebuild/common-linux.sh
```diff
@@ -13,4 +13,4 @@ ENDPOINT=$(aws secretsmanager get-secret-value --secret-id "unit-test/endpoint"
 cd $CODEBUILD_SRC_DIR
 
 ulimit -c unlimited
-mvn -B test -DredirectTestOutputToFile=true -DreuseForks=false -Dendpoint=$ENDPOINT -Dcertificate=/tmp/certificate.pem -Dprivatekey=/tmp/privatekey.pem -Drootca=/tmp/AmazonRootCA1.pem
+mvn -B test -DredirectTestOutputToFile=true -DreuseForks=false -Dendpoint=$ENDPOINT -Dcertificate=/tmp/certificate.pem -Dprivatekey=/tmp/privatekey.pem -Drootca=/tmp/AmazonRootCA1.pem -Daws.crt.debugnative=true
```
justinboswell: Do you mean to check this in?

src/native/http_request_response.c
```diff
@@ -289,7 +289,10 @@ static int s_on_incoming_headers_fn(
     size_t num_headers,
     void *user_data) {
 
-    (void)block_type;
+    if (block_type != AWS_HTTP_HEADER_BLOCK_MAIN) {
```
justinboswell: What's this about? Could you add a comment if this is intentional?

src/native/logging.h
```diff
@@ -0,0 +1,24 @@
...
+ * permissions and limitations under the License.
+ */
+
+#ifndef AWS_JNI_LOGGING_H
+#define AWS_JNI_LOGGING_H
+
+/*******************************************************************************
+ * aws_cleanup_logging - cleans up the native logger; invoked on atexit
+ ******************************************************************************/
+void aws_cleanup_logging();
```
justinboswell: This should be ```aws_jni_cleanup_logging```

src/main/java/software/amazon/awssdk/crt/CrtResource.java
```diff
@@ -16,96 +16,291 @@
...
+     *
+     * A solution that avoids this complication while not changing the buffer/buffer pool properties dramatically would
+     * be welcomed.
+     */
+
+    /**
+     * Is this an actual native resource (false) or does it just track native resources and use the close/shutdown/referencing
+     * aspects (true)?
+     */
+    protected boolean isAlwaysNullResource() { return false; }
```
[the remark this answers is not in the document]
bretambrose: Yeah, that's better.

src/native/event_loop_group.c
bretambrose: Yeah, I wasn't sure what to do with this so that was the placeholder and then I forgot about it

codebuild/common-linux.sh
bretambrose: Yeah, it makes the resource count assertions actually do something in the tests

src/native/http_request_response.c
bretambrose: Uh, oops, lemme push that up to Java.

src/main/java/software/amazon/awssdk/crt/CrtResource.java
bretambrose: If an auto-close lock already existed I'd happily use it but I'm reluctant to add it just for this.

justinboswell: That's fair.

PR #98, opened 2019-10-07
Added support for configuring TLS cert/key/root from ByteBuffers

justinboswell:
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #99, opened 2019-10-09
PemUtils clean up

justinboswell:
Follow up from: https://github.com/awslabs/aws-crt-java/pull/98

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/io/TlsContextOptions.java
+++ b/src/main/java/software/amazon/awssdk/crt/io/TlsContextOptions.java
@@ -14,6 +14,8 @@
  */
 package software.amazon.awssdk.crt.io;
 
+import java.util.IllegalFormatException;
+
 import software.amazon.awssdk.crt.CrtResource;
 import software.amazon.awssdk.crt.CrtRuntimeException;
 import software.amazon.awssdk.crt.utils.PemUtils;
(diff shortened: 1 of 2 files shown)


src/main/java/software/amazon/awssdk/crt/io/TlsContextOptions.java
```diff
@@ -207,8 +215,10 @@ public void overrideDefaultTrustStoreFromPath(String caPath, String caFile) {
      * @param caRoot Buffer containing the root certificate chain. Must be in PEM format.
      */
     public void overrideDefaultTrustStore(String caRoot) throws IllegalArgumentException {
-        caRoot = PemUtils.removeInvalidPemChars(caRoot);
-        caRoot = PemUtils.cleanUpPem(caRoot);
+        String cleanCaRoot = PemUtils.cleanUpPem(caRoot);
+        if (!cleanCaRoot.equals(caRoot)) {
+            throw new IllegalArgumentException("Trust Store contains excess or illegal characters");
```
alexw91: CA files on CentOS host contain comments, this change will break them. 

See: 
 - https://github.com/awslabs/s2n/issues/1136
 - https://gist.github.com/alexw91/502f439535bf9886851d0ba9ce27043a

justinboswell: Gah, good to know. I'm tempted to just clean stuff up and call it good then. When we have a mechanism for logging from the Java side of the CRT, we can add a warning if we want.

PR #113, opened 2019-10-31
All uses of GetStringUTF8Chars now release the strings when done

justinboswell:

### DIFF
--- a/src/native/crt.c
+++ b/src/native/crt.c
@@ -565,7 +565,7 @@ struct aws_byte_cursor aws_jni_byte_cursor_from_jstring_acquire(JNIEnv *env, jst
 }
 
 void aws_jni_byte_cursor_from_jstring_release(JNIEnv *env, jstring str, struct aws_byte_cursor cur) {
-    (*env)->ReleaseStringUTFChars(env, str, (const char*)cur.ptr);
+    (*env)->ReleaseStringUTFChars(env, str, (const char *)cur.ptr);
 }
 
 struct aws_byte_cursor aws_jni_byte_cursor_from_direct_byte_buffer(JNIEnv *env, jobject byte_buffer) {
--- a/src/native/crt.c
+++ b/src/native/crt.c
@@ -559,9 +559,13 @@ jobject aws_jni_direct_byte_buffer_from_byte_buf(JNIEnv *env, const struct aws_b
     return aws_jni_direct_byte_buffer_from_raw_ptr(env, (void *)dst->buffer, (jlong)dst->capacity);
 }
 
-struct aws_byte_cursor aws_jni_byte_cursor_from_jstring(JNIEnv *env, jstring str) {
+struct aws_byte_cursor aws_jni_byte_cursor_from_jstring_acquire(JNIEnv *env, jstring str) {
     return aws_byte_cursor_from_array(
         (*env)->GetStringUTFChars(env, str, NULL), (size_t)(*env)->GetStringUTFLength(env, str));
+}
+
+void aws_jni_byte_cursor_from_jstring_release(JNIEnv *env, jstring str, struct aws_byte_cursor cur) {
+    (*env)->ReleaseStringUTFChars(env, str, (const char*)cur.ptr);
 }
 
 struct aws_byte_cursor aws_jni_byte_cursor_from_direct_byte_buffer(JNIEnv *env, jobject byte_buffer) {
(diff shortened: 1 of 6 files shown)


src/native/crt.c
```diff
@@ -582,7 +586,10 @@ struct aws_byte_cursor aws_jni_byte_cursor_from_direct_byte_buffer(JNIEnv *env,
 
 struct aws_string *aws_jni_new_string_from_jstring(JNIEnv *env, jstring str) {
     struct aws_allocator *allocator = aws_jni_get_allocator();
-    return aws_string_new_from_c_str(allocator, (*env)->GetStringUTFChars(env, str, NULL));
+    const char *str_chars = (*env)->GetStringUTFChars(env, str, NULL);
```
bdonlan: Need error handling.

src/native/http_connection_manager.c
```diff
@@ -113,7 +113,7 @@ JNIEXPORT jlong JNICALL Java_software_amazon_awssdk_crt_http_HttpClientConnectio
     }
 
     struct aws_allocator *allocator = aws_jni_get_allocator();
-    struct aws_byte_cursor endpoint = aws_jni_byte_cursor_from_jstring(env, jni_endpoint);
+    struct aws_byte_cursor endpoint = aws_jni_byte_cursor_from_jstring_acquire(env, jni_endpoint);
 
     if (jni_port <= 0 || 65535 < jni_port) {
         aws_jni_throw_runtime_exception(env, "Port must be between 1 and 65535");
```
bdonlan: Memory leak in error paths.

src/native/mqtt_connection.c
```diff
@@ -390,7 +395,7 @@ void JNICALL Java_software_amazon_awssdk_crt_mqtt_MqttClientConnection_mqttClien
         return;
     }
 
-    struct aws_byte_cursor endpoint = aws_jni_byte_cursor_from_jstring(env, jni_endpoint);
+    struct aws_byte_cursor endpoint = aws_jni_byte_cursor_from_jstring_acquire(env, jni_endpoint);
     uint16_t port = jni_port;
     if (!port) {
         aws_jni_throw_runtime_exception(
```
bdonlan: error path leak

src/native/mqtt_connection.c
```diff
@@ -672,7 +679,8 @@ jshort JNICALL Java_software_amazon_awssdk_crt_mqtt_MqttClientConnection_mqttCli
     /* pin the topic so it lives as long as the unsub request */
     jni_topic = (*env)->NewGlobalRef(env, jni_topic);
     unsub_ack->jni_objects[0] = jni_topic;
-    struct aws_byte_cursor topic = aws_jni_byte_cursor_from_jstring(env, jni_topic);
+    struct aws_byte_cursor topic = aws_jni_byte_cursor_from_jstring_acquire(env, jni_topic);
+    unsub_ack->strings[0] = (const char*)topic.ptr; /* string will be cleaned up in callback after unsub */
 
     uint16_t msg_id =
         aws_mqtt_client_connection_unsubscribe(connection->client_connection, &topic, s_on_op_complete, unsub_ack);
```
bdonlan: do we free the string in error paths?

src/native/mqtt_connection.c
```diff
@@ -672,7 +679,8 @@ jshort JNICALL Java_software_amazon_awssdk_crt_mqtt_MqttClientConnection_mqttCli
     /* pin the topic so it lives as long as the unsub request */
     jni_topic = (*env)->NewGlobalRef(env, jni_topic);
     unsub_ack->jni_objects[0] = jni_topic;
-    struct aws_byte_cursor topic = aws_jni_byte_cursor_from_jstring(env, jni_topic);
+    struct aws_byte_cursor topic = aws_jni_byte_cursor_from_jstring_acquire(env, jni_topic);
```
bdonlan: for clarity I would recommend copying to a C-style string instead of splitting up the JNI acquire/release.

src/native/crt.c
justinboswell: For what error? If the string returned is null, it will be returned that way. Nothing else in here can fail?

src/native/mqtt_connection.c
justinboswell: Yes, cleaning up the async_callback will release the string and the pinned reference.

src/native/mqtt_connection.c
justinboswell: That would create copies, which are then copied again.

src/native/crt.c
bdonlan: Does both `aws_string_new_from_c_str` and `ReleaseStringUTFChars` accept null args?

justinboswell: yes.

PR #129, opened 2019-11-15
Topic name should come from the received message

bretambrose:
Stop caching the topic of a subscription in the callback handler so that wildcard subscribes don't appear to be received message topics.

### DIFF
--- a/src/native/crt.c
+++ b/src/native/crt.c
@@ -302,3 +302,15 @@ jlong JNICALL Java_software_amazon_awssdk_crt_CRT_awsNativeMemory(JNIEnv *env, j
     }
     return allocated;
 }
+
+jstring aws_jni_string_from_cursor(JNIEnv *env, const struct aws_byte_cursor *native_data) {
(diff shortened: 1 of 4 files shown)


src/native/crt.c
```diff
@@ -302,3 +302,15 @@ jlong JNICALL Java_software_amazon_awssdk_crt_CRT_awsNativeMemory(JNIEnv *env, j
...
     return allocated;
 }
+
+jstring aws_jni_string_from_cursor(JNIEnv *env, const struct aws_byte_cursor *native_data) {
+    struct aws_string *string = aws_string_new_from_array(aws_jni_get_allocator(), native_data->ptr, native_data->len);
+    if (string == NULL) {
+        return NULL;
+    }
+
+    jstring java_string = (*env)->NewStringUTF(env, (const char *)string->bytes);
```
JonathanHenson: we have a function for this now
aws_string_c_str

bretambrose: Since it's a cursor, there's no guarantee that it's null-terminated.

bretambrose: NM, I understand now

PR #135, opened 2019-11-26
Auth bindings

bretambrose:
* Credentials
 * Credentials Providers
 * Signing
 * Rework jni id caching


src/main/java/software/amazon/awssdk/crt/auth/signing/AwsSigningConfig.java
```diff
@@ -0,0 +1,98 @@
...
+        AwsSigningAlgorithm(int nativeValue) {
+            this.nativeValue = nativeValue;
+        }
+
+        int getNativeValue() { return nativeValue; }
+
+        int nativeValue;
+    }
+
+    int signingAlgorithm = AwsSigningAlgorithm.SIGV4_HEADER.getNativeValue();
```
JonathanHenson: why are these spaced out and why aren't they private?

src/native/aws_signing.c
```diff
@@ -0,0 +1,325 @@
...
+
+static jobject s_create_signed_java_http_request(
+    JNIEnv *env,
+    struct aws_http_message *native_request,
+    jobject java_original_request) {
+    jstring jni_uri = NULL;
+    jobjectArray jni_headers = NULL;
+    jobject http_request = NULL;
+
+    jstring jni_method = (*env)->GetObjectField(env, java_original_request, http_request_properties.method_field_id);
```
JonathanHenson: since jni_method and jni_uri are seated on line 104, should we be calling DeleteLocalRef on those as well?

src/native/aws_signing.c
```diff
@@ -0,0 +1,325 @@
...
+    JNIEnv *env,
+    struct s_aws_sign_request_callback_data *callback_data,
+    int error_code) {
+
+    jint jni_error_code = error_code;
+    struct aws_byte_cursor error_cursor = aws_byte_cursor_from_c_str(aws_error_name(error_code));
+    jstring jni_error_string = aws_jni_string_from_cursor(env, &error_cursor);
+    AWS_FATAL_ASSERT(jni_error_string);
+
+    jobject crt_exception = (*env)->NewObject(
```
JonathanHenson: do we need to delete the returned reference?

src/native/aws_signing.c
```diff
@@ -0,0 +1,325 @@
...
+
+    bool result = (*env)->CallBooleanMethod(
+        env, callback_data->java_sign_param_predicate, predicate_properties.test_method_id, (jobject)parameter_name);
+
+    (*env)->DeleteLocalRef(env, parameter_name);
+
+    return result;
+}
+
+static int s_build_signing_config(
```
JonathanHenson: verify ref semantics here.

bretambrose: The result is passed into the signing function in the direction of Java -> Native.  When combined with the fact that the signing config is copied internally at the start of signing, the automatic local ref destruction that happens on return up the call stack exactly matches the reference life time we need.

src/native/aws_signing.c
bretambrose: Yes

src/native/aws_signing.c
bretambrose: Yes

PR #138, opened 2019-12-10
Ez config

bretambrose:
Support work for the mqtt builder.  Actual builder is going in the v2 SDK.


src/main/java/software/amazon/awssdk/crt/mqtt/MqttClientConnection.java
```diff
@@ -106,6 +117,159 @@ protected boolean canReleaseReferencesImmediately() {
...
+        return userConnectionCallbacks;
+    }
+
+    /**
+     * Configures the client_id to use with this connection
+     *
+     * @param clientId The client id for this connection. Needs to be unique across
+     *                  all devices/clients.
+     */
+    public void setClientId(String clientId) {
```
justinboswell: The way we've done this in the other builders was to  have these params be part of the constructor, and have the builder invoke the constructor once it had all the params. By making this public, it makes the API to this class VERY confusing. I would suggest making this private or something if it must be done this way, but I'd rather use a constructor like we do elsewhere.

I'd much rather keep the config object and have the builder close it when it's done creating a connection

bretambrose: Done

src/native/mqtt_connection.c
```diff
@@ -841,6 +853,114 @@ JNIEXPORT void JNICALL Java_software_amazon_awssdk_crt_mqtt_MqttClientConnection
     aws_jni_byte_cursor_from_jstring_release(env, jni_pass, password);
 }
 
+///////
```
graebm: trivial: ////////// ??????

PR #139, opened 2019-12-10
Added ability to convert from aws-c-http http request object to java …

JonathanHenson:
…http request, and httprequest is now mutable in java land.

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/auth/signing/AwsSigner.java
+++ b/src/main/java/software/amazon/awssdk/crt/auth/signing/AwsSigner.java
@@ -27,8 +27,7 @@ public class AwsSigner {
 
         try {
             //JNI is much easier to deal with primitives, so go ahead and hoist the headers into an array.
-            HttpHeader[] tempHeadersToSign = new HttpHeader[request.getHeaders().size()];
-            tempHeadersToSign = request.getHeaders().toArray(tempHeadersToSign);
+            HttpHeader[] tempHeadersToSign = request.getHeaders().toArray(new HttpHeader[]{});
             awsSignerSignRequest(request, tempHeadersToSign, config, future);
         } catch (Exception e) {
             future.completeExceptionally(e);
--- a/src/main/java/software/amazon/awssdk/crt/auth/signing/AwsSigner.java
+++ b/src/main/java/software/amazon/awssdk/crt/auth/signing/AwsSigner.java
@@ -27,8 +27,7 @@ public class AwsSigner {
 
         try {
             //JNI is much easier to deal with primitives, so go ahead and hoist the headers into an array.
-            HttpHeader[] tempHeadersToSign = request.getHeaders().toArray(new HttpHeader[]{});
-            awsSignerSignRequest(request, tempHeadersToSign, config, future);
+            awsSignerSignRequest(request, request.getHeadersAsArray(), config, future);
         } catch (Exception e) {
             future.completeExceptionally(e);
         }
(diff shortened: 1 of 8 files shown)


src/main/java/software/amazon/awssdk/crt/http/HttpRequest.java
```diff
@@ -56,10 +61,53 @@ public String getEncodedPath() {
...
 
+    public void addHeader(final HttpHeader header) {
+        headers.add(header);
+    }
+
+    public void addHeader(final String headerName, final String headerValue) {
+        headers.add(new HttpHeader(headerName, headerValue));
+    }
+
+    public void addHeaders(final HttpHeader headers) {
```
ColdenCullen: Should this be an array or something?

src/main/java/software/amazon/awssdk/crt/auth/signing/AwsSigner.java
```diff
@@ -25,7 +26,10 @@
         CompletableFuture<HttpRequest> future = new CompletableFuture<HttpRequest>();
 
         try {
-            awsSignerSignRequest(request, config, future);
+            //JNI is much easier to deal with primitives, so go ahead and hoist the headers into an array.
+            HttpHeader[] tempHeadersToSign = new HttpHeader[request.getHeaders().size()];
+            tempHeadersToSign = request.getHeaders().toArray(tempHeadersToSign);
+            awsSignerSignRequest(request, tempHeadersToSign, config, future);
```
graebm: trivial: worth adding a HttpRequest.getHeadersArray() or a helper function somewhere?

src/main/java/software/amazon/awssdk/crt/http/HttpRequest.java
```diff
@@ -56,10 +61,52 @@ public String getEncodedPath() {
...
+        this.headers.add(headers);
+    }
+
+    public void addQueryParam(final String paramName, final String paramValue, boolean urlEncode) {
```
graebm: nuke this function if we don't need it now. no other language has this

src/main/java/software/amazon/awssdk/crt/http/HttpRequest.java
JonathanHenson: fixed.

src/native/http_request_utils.h
```diff
@@ -36,6 +36,8 @@ struct aws_http_message *aws_http_request_new_from_java_http_request(
     jobjectArray jni_headers,
     jobject jni_body_stream);
 
+jobject aws_java_http_request_from_native(JNIEnv *env, struct aws_http_message *message);
```
graebm: if this fails, a java exception has been set, right? Add a comment to that effect
Looking in the function, it does seem that the only things that can fail are the things that create java objects

src/native/http_request_utils.h
```diff
@@ -36,6 +36,8 @@ struct aws_http_message *aws_http_request_new_from_java_http_request(
...
+
```
graebm: We'll also need a function that modifies an existing native aws_http_request to match the modifications made to the Java HtttpRequest? Can you add that in this PR please?

src/main/java/software/amazon/awssdk/crt/auth/signing/AwsSigner.java
JonathanHenson: i can do that

src/main/java/software/amazon/awssdk/crt/http/HttpRequest.java
JonathanHenson: removed

PR #140, opened 2019-12-12
client-bootstrap-cb

JonathanHenson:
Expose client bootstrap callback in Java bindings.


src/main/java/software/amazon/awssdk/crt/io/ClientBootstrap.java
```diff
@@ -14,15 +14,36 @@
...
+    private final CompletableFuture<Void> shutdownComplete = new CompletableFuture<>();
+
+    /**
+     * Creates a new ClientBootstrap. Most applications will only ever need one instance of this.
+     * @param numThreads The number of Threads to use in the EventLoop
+     * @throws CrtRuntimeException If the system is unable to allocate space for a native client bootstrap object
+     */
+    public ClientBootstrap(int numThreads) throws CrtRuntimeException {
+        try(EventLoopGroup elg = new EventLoopGroup(numThreads);
+            HostResolver hr = new HostResolver(elg)) {
```
graebm: But I removed all the automagic creation of resources from Python just the other week?
On the suspicion that it would bite users in the butt if we're automatically creating big heavy resources on their behalf without them realizing it?

bretambrose: Looks like I removed all the usages of this overload but forgot to delete it afterwards

graebm: But it looks like NEW code in this PR. Maybe a bad merge brought it back to life

JonathanHenson: It did, I'll remove it.

JonathanHenson: it was a confusing merge.

JonathanHenson: I have no idea what you're talking about.

PR #143, opened 2019-12-28
check return codes in tls_context_options.c

graebm:
A customer had typed their cert path in wrong, and we didn't detect it until way later when TLS negotiation failed. Turns out there were a lot of return codes we were never checking in this .c file.

There are **lots** of other native/*.c files that need an error-handling pass like this. I got overwhelmed, so this PR just:
1) tls_context_options.c
2) host_resolver.c

### DIFF
--- a/src/native/host_resolver.c
+++ b/src/native/host_resolver.c
@@ -52,7 +52,13 @@ JNIEXPORT jlong JNICALL Java_software_amazon_awssdk_crt_io_HostResolver_hostReso
     }
 
     struct aws_host_resolver *resolver = aws_mem_calloc(allocator, 1, sizeof(struct aws_host_resolver));
-    aws_host_resolver_init_default(resolver, allocator, (size_t)max_entries, el_group);
+    AWS_FATAL_ASSERT(resolver);
+
+    if (aws_host_resolver_init_default(resolver, allocator, (size_t)max_entries, el_group)) {
+        aws_jni_throw_runtime_exception(env, "aws_host_resolver_init_default failed");
+        aws_mem_release(allocator, resolver);
+        return (jlong)NULL;
+    }
 
     return (jlong)resolver;
 }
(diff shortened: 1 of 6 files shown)


src/native/host_resolver.c
```diff
@@ -52,7 +52,13 @@ JNIEXPORT jlong JNICALL Java_software_amazon_awssdk_crt_io_HostResolver_hostReso
     }
 
     struct aws_host_resolver *resolver = aws_mem_calloc(allocator, 1, sizeof(struct aws_host_resolver));
-    aws_host_resolver_init_default(resolver, allocator, (size_t)max_entries, el_group);
+    AWS_FATAL_ASSERT(resolver);
+
+    if (aws_host_resolver_init_default(resolver, allocator, (size_t)max_entries, el_group)) {
+        aws_jni_throw_runtime_exception(env, "aws_host_resolver_init_default failed");
```
JonathanHenson: shall we throw in an aws_error_debug_str() or whatever here?

src/native/tls_context_options.c
```diff
@@ -93,29 +93,56 @@ jlong JNICALL Java_software_amazon_awssdk_crt_io_TlsContextOptions_tlsContextOpt
...
+            aws_jni_throw_runtime_exception(env, "failed to get privateKey string");
+            goto on_error;
+        }
 
         struct aws_byte_cursor cert_cursor = aws_byte_cursor_from_string(tls->certificate);
         struct aws_byte_cursor key_cursor = aws_byte_cursor_from_string(tls->private_key);
 
-        aws_tls_ctx_options_init_client_mtls(&tls->options, aws_jni_get_allocator(), &cert_cursor, &key_cursor);
+        if (aws_tls_ctx_options_init_client_mtls(&tls->options, allocator, &cert_cursor, &key_cursor)) {
+            aws_jni_throw_runtime_exception(env, "aws_tls_ctx_options_init_client_mtls failed");
```
JonathanHenson: might be nice to include the error strings here

src/native/host_resolver.c
graebm: `aws_jni_throw_runtime_exception()` already prints out the name/debug_str of `aws_last_error()`

PR #144, opened 2020-01-03
Check java.io.tmpdir

graebm:
We had a customer whose "java.io.tmpdir" was set to a non-existent directory. Resulting exception looked like:

`Unable to initialize AWS CRT: java.nio.file.NoSuchFileException: /local/home/...redacted.../tmp/AWSCRT_15768129594383460159480481437974libaws-crt-jni.so`

I can't find much official documentation on "java.io.tmpdir" beyond [this](https://docs.oracle.com/javase/7/docs/api/java/lang/System.html#getProperties()), which basically just says "yes this is an official property that will be defined". I would hesitate to try and create the directory if it doesn't already exist. Instead, I'm throwing more explicit errors if the tmpdir is invalid. So the user will instead see:

`Unable to unpack AWS CRT lib: java.nio.file.NotDirectoryException: java.io.tmpdir="/local/home/...redacted.../tmp"`

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/CRT.java
+++ b/src/main/java/software/amazon/awssdk/crt/CRT.java
@@ -95,14 +95,27 @@ public final class CRT {
 
     private static void loadLibraryFromJar() {
         try {
-            // Check tmpdir and give explicit errors
-            Path tmpdirPath = Paths.get(System.getProperty("java.io.tmpdir"));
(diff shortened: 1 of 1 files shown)


JonathanHenson:
actually, I talked to ryan schmitt yesterday (the best java expert I know), and he advised us to just create it if it doesn't exist.

PR #146, opened 2020-01-07
Weak refs

bretambrose:
Converts several strong (global) refs to weak global refs, with appropriate conversion and checking before use during callbacks.

This is required before we attempt to integrate the Janitor approach to cleaning up CrtResource objects (since these global refs would prevent the ref count from dropping to zero on the referenced objects).

Conversion criteria was:

1. A global ref to a CrtResource-extending Java object, embedded in a C struct.
1. The C struct that holds the ref does not have a "natural" lifetime and thus lives indefinitely without a close()/final decref on the Java object.  

To motivate (2), http_stream_callback_data.java_http_stream meets the first condition but not the second, and it's reasonable to not bother making it weak because the C object will go away as soon as the stream complete callback is finished.  If we made it weak it could potentially clean up a little earlier.  event_loop_group_cleanup_callback_data.event_loop_group was similar; it's active only during the shutdown period of an event loop group.

### DIFF
--- a/src/native/client_bootstrap.c
+++ b/src/native/client_bootstrap.c
@@ -38,7 +38,7 @@
 
 struct shutdown_callback_data {
     JavaVM *jvm;
-    jobject java_client_bootstrap;
+    jweak weak_java_client_bootstrap;
 };
 
 static void s_shutdown_callback_data_destroy(JNIEnv *env, struct shutdown_callback_data *callback_data) {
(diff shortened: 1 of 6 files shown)


src/native/client_bootstrap.c
```diff
@@ -38,16 +38,16 @@
 
 struct shutdown_callback_data {
     JavaVM *jvm;
-    jobject java_client_bootstrap;
+    jweak weak_java_client_bootstrap;
```
justinboswell: the ```weak_``` feels a bit overkill

justinboswell: Maybe just like ```client_bootstrap_ref``` or just ```java_client_bootstrap```?

bretambrose: I can return it to the way it was at this point.  Was more to let the compile catch any place I missed.

PR #158, opened 2020-02-20
Added ARM CI

justinboswell:


.builder/jdk.py
```diff
@@ -0,0 +1,120 @@
...
+                for paths in required_files:
+                    for path in paths:
+                        full_path = os.path.join(prefix, path)
+                        if glob.glob(full_path, recursive=True):
+                            found += 1
+                            break
+
+                if found >= len(required_files):
+                    print('Found existing JDK8 at {}'.format(prefix))
+                    env.variables['java_home'] = self.path = prefix
```
graebm: trivial: double-assignment is weird to me

.builder/jdk.py
```diff
@@ -0,0 +1,120 @@
...
+                    self.installed = True
+                    return
+
+        target = '{}-{}'.format(env.spec.target, env.spec.arch)
```
graebm: trivial: You meant to fall through here if we looked for, but didn't find, an existing local installation? In any case, add a comment where the code resumes explaining the circumstances

`# Install JDK. We didn't find (or we're not using) an existing installation.`

.builder/jdk.py
```diff
@@ -0,0 +1,120 @@
...
+        if target not in URLs:
+            print('WARNING: No pre-built binaries for {} are available'.format(target))
+            return
```
graebm: trivial: "...prebuillt JDK binaries..."

Also, why not raise an exception? Is this just how `Builder.Import.instal()` is supposed to work?

CMakeLists.txt
```diff
@@ -111,10 +115,20 @@ file(GLOB CRT_JAVA_SRC
...
 set(TARGET_LIB_DIR "lib/${TARGET_OS}/${TARGET_ARCH}")
 
 # shared lib that contains the CRT and JNI bindings, to be loaded by java
 add_library(${PROJECT_NAME} SHARED ${CRT_JAVA_HEADERS} ${CRT_JAVA_SRC})
+aws_use_package(aws-c-http)
+aws_use_package(aws-c-mqtt)
+aws_use_package(aws-c-cal)
+aws_use_package(aws-c-auth)
+aws_use_package(aws-c-common)
+aws_use_package(aws-c-io)
```
graebm: What about aws-c-compression?

justinboswell: oh, this was an overreaction to not finding libcrypto, should revert

.builder/jdk.py
justinboswell: There's a return. Basically, there's no way we CAN install the JDK, so I've made it say "please install JDK8 or greater and set JAVA_HOME"

.builder/jdk.py
justinboswell: Done.

PR #171, opened 2020-03-17
Fixed deployment/CI versions

justinboswell:


.github/workflows/ci.yml
```diff
@@ -7,7 +7,8 @@ on:
       - '!master'
 
 env:
-  BUILDER_VERSION: v0.5.3
+  BUILDER_VERSION: more-fixes
+  BUILDER_SOURCE: channels
```
graebm: does this need to change before merge?

.builder/jdk.py
```diff
@@ -49,8 +49,10 @@ def install(self, env):
 
         sh = env.shell
 
+        target = '{}-{}'.format(env.spec.target, env.spec.arch)
+
         # If this is a local build, check the local machine
-        if not env.toolchain.cross_compile:
+        if not env.toolchain.cross_compile or target not in URLs:
```
graebm: do you still need jdk.py here? didn't you just move it into aws-crt-builder?

.github/workflows/ci.yml
justinboswell: yes

.builder/jdk.py
justinboswell: yes, it should have been deleted, will fix

PR #172, opened 2020-03-18
Update http api to use activate. It's hidden behind the make request …

JonathanHenson:
…function though since completeable futures are used, so no need for any code changes from java users. Also added the option to reenable tcp back pressure on the http connection manager.


src/native/aws_signing.c
```diff
@@ -185,6 +188,10 @@ static void s_aws_signing_complete(struct aws_signing_result *result, int error_
 
     (*env)->DeleteLocalRef(env, java_signed_request);
 
+    /* I have no idea what we should do here... but the JVM really doesn't like us NOT calling this function after
+       we cross the barrier. */
+    (*env)->ExceptionCheck(env);
```
graebm: Almost everywhere else we do this it's wrapped in AWS_FATAL_ASSERT()
should probably do the same here for consistency if nothing else

src/native/aws_signing.c
```diff
@@ -204,6 +211,10 @@ static bool s_should_sign_param(const struct aws_byte_cursor *name, void *user_d
 
     (*env)->DeleteLocalRef(env, parameter_name);
 
+    if ((*env)->ExceptionCheck(env)) {
+        return false;
+    }
```
graebm: don't silently swallow this exception, probably just AWS_FATAL_ASSERT() like we do elsewhere

src/main/java/software/amazon/awssdk/crt/http/HttpClientConnectionManagerOptions.java
```diff
@@ -157,5 +158,18 @@ public HttpClientConnectionManagerOptions withProxyOptions(HttpProxyOptions prox
...
+
+    /**
+     * If true, the read back pressure mechanism is enabled for connections.
+     */
+    public boolean isReadBackPressureEnabled() { return enableReadBackPressure; }
+
+    /**
+     * If true, the read back pressure mechanism is enabled for connections. Defaults to false
+     */
+    public HttpClientConnectionManagerOptions withReadBackPressureEnabled(boolean enableReadBackPressure) {
```
graebm: Did you see my comment about this on crt-cpp?
https://github.com/awslabs/aws-crt-cpp/pull/104#discussion_r394011329

src/native/http_request_response.c
```diff
@@ -337,11 +308,6 @@ JNIEXPORT jobject JNICALL Java_software_amazon_awssdk_crt_http_HttpClientConnect
     struct aws_http_stream *native_stream = NULL;
     jobject jHttpStream = NULL;
 
-    // There's a Data Race between this thread writing to callback_data->java_http_stream and the EventLoop thread
-    // reading callback_data->java_http_stream when calling the callbacks, add a lock so that
-    // callbacks will wait for setup to complete.
-    aws_mutex_lock(&callback_data->setup_lock);
```
graebm: yay! so long hacks!

src/native/http_request_response.c
```diff
@@ -110,23 +107,12 @@ static struct http_stream_callback_data *http_stream_callback_alloc(JNIEnv *env,
     (void)jvmresult;
     AWS_FATAL_ASSERT(jvmresult == 0);
 
-    aws_mutex_init(&callback->setup_lock);
-
     callback->java_http_response_stream_handler = (*env)->NewGlobalRef(env, java_callback_handler);
```
graebm: every CRT has some mechanism that keeps the stream object alive from `creation -> stream_complete`
It needs to change to keep the stream alive from `activate -> stream_complete` instead

I'm not sure exactly how it works in java. you'll need to figure it out

src/test/java/software/amazon/awssdk/crt/test/HttpRequestResponseTest.java
```diff
@@ -357,5 +357,4 @@ public void testHttpUpload() throws Exception {
         Assert.assertNotNull("Response Body did not contain \"data\" JSON key:\n" + body, echoedBody);
         Assert.assertEquals(bodyToSend, echoedBody);
     }
-
 }
```
graebm: Add a unit test for creating a request without activating it. make sure things clean up properly

src/main/java/software/amazon/awssdk/crt/http/HttpClientConnection.java
```diff
@@ -63,9 +63,12 @@ protected HttpClientConnection(HttpClientConnectionManager manager, long connect
                     streamHandler);
                 if (stream == null || stream.isNull()) {
                     streamFuture.completeExceptionally(new RuntimeException("HttpStream creation failed"));
+                    return streamFuture;
                 }
+
+                stream.activate();
                 streamFuture.complete(stream);
```
graebm: How is automatically calling activate() any different from not having activate() at all?

Explicit activate() is to solve the issue where people write code where they do something with the returned stream (ex: store it as a member variable somewhere), then they expect to be able to access the stream via that member variable from the callbacks.

If we automatically call activate() then yes, we've solved our internal implementation issues, but anyone who wraps our stuff in their own system will still have these issues.

On a side note, I have no idea why this function returns a CompletableFuture<HttpStream> when it could just return an HttpStream.

JonathanHenson: not done any more, and no longer returning a future

PR #189, opened 2020-05-05
Rhel 5 workarounds

bretambrose:
* fixes for rhel5 build


src/native/crt.c
```diff
@@ -186,7 +186,7 @@ void s_detach_jvm_from_thread(void *user_data) {
 }
 
 JNIEnv *aws_jni_get_thread_env(JavaVM *jvm) {
-    JNIEnv *env = NULL;
+    void *env = NULL;
```
justinboswell, line 189: Let's rename this var for ```void*``` and keep env the same, so android will never see it differently, same as the technique below.

PR #198, opened 2020-05-12
Added Process class for accessing ulimits.

JonathanHenson:


src/main/java/software/amazon/awssdk/crt/Process.java
```diff
@@ -0,0 +1,74 @@
...
+
+package software.amazon.awssdk.crt;
+
+import static software.amazon.awssdk.crt.CRT.awsErrorString;
+import static software.amazon.awssdk.crt.CRT.awsLastError;
+
+/**
+ * Encapsulates Process information and manipulation of process level operations.
+ */
+public class Process {
```
justinboswell, line 24: final?

JonathanHenson: not quite yet, but likely. Also I didn't see this til after I clicked the merge button.

PR #200, opened 2020-05-14
X509 Proxy support

bretambrose:
* Adds proxy configuration to the X509 credentials provider


src/main/java/software/amazon/awssdk/crt/auth/credentials/X509CredentialsProvider.java
```diff
@@ -126,7 +142,39 @@ private X509CredentialsProvider(X509CredentialsProviderBuilder builder) {
...
+        }
+
+        long nativeHandle = x509CredentialsProviderNew(
+                this,
+                clientBootstrap.getNativeHandle(),
+                tlsContext.getNativeHandle(),
+                thingName.getBytes(UTF8),
+                roleAlias.getBytes(UTF8),
+                endpoint.getBytes(UTF8),
+                proxyHost,
```
JonathanHenson, line 172: should this be serialized via getBytes() as well?

src/main/java/software/amazon/awssdk/crt/auth/credentials/X509CredentialsProvider.java
```diff
@@ -126,7 +142,39 @@ private X509CredentialsProvider(X509CredentialsProviderBuilder builder) {
...
+                proxyPort,
+                proxyTlsContextHandle,
+                proxyAuthorizationType,
+                proxyAuthorizationUsername,
```
JonathanHenson, line 176: what about these?

src/main/java/software/amazon/awssdk/crt/auth/credentials/X509CredentialsProvider.java
bretambrose: We could; it would lead to a bit more refactor in the connection manager

PR #207, opened 2020-06-25
Add MqttClientConnection as an interface for usage in IoTJobsClient

### LINKED ISSUE #124: Add interfaces to make it possible to use different MQTT clients
The current concrete implementation of MqttClientConnection makes it so that there isn't a way to use a different MQTT client. A customer requested an example Greengrass Lambda function that uses this SDK to handle jobs on Greengrass - aws-samples/aws-greengrass-lambda-functions#792 - and if there was an interface I could use to replace the implementation with a Greengrass compatible one I could create the example.

I'll work on a PR for this.

Originally this issue was opened in the v2 Java Device SDK repo - https://github.com/awslabs/aws-iot-device-sdk-java-v2/issues/24

MikeDombo:
*Issue #, if available:*

*Description of changes:*
This change creates the interfaces `MqttPublishInterface`, `MqttSubscribeInterface`, and `MqttSubscribeHandlerInterface` (if you have a preferred name, let me know and I'll change it) which are implemented by `MqttClientConnection`. The purpose of this change is to enable a change to the `IotJobsClient` to accept the interface instead of the concrete implementation. 

Should this PR be accepted, it should be followed by a change in `AWS-IoT-Device-SDK-v2` for the `IotJobsClient` to use this added interface instead of the concrete implementation.

This PR would resolve #124.


MikeDombo:
Hi Brett, thanks for the review.

What would you think about having maybe 3 interfaces so that the various use cases can have minimal requirements on the client they are given? 

1 interface for pubsub with handler, 1 for pubsub without handler and with global handler, and then 1 which is the same as the concrete implementation?

PR #213, opened 2020-07-23
New windowing + larger io messages by default

bretambrose:


src/native/crt.c
```diff
@@ -234,6 +237,7 @@ void JNICALL Java_software_amazon_awssdk_crt_CRT_awsCrtInit(
     }
 
     g_memory_tracing = jni_memtrace;
+    g_aws_channel_max_fragment_size = KB_256;
```
JonathanHenson, line 240: could you just put a comment here why we're doing this to be nice to our future selves?

PR #222, opened 2020-08-19
Gradle JAR separation

justinboswell:
*Issue #, if available:*

*Description of changes:*


common.gradle.kts
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+extra["getHostOs"] = fun(): String {
```
kiiadi, line 6: might be nicer to include these in `buildSrc` https://docs.gradle.org/current/userguide/organizing_gradle_projects.html#sec:build_sources which would allow you to reference the functions directly:

instead of having to do this:

```kt
val getHostOs = extra["getHostOs"] as () -> String
val getHostArch = extra["getHostArch"] as () -> String
```

See example here : https://github.com/aws/aws-toolkit-jetbrains/blob/master/buildSrc/src/software/aws/toolkits/gradle/IdeVersions.kt

That is then referenced in a `build.gradle.kts` here: https://github.com/aws/aws-toolkit-jetbrains/blob/108c346cb153aa352b00eb539f16fa9607ebd785/jetbrains-ultimate/build.gradle.kts#L18

justinboswell: You are absolutely right. I started with that, it didn't work right away, so I hacked it up this way, with intent to fix it. This is very much WIP :) That said, this example is much easier to follow than the docs, so maybe I'll figured out what I screwed up last time!

PR #225, opened 2020-08-28
Static elg hr

bretambrose:
Adds support for static fallback event loop group and host resolver.  This allows us to provide a safe alternative for users that don't want to manage CRT resources with the associated ref-juggling.

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/io/ClientBootstrap.java
+++ b/src/main/java/software/amazon/awssdk/crt/io/ClientBootstrap.java
@@ -25,6 +25,14 @@ public final class ClientBootstrap extends CrtResource {
      * or if the system is unable to allocate space for a native client bootstrap object
      */
     public ClientBootstrap(EventLoopGroup elg, HostResolver hr) throws CrtRuntimeException {
+        if (elg == null) {
+            elg = EventLoopGroup.getOrCreateStaticDefault();
+        }
+
+        if (hr == null) {
+            hr = HostResolver.getOrCreateStaticDefault();
+        }
+
         acquireNativeHandle(clientBootstrapNew(this, elg.getNativeHandle(), hr.getNativeHandle()));
 
         // Order is likely important here
--- a/src/main/java/software/amazon/awssdk/crt/io/EventLoopGroup.java
+++ b/src/main/java/software/amazon/awssdk/crt/io/EventLoopGroup.java
@@ -73,7 +73,7 @@ public final class EventLoopGroup extends CrtResource {
      */
     public static void setStaticDefaultNumThreads(int numThreads) {
         synchronized (EventLoopGroup.class) {
-            staticDefaultNumThreads = numThreads;
+            staticDefaultNumThreads = Math.max(1, numThreads);
         }
     }
 
@@ -107,7 +107,7 @@ public final class EventLoopGroup extends CrtResource {
         return elg;
     }
 
-    private static int staticDefaultNumThreads = 1;
+    private static int staticDefaultNumThreads = Math.max(1, Runtime.getRuntime().availableProcessors());
     private static EventLoopGroup staticDefaultEventLoopGroup;
 
     /*******************************************************************************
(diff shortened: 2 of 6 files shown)


src/main/java/software/amazon/awssdk/crt/io/EventLoopGroup.java
```diff
@@ -58,6 +58,58 @@ private void onCleanupComplete() {
...
+                staticDefaultEventLoopGroup = new EventLoopGroup(staticDefaultNumThreads);
+            }
+
+            elg = staticDefaultEventLoopGroup;
+        }
+
+        return elg;
+    }
+
+    private static int staticDefaultNumThreads = 1;
```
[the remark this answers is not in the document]
bretambrose: I feel like I would be more surprised to see 12-16 threads spring up out of nowhere when I write "client.builder().create()" then seeing 1 thread spring up, but that's just me.  What happens with the comparable netty event loop group?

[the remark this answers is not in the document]
zoewangg: Netty uses 2 * cores

https://github.com/netty/netty/blob/b908c82b7b60afae16f510f6cbb5093b9a69ee33/transport/src/main/java/io/netty/channel/MultithreadEventLoopGroup.java#L40

It might be an issue if customers are using multiple SDK clients because it seems the default number of threads can't be changed if it's already created.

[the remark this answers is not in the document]
bretambrose: Hmm, seems reasonable to follow their lead then.

src/main/java/software/amazon/awssdk/crt/io/EventLoopGroup.java
```diff
@@ -58,6 +58,58 @@ private void onCleanupComplete() {
...
+
+    /**
+     * Sets the number of threads for the static default event loop group, should it ever be created.  Has no
+     * effect if the static default event loop group has already been created.
+     *
+     * @param numThreads number of threads for the static default event loop group
+     */
+    public static void setStaticDefaultNumThreads(int numThreads) {
+        synchronized (EventLoopGroup.class) {
+            staticDefaultNumThreads = Math.max(1, numThreads);
```
JonathanHenson, line 76: is this the right default? Should we default to num cores? using 0 would do the trick.

bretambrose: Default is num cores; this is how to override default when you don't like it but still don't want to manage your own elg.

src/main/java/software/amazon/awssdk/crt/io/HostResolver.java
```diff
@@ -36,6 +36,57 @@ protected void releaseNativeHandle() {
...
+     * if they are passed a null value.
+     */
+
+    /**
+     * Sets the max number of cached host entries for the static default resolver, if it's ever created/used. Has no
+     * effect if the static default host resolver has already been created.
+     *
+     * @param maxEntries maximum number of host entries cached
+     */
+    public static void setStaticDefaultMaxEntries(int maxEntries) {
```
graebm, line 51: same. `setStaticDefault(HostResolver)` seems simpler

bretambrose: Nothing stops us from adding those later, but this is the "I don't want to manage crt resources" interface and presumably we would be taking refs to passed in objects which implies the user still kind of needs to think about it too.

src/main/java/software/amazon/awssdk/crt/io/ClientBootstrap.java
```diff
@@ -25,6 +25,14 @@
      * or if the system is unable to allocate space for a native client bootstrap object
      */
     public ClientBootstrap(EventLoopGroup elg, HostResolver hr) throws CrtRuntimeException {
+        if (elg == null) {
```
DavidOgunsAWS, line 28: It would be better to keep this constructor's requirements to have a non-null eventLoopGroup and hostResolver and add an overload with no parameters use the static.

PR #239, opened 2020-10-26
Event stream rpc java bindings

JonathanHenson:


src/test/java/software/amazon/awssdk/crt/test/EventStreamClientConnectionTest.java
```diff
@@ -0,0 +1,674 @@
...
+                clientConnectionClosed[0] = true;
+                this.close();
+            }
+        });
+
+        connectFuture.get();
+        assertNotNull(clientConnectionArray[0]);
+        clientConnectionArray[0].closeConnection(0);
+        // this should be damn near instant, but give it a second just to be safe.
+        Thread.sleep(1000);
```
justinboswell, line 89: Can we not drive this with a future that gets finished in `onConnectionClosed()`?

I'm a little worried about these sporadically failing in CI because the sleep wasn't long enough on multi-tenant boxes.

src/native/event_stream_rpc_client.c
```diff
@@ -0,0 +1,613 @@
...
+    jint jvmresult = (*env)->GetJavaVM(env, &callback_data->jvm);
+    if (jvmresult != 0) {
+        aws_jni_throw_runtime_exception(env, "ClientConnection.clientConnect: Unable to get JVM");
+        goto error;
+    }
+
+    callback_data->java_connection_handler = (*env)->NewGlobalRef(env, jni_client_connection_handler);
+
+    if (!callback_data->java_connection_handler) {
+        aws_jni_throw_runtime_exception(env, "ClientConnection.clientConnect: Unable to create global weak ref");
```
justinboswell, line 168: This isn't a weak ref on the line above. Should it be?

src/main/java/software/amazon/awssdk/crt/eventstream/MessageFlushCallback.java
```diff
@@ -0,0 +1,5 @@
+package software.amazon.awssdk.crt.eventstream;
+
+public interface MessageFlushCallback {
```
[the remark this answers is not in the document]
justinboswell: Oh, was this because it has to be resolvable from native? Ugh.

src/test/java/software/amazon/awssdk/crt/test/EventStreamClientConnectionTest.java
justinboswell: This applies to all the sleeping tests.

justinboswell:
We need to do a javadoc pass over the public interfaces here. Can do that in a follow-up, but we need to do it.

justinboswell:
```
/local/home/boswej/dev/aws-crt-java/src/main/java/software/amazon/awssdk/crt/eventstream/ClientConnection.java:41: warning: no @return
    public boolean isClosed() {
```

src/main/java/software/amazon/awssdk/crt/eventstream/ClientConnection.java
```diff
@@ -0,0 +1,197 @@
...
+import software.amazon.awssdk.crt.io.SocketOptions;
+
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Wrapper around an event stream rpc client initiated connection.
+ */
+public class ClientConnection extends CrtResource {
```
graebm, line 17: Should we name this `EventStreamRpcClientConnection`?

in the `software.amazon.awssdk.crt.http` package, we still prefix the name of all classes like `HttpHeaders` and `HttpClientConnection`. This seemed like a good idea because there are many kinds of Headers and ClientConnections in this world. But `Http` is only 4 letters, and `EventStreamRpc` is 14

src/main/java/software/amazon/awssdk/crt/eventstream/ClientConnection.java
```diff
@@ -0,0 +1,197 @@
...
+     */
+    public static CompletableFuture<Void> connect(final String hostName, short port, final SocketOptions socketOptions,
+                                           final ClientTlsContext tlsContext, final ClientBootstrap bootstrap,
+                                           final ClientConnectionHandler connectionHandler) {
+        long tlsContextHandle = tlsContext != null ? tlsContext.getNativeHandle() : 0;
+
+        CompletableFuture<Void> future = new CompletableFuture<>();
+        ClientConnectionHandler handlerShim = new ClientConnectionHandler() {
+            @Override
+            protected void onConnectionSetup(ClientConnection connection, int errorCode) {
```
graebm, line 145: I'm debating whether or not it would be better if we did this:

If the connection fails, don't invoke ANY of the handler callbacks, just have the connect() function's future complete exceptionally. Now we can remove the error_code argument from ClientConnectionHandler.onConnectionSetup(). Now none of the handler callbacks fire if the connection failed. Now there's only one way to check for errors instead of 2.

Thoughts?

src/main/java/software/amazon/awssdk/crt/eventstream/Header.java
```diff
@@ -0,0 +1,372 @@
...
+        return headerValue[0];
+    }
+
+    private void setValue(byte value) {
+        headerType = HeaderType.Byte;
+        headerValue = new byte[] { value};
+    }
+
+    public short getValueAsShort() {
+        checkType(HeaderType.Int16);
```
graebm, line 200: Confusing that we use the Java term "short" in the function name, but use the protocol term "int16" in the enum name.

We should pick one approach or the other. Always use the language-idiomatic term (ex: "short", "date", "bytes", etc), or always use the protocol term (ex: "int16", "timestamp", "byteBuf", etc).

I'm leaning towards calling protocol terminology everywhere. Like, the docs for a service will say "send an INT16 header named 'blah'", so it's nice if the same terminology is used in each language API.

src/main/java/software/amazon/awssdk/crt/eventstream/ClientConnection.java
graebm: Also, I don't see the "Rpc" being using anywhere in the java code, to distinguish it from pure "event-loop" protocol

src/main/java/software/amazon/awssdk/crt/eventstream/ClientConnectionContinuationHandler.java
```diff
@@ -0,0 +1,67 @@
...
+     * either call close() manually or invoke super.onContinuationClosed() before returning.
+     */
+    protected void onContinuationClosed() {
+        this.close();
+    }
+
+    void onContinuationClosedShim() {
+        onContinuationClosed();
+        closedFuture.complete(null);
+    }
```
graebm, line 51: why not call `this.close()` from the `onContinuationClosedShim()`? Then you don't need to worry about the user forgetting to call it from their override.

Or are you trying to give the user control over calling close() before or after they take their custom actions?

src/main/java/software/amazon/awssdk/crt/eventstream/Header.java
```diff
@@ -0,0 +1,372 @@
...
+
+    private void checkType(HeaderType headerType) {
+        if (this.headerType != headerType) {
+            throw new CrtRuntimeException("Invalid Event-stream header type");
+        }
+    }
+
+    private static void checkHeaderNameLen(final String headerName) {
+        if (headerName.length() > Byte.MAX_VALUE) {
+            throw new CrtRuntimeException("Header name must be less than 127 bytes.");
```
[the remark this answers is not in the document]
JonathanHenson: no, it's signed thanks to java.

src/main/java/software/amazon/awssdk/crt/eventstream/ClientConnection.java
JonathanHenson: we cannot depend purely on futures, or older android customers will not be able to use CRT.

src/main/java/software/amazon/awssdk/crt/eventstream/Header.java
JonathanHenson: I think java devs would rather deal with type names mapped into their domain language. However the int* names refer to the protocol. I dunno. I don't care for it either, but it seems less bad than the alternative to me.

src/main/java/software/amazon/awssdk/crt/eventstream/ClientConnectionContinuationHandler.java
JonathanHenson: I'd rather the user have the ability to change the behavior if they need to (like defer close() for whatever reason). Worst case we can come back and change it if I'm wrong. close() is idempotent.

justinboswell (approved):

PR #240, opened 2020-10-27
Changing the jni connection handler ref from weak ref to global ref. …

abanthiy:
…Client connection handler was not getting called with weak ref in some cases

*Issue #, if available:*
When client triggers connect to the server, upon successful connect system call, the jni connection handler does not get invoked about 80% times. I believe the reason is that the reference to the connection handler gets lost. 

*Description of changes:*
I am changing the reference for the jni handler from weak ref to global ref. Corresponding change when deleting the ref for the handler. 


JonathanHenson:
I merged these changes in the main event-stream-rpc PR. Thanks!

PR #244, opened 2020-11-09
Fixing CI issues

justinboswell:

### DIFF
--- a/src/native/aws_signing.c
+++ b/src/native/aws_signing.c
@@ -142,6 +142,7 @@ static void s_aws_signing_complete(struct aws_signing_result *result, int error_
 
     (*env)->CallBooleanMethod(
         env, callback_data->java_future, completable_future_properties.complete_method_id, java_signed_request);
+    AWS_FATAL_ASSERT(!(*env)->ExceptionCheck(env));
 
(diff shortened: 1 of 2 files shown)


bretambrose (commented):
Thanks for finding these.  Had we decided that check was insufficient and that we needed clear?  Or was there a good heuristic there (clear if in a c thread, check if in a java thread, etc...)?

bretambrose (approved):

src/native/aws_signing.c
```diff
@@ -142,6 +142,7 @@ static void s_aws_signing_complete(struct aws_signing_result *result, int error_
 
     (*env)->CallBooleanMethod(
         env, callback_data->java_future, completable_future_properties.complete_method_id, java_signed_request);
+    AWS_FATAL_ASSERT(!(*env)->ExceptionCheck(env));
```
JonathanHenson, line 145: why not:

AWS_FATAL_ASSERT(!aws_jni_check_and_clear_exception(env));

?

alexw91 (approved):

src/native/aws_signing.c
justinboswell: I will switch all of them to that when your branch merges, that only exists in your branch ATM.

PR #268, opened 2021-01-27
s3 sample

graebm:

### DIFF
--- a/samples/s3/pom.xml
+++ b/samples/s3/pom.xml
@@ -1,0 +1,76 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>com.example.s3</groupId>
+  <artifactId>s3</artifactId>
+  <version>1.0-SNAPSHOT</version>
+
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <maven.compiler.source>1.8</maven.compiler.source>
+    <maven.compiler.target>1.8</maven.compiler.target>
+  </properties>
+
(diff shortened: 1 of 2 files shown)


samples/s3/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+    </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.11</version>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+
+  <build>
```
DavidOgunsAWS, line 31: I don't think anything under this tag is non-default. This plugins are assumed by maven and will be included and associated with the tool itself. Try removing it and everything should work just fine.

samples/s3/src/main/java/com/example/s3/PutGet.java
```diff
@@ -0,0 +1,147 @@
...
+            }
+            System.out.println("Upload complete");
+
+            System.out.println("Downloading...");
+            GetObjectRequest getRequest = GetObjectRequest.builder().bucket(bucket).key(key).build();
+            GetDataConsumer getDataConsumer = new PutGet.GetDataConsumer(getFilepath);
+            CompletableFuture<GetObjectOutput> getFuture = s3Client.getObject(getRequest, getDataConsumer);
+            try {
+                getFuture.get();
+            } catch (InterruptedException | ExecutionException e) {
```
DavidOgunsAWS, line 76: extreme nit: ExecutionExceptions usually wrap the relevant exception in the causedBy and don't reveal anything itself. The full stack trace will still show what matters, but usually you can unwrap those once before rewrapping in RuntimeException and throwing that.

DavidOgunsAWS (approved):

samples/s3/pom.xml
graebm: Cool. I cut out EVERY plugin, and it still seems to work.

These all came from letting maven create an "archetype" project for me.

samples/s3/src/main/java/com/example/s3/PutGet.java
graebm: I'm going to keep it as is, for the sake of keeping the sample as simple as possible. If the stacktrace shows the issue, that's sufficient

Thanks for the info, though

rccarper (approved):

PR #270, opened 2021-02-03
MqttMessage now contains qos, retain, dup

graebm:
This allows users to know the `qos`, `retain`, and `dup` of received messages.

Functions that took `MqttMessage`, `qos`, and `retain` as separate parameters have been deprecated in favor of functions that just take an `MqttMessage`.

### DIFF
--- a/src/test/java/software/amazon/awssdk/crt/test/ProcessTest.java
+++ b/src/test/java/software/amazon/awssdk/crt/test/ProcessTest.java
@@ -15,7 +15,7 @@ import software.amazon.awssdk.crt.Process;
 
 public class ProcessTest extends CrtTestFixture  {
     public ProcessTest() {}
-    
+
     @Test
     public void testGetPid() {
         int pid = Process.getPid();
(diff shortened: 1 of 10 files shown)


src/test/java/software/amazon/awssdk/crt/test/ProcessTest.java
```diff
@@ -43,7 +43,7 @@ public void testSetSoftIOHandleLimits() {
             Process.setMaxIOHandlesSoftLimit(softLimit - 1);
         } catch (CrtRuntimeException ex) {
             // make sure it's the not-implemented exception if it was thrown.
-            assertEquals(36, ex.errorCode);
+            assertEquals("AWS_ERROR_UNIMPLEMENTED", ex.errorName);
```
sdavtaker, line 46: Can we build an enum class for these AWS_ERROR_* ?

sdavtaker (approved):

src/test/java/software/amazon/awssdk/crt/test/ProcessTest.java
bretambrose: Java CRT exception and error codes needs a major overhaul that's probably out of scope here.

bretambrose (approved):

src/test/java/software/amazon/awssdk/crt/test/ProcessTest.java
graebm: Yes, but not in this PR

We should have a definitive list of errors in code, maybe as enums, maybe as a concrete class types per C error code. This definitive list should be code-generated and not hand-typed, because it's not uncommon for new codes to appear. This is a task in our backlog...

PR #274, opened 2021-02-10
Add bindings to support using chunked encoding.

JonathanHenson:

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/http/HttpStream.java
+++ b/src/main/java/software/amazon/awssdk/crt/http/HttpStream.java
@@ -5,7 +5,11 @@
 
 package software.amazon.awssdk.crt.http;
 
+import software.amazon.awssdk.crt.CRT;
 import software.amazon.awssdk.crt.CrtResource;
+import software.amazon.awssdk.crt.CrtRuntimeException;
+
+import java.util.concurrent.CompletableFuture;
 
 /**
  * An HttpStream represents a single Http Request/Response pair within a HttpClientConnection, and wraps the native resources
@@ -14,6 +18,10 @@ import software.amazon.awssdk.crt.CrtResource;
  * Can be used to update the Window size, or to abort the stream early in the middle of sending/receiving Http Bodies.
  */
 public class HttpStream extends CrtResource {
+
+    public interface HttpStreamWriteChunkCompletionCallback {
+        void onChunkCompleted(int errorCode);
+    }
 
     /* Native code will call this constructor during HttpClientConnection.makeRequest() */
     protected HttpStream(long ptr) {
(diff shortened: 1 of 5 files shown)


src/test/java/software/amazon/awssdk/crt/test/HttpRequestResponseTest.java
```diff
@@ -171,44 +178,65 @@ private boolean shouldRetry(TestHttpResponse response) {
         return false;
     }
 
-    public TestHttpResponse testRequest(String method, String endpoint, String path, String requestBody, int expectedStatus) throws Exception {
+    public TestHttpResponse testRequest(String method, String endpoint, String path, String requestBody, boolean useChunkedEncoding, int expectedStatus) throws Exception {
```
justinboswell, line 181: trivial: just add an overload signature for this function that is the old one, which calls this overload with `useChunkedEncoding` set to `false`, much smaller change.

bretambrose (approved):

src/main/java/software/amazon/awssdk/crt/http/HttpStream.java
```diff
@@ -55,6 +63,57 @@ public void incrementWindow(int windowSize) {
...
+
+    /**
+     * Use only for Http 1.1 Chunked Encoding. At some later point we may adapt this interface for H2, but not yet. If
+     * using chunked encoding, remember, you'll need to send an empty chunk to signal the end of your stream. You must call
+     * activate() before using this function.
+     *
+     * @param chunkData chunk of data to send.
+     * @return completable future which will complete upon the data being flushed to the wire or an error occurring.
+     */
+    public CompletableFuture<Void> writeChunk(final byte[] chunkData) {
```
justinboswell, line 99: Should we make a `finishChunks()` or something that automates writing an empty chunk to finish the stream?

justinboswell (approved):
LGTM, feedback is optional

src/main/java/software/amazon/awssdk/crt/http/HttpStream.java
```diff
@@ -55,6 +63,57 @@ public void incrementWindow(int windowSize) {
         }
     }
 
+    /**
+     * Use only for Http 1.1 Chunked Encoding. At some later point we may adapt this interface for H2, but not yet. If
```
rschmitt, line 67: How would h2 be any different?

src/main/java/software/amazon/awssdk/crt/http/HttpStream.java
```diff
@@ -55,6 +63,57 @@ public void incrementWindow(int windowSize) {
...
+     * using chunked encoding, remember, you'll need to send an empty chunk to signal the end of your stream. You must call
```
rschmitt, line 68: I think this distinction will push unnecessary complexity into customer code. If you use a `Content-Length`, can you publish an empty chunk at the end of the stream?

src/main/java/software/amazon/awssdk/crt/http/HttpStream.java
JonathanHenson: If you were sending something over h2 for a use case you'd normally use chunked encoding for in http 1.1, the API should be identical. But, I don't think it's plumbed together in aws-c-http at the moment. When it is, I'll connect the pieces here and change the comment.

src/test/java/software/amazon/awssdk/crt/test/HttpRequestResponseTest.java
JonathanHenson: Seems to just make it less clear and explicit to me, but doesn't shave off the majority of the code I needed to change.

justinboswell (approved):

PR #281, opened 2021-02-25
custom credential provider

TingDaoK:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/native/credentials_provider.c
+++ b/src/native/credentials_provider.c
@@ -282,12 +282,17 @@ static int s_credentials_provider_delegate_get_credentials(
     struct aws_allocator *allocator = aws_jni_get_allocator();
 
     JNIEnv *env = aws_jni_get_thread_env(callback_data->jvm);
+    int return_value = AWS_OP_ERR;
 
     // Fetch credentials from java
     jobject java_credentials = (*env)->CallObjectMethod(
         env,
         callback_data->jni_delegate_credential_handler,
         credentials_handler_properties.on_handler_get_credentials_method_id);
+    if (aws_jni_check_and_clear_exception(env)) {
+        goto call_failed;
+    }
+
     jbyteArray java_access_key_id =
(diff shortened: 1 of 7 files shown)


src/native/credentials_provider.c
```diff
@@ -52,6 +54,9 @@ static void s_on_shutdown_complete(void *user_data) {
     }
 
     (*env)->DeleteWeakGlobalRef(env, callback_data->java_crt_credentials_provider);
+    if (callback_data->jni_delegate_credential_handler != NULL) {
```
bretambrose, line 57: The cleanup of the callback data needs to be refactored into a separate function and then invoked both here and in the failure case of the delegate constructor since global and weak refs could have been built at that point.  In fact it looks like none of the constructors delete the existing weak global ref (which might be safe) but it definitely isn't now that there's potentially a strong global ref.

src/native/credentials_provider.c
```diff
@@ -268,6 +273,96 @@ JNIEXPORT jlong JNICALL
...
+                .shutdown_user_data = callback_data,
+            },
+    };
+
+    jint jvmresult = (*env)->GetJavaVM(env, &callback_data->jvm);
+    AWS_FATAL_ASSERT(jvmresult == 0);
+
+    struct aws_credentials_provider *provider = aws_credentials_provider_new_delegate(allocator, &options);
+    if (provider == NULL) {
+        aws_mem_release(allocator, callback_data);
```
bretambrose, line 357: See above comment, we need to fully clean up callback_data since it holds multiple refs potentially at this point.

bretambrose (changes requested):

src/native/credentials_provider.c
TingDaoK: Thanks! Good catch! Updated🙆‍♂️

bretambrose (approved):

PR #287, opened 2021-03-04
Gentle shutdown by default

bretambrose:
* By default, do a gentle shutdown that does not require all native threads to be joined before jni exit
* https://github.com/aws/aws-iot-device-sdk-java-v2/issues/139

### DIFF
--- a/src/native/crt.c
+++ b/src/native/crt.c
@@ -213,7 +213,7 @@ JNIEnv *aws_jni_get_thread_env(JavaVM *jvm) {
     return env;
 }
 
-static void s_jni_atexit(void) {
+static void s_jni_atexit_strict(void) {
     aws_s3_library_clean_up();
     aws_event_stream_library_clean_up();
     aws_auth_library_clean_up();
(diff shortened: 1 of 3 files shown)


src/native/crt.c
```diff
@@ -233,6 +233,38 @@ static void s_jni_atexit(void) {
     }
 }
 
+#define DEFAULT_MANAGED_SHUTDOWN_WAIT_IN_SECONDS 1
+
+static void s_jni_atexit_gentle(void) {
+
+    /* If not doing strict shutdown, wait only a short time before shutting down */
+    aws_thread_set_managed_join_timeout_ns(
+        aws_timestamp_convert(DEFAULT_MANAGED_SHUTDOWN_WAIT_IN_SECONDS, AWS_TIMESTAMP_SECS, AWS_TIMESTAMP_NANOS, NULL));
```
graebm, line 242: we're going to add a 1 second pause on the shutdown of any application that forgets to close a handle?

what if we just **didn't even try to clean up anything**?

bretambrose: We could do that too, but it makes me sad.

graebm: I don’t think a 1sec pause is cool. like imagine someone wrote a tiny helper CLI type application that gets run in a loop by a larger script. This would make it painfully slow.

Let's either wait some infinitesimally short time, or not even try to clean up.

graebm: chatted offline. there are advantages to at least trying to clean up.

we can always tweak this behavior going forward...

graebm (approved):

PR #301, opened 2021-03-31
S3 Message Customization

rccarper:
*Description of changes:*
Adding support for custom headers and custom query parameters to generated requests.

Note: Diffs may appear larger than they are due to formatting changes.

Only GetObjectRequest and PutObjectRequest are in this PR.  Other generated files are in this PR: https://github.com/awslabs/aws-crt-java/pull/302


s3-native-client/build.gradle.kts
```diff
@@ -27,6 +27,6 @@ dependencies {
     implementation(project(":smithy-crt"))
     implementation(rootProject)
 
-    testImplementation(testFixtures(rootProject))
+    //testImplementation(testFixtures(rootProject))
```
bretambrose, line 30: Remove?

s3-native-client/src/main/java/com/amazonaws/s3/S3NativeClient.java
```diff
@@ -17,41 +19,82 @@
...
+            return;
+        }
+
+        String trimmedCustomQueryParameters = customQueryParameters.trim();
+
+        if (trimmedCustomQueryParameters.equals("")) {
+            return;
+        }
+
+        queryParameters.append("?" + trimmedCustomQueryParameters);
```
bretambrose, line 78: If it's assumed(required) that queryParameters comes in empty then maybe it should be created and returned instead.

s3-native-client/src/main/java/com/amazonaws/s3/S3NativeClient.java
```diff
@@ -79,30 +122,32 @@ public void onFinished(int errorCode) {
...
         };
 
         List<HttpHeader> headers = new LinkedList<>();
+
         // TODO: additional logic needed for *special* partitions
-        headers.add(new HttpHeader("Host", request.bucket() + ".s3."+ signingRegion + ".amazonaws.com"));
+        headers.add(new HttpHeader("Host", request.bucket() + ".s3." + signingRegion + ".amazonaws.com"));
         populateGetObjectRequestHeaders(header -> headers.add(header), request);
-        final StringBuilder keyString = new StringBuilder("/" + request.key());
+        final StringBuilder queryParameters = new StringBuilder("/" + request.key());
```
bretambrose, line 137: This feels weird/confusing, query parameters are separate from path and this is kind of mixing them state-wise.

s3-native-client/src/main/java/com/amazonaws/s3/S3NativeClient.java
```diff
@@ -302,35 +353,33 @@ protected void populatePutObjectRequestHeaders(final Consumer<HttpHeader> header
...
         if (request.sSEKMSKeyId() != null) {
-            headerConsumer.accept(new HttpHeader("x-amz-server-side-encryption-aws-kms-key-id",
-                    request.sSEKMSKeyId()));
+            headerConsumer.accept(new HttpHeader("x-amz-server-side-encryption-aws-kms-key-id", request.sSEKMSKeyId()));
         }
         if (request.sSEKMSEncryptionContext() != null) {
-            headerConsumer.accept(new HttpHeader("x-amz-server-side-encryption-context",
-                    request.sSEKMSEncryptionContext()));
+            headerConsumer
+                    .accept(new HttpHeader("x-amz-server-side-encryption-context", request.sSEKMSEncryptionContext()));
```
bretambrose, line 382: These strings should go in constants somewhere.

s3-native-client/src/main/java/com/amazonaws/s3/S3NativeClient.java
rccarper: I've tried to better manage the state "getEncodedPath" function.

s3-native-client/src/main/java/com/amazonaws/s3/S3NativeClient.java
rccarper: Agreed, there's some house keeping that needs to be done here.  I'd like that to be a separate PR since this code is only changed due to formatting changes.

s3-native-client/src/generated/java/com/amazonaws/s3/model/GetObjectRequest.java
```diff
@@ -564,6 +591,16 @@ public final Builder expectedBucketOwner(String expectedBucketOwner) {
             return this;
         }
 
+        public final Builder customHeaders(HttpHeader[] customHeaders) {
+            this.customHeaders = customHeaders;
+            return this;
+        }
+
+        public final Builder customQueryParameters(String customQueryParameters) {
```
zoewangg, line 599: Would it make sense to allow users to provide a single query parameter instead? Otherwise users would need to build the query parameters by themselves. 

This is Java SDK v2's API
https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/RequestOverrideConfiguration.java#L254-L256

s3-native-client/src/main/java/com/amazonaws/s3/S3NativeClient.java
```diff
@@ -17,41 +19,83 @@
...
+    }
+
+    private String getEncodedPath(String key, String customQueryParameters) {
+        String encodedPath = "/" + key;
+
+        if (customQueryParameters == null) {
+            return encodedPath;
+        }
+
+        String trimmedCustomQueryParameters = customQueryParameters.trim();
```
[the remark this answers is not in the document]
rccarper: Because they're already packaged as a string right now, they should already be passed in with the necessary base 64 encoding.  (Maybe there should be some better documentation there)

s3-native-client/src/generated/java/com/amazonaws/s3/model/GetObjectRequest.java
rccarper: Hmmm I do like that idea.  But, I think we'd have to encode the query parameters, and I'm not sure we have a Base64 encoder we can use inside of the Java CRT right now.  I'll double check!

s3-native-client/src/main/java/com/amazonaws/s3/S3NativeClient.java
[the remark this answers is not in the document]
bretambrose: I think uri encoding is the question here rather than base64.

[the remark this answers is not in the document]
bretambrose: Also I feel like a solution that requires the caller to have done the encoding may be better, in order to avoid any of the weirdness (similar to signing) where parameters may be path-embedded and whether or not they should be encoded varies.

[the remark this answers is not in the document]
rccarper: Ah, my mistake!  I like Bret's thought on this.

s3-native-client/src/generated/java/com/amazonaws/s3/model/GetObjectRequest.java
rccarper: I guess we can require the customer to have properly prepared the parameters and still build the actual string for them.  I'll look at doing that.

rccarper: Changed it to a Map<String,String>.

s3-native-client/src/main/java/com/amazonaws/s3/S3NativeClient.java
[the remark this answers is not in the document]
zoewangg: Looking at the code again, I now lean towards to requiring the caller to pass the encoded and combined param string. That way we don't need to worry about encoding and supporting query params with multiple values in this client

zoewangg (approved):

bretambrose (approved):

PR #319, opened 2021-05-21
add profile credentials provider support

ianbotsf:
*Issue #, if available:*

*Description of changes:* Adds CRT bindings and Java wrapper for a profile credentials provider.

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/auth/credentials/ProfileCredentialsProvider.java
+++ b/src/main/java/software/amazon/awssdk/crt/auth/credentials/ProfileCredentialsProvider.java
@@ -42,9 +42,13 @@ public class ProfileCredentialsProvider extends CredentialsProvider {
     private ProfileCredentialsProvider(BuilderImpl builder) {
         super();
 
+        ClientBootstrap bootstrap = builder.clientBootstrap == null
+                ? new ClientBootstrap(null, null)
+                : builder.clientBootstrap;
+
         long nativeHandle = profileCredentialsProviderNew(
                 this,
-                toNativeHandle(builder.clientBootstrap),
+                toNativeHandle(bootstrap),
                 toNativeHandle(builder.tlsContext),
                 toByteArray(builder.profileName),
                 toByteArray(builder.configFileNameOverride),
@@ -52,7 +56,7 @@ public class ProfileCredentialsProvider extends CredentialsProvider {
         );
 
         acquireNativeHandle(nativeHandle);
-        if (builder.clientBootstrap != null) { addReferenceTo(builder.clientBootstrap); }
+        addReferenceTo(bootstrap);
         if (builder.tlsContext != null) { addReferenceTo(builder.tlsContext); }
     }
 
@@ -107,7 +111,7 @@ public class ProfileCredentialsProvider extends CredentialsProvider {
     }
 
     static final class BuilderImpl implements Builder {
-        private ClientBootstrap clientBootstrap = new ClientBootstrap(null, null);
+        private ClientBootstrap clientBootstrap;
         private TlsContext tlsContext;
         private String profileName;
         private String configFileNameOverride;
--- a/src/main/java/software/amazon/awssdk/crt/auth/credentials/ProfileCredentialsProvider.java
+++ b/src/main/java/software/amazon/awssdk/crt/auth/credentials/ProfileCredentialsProvider.java
@@ -16,85 +16,30 @@ import java.nio.charset.StandardCharsets;
  */
 public class ProfileCredentialsProvider extends CredentialsProvider {
     /**
-     * A builder class for the profile credentials provider.
+     * Create a profile credentials provider using the default file locations and profile name.
+     * @return A profile credentials provider.
      */
-    static public class ProfileCredentialsProviderBuilder {
-        private ClientBootstrap clientBootstrap;
-        private TlsContext tlsContext;
-        private String profileName;
-        private String configFileNameOverride;
-        private String credentialsFileNameOverride;
-
-        public ProfileCredentialsProviderBuilder() {}
-
-        /**
-         * Sets the client bootstrap (host resolver and event loop group) to use when making the connections
-         * required by this provider.
-         * @param clientBootstrap client bootstrap to use
-         * @return The current builder
-         */
-        public ProfileCredentialsProviderBuilder withClientBootstrap(ClientBootstrap clientBootstrap) {
-            this.clientBootstrap = clientBootstrap;
-
-            return this;
-        }
-
-        ClientBootstrap getClientBootstrap() { return clientBootstrap; }
-
-        /**
-         * Sets the tls context to use for any secure network connections made while sourcing credentials.
-         * @param tlsContext the tls context to use when establishing network connections
-         * @return The current builder
-         */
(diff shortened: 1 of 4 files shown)


src/test/java/software/amazon/awssdk/crt/test/CredentialsProviderTest.java
```diff
@@ -151,4 +157,34 @@ public Credentials getCredentials() {
...
+                assertNotNull(provider);
+                assertTrue(provider.getNativeHandle() != 0);
+            }
+        } finally {
+            Files.deleteIfExists(credsPath);
+        }
+    }
+
+    @Test(expected = CrtRuntimeException.class)
+    public void testCreateDestroyProfile_InvalidProfile() {
```
bretambrose, line 181: Why is this test going to throw an exception?

src/main/java/software/amazon/awssdk/crt/auth/credentials/ProfileCredentialsProvider.java
```diff
@@ -0,0 +1,133 @@
...
+
+        public ProfileCredentialsProvider build() { return new ProfileCredentialsProvider(this); }
+    }
+
+    private ProfileCredentialsProvider(ProfileCredentialsProviderBuilder builder) {
+        super();
+
+        long nativeHandle = profileCredentialsProviderNew(
+                this,
+                toNativeHandle(builder.clientBootstrap),
```
bretambrose, line 102: client bootstrap and tls context need to be required I think.  You can't do role assumption without them.

ianbotsf: It appears the underlying code [utilizes a default TLS context](https://github.com/awslabs/aws-c-auth/blob/a5dbb0cb7fccb6405b16245f4be7cbd2225496bd/source/credentials_provider_profile.c#L250-L258) if none is provided (but doesn't do the same for client bootstrap).

It also appears that role assumption only occurs [when the `role_arn` key is found in the profile](https://github.com/awslabs/aws-c-auth/blob/a5dbb0cb7fccb6405b16245f4be7cbd2225496bd/source/credentials_provider_profile.c#L400-L407).

Do we want to require a bootstrap even if the profile isn't guaranteed to contain `role_arn`? If so, is there a sensible default for the Java builder to use?

src/test/java/software/amazon/awssdk/crt/test/CredentialsProviderTest.java
ianbotsf: Because the config/credentials files will not contain a default profile, triggering [this error condition](https://github.com/awslabs/aws-c-auth/blob/a5dbb0cb7fccb6405b16245f4be7cbd2225496bd/source/credentials_provider_profile.c#L392-L399). I will add comments more clearly indicating why the exception is expected.

bretambrose: But you're not overriding the default which means its going to the local users files which may very well have those profiles (like mine when I ran it, for example).

ianbotsf: Oh good point. I'll manually create empty files without profiles to verify this error condition.

src/main/java/software/amazon/awssdk/crt/auth/credentials/ProfileCredentialsProvider.java
bretambrose: Technically creating a new ClientBootstrap(null, null) will use a shared, lazily created host resolver and event loop group, so that may be a possibily.  Failing creation if it's not provided is also a possibility.

ianbotsf: Using `new ClientBootstrap(null, null)` makes sense. Will update!

src/test/java/software/amazon/awssdk/crt/test/CredentialsProviderTest.java
```diff
@@ -151,4 +157,41 @@ public Credentials getCredentials() {
...
+
+        try (ProfileCredentialsProvider provider = builder.build()) {
+            assertNotNull(provider);
+            assertTrue(provider.getNativeHandle() != 0);
+        } finally {
+            Files.deleteIfExists(credsPath);
+        }
+    }
+
+    @Test(expected = CrtRuntimeException.class) // Because the profile files don't contain a [default] section
```
bretambrose, line 180: A little confused here.  The lack of any information in the test file shouldn't afffect provider creation only credentials sourcing.

src/test/java/software/amazon/awssdk/crt/test/CredentialsProviderTest.java
```diff
@@ -151,4 +157,41 @@ public Credentials getCredentials() {
             fail(ex.getMessage());
         }
     }
+
+    @Test
```
bretambrose, line 161: Either in separate or combined tests, lets actually source credentials from your test files (or fail to source them).

src/main/java/software/amazon/awssdk/crt/auth/credentials/ProfileCredentialsProvider.java
```diff
@@ -0,0 +1,165 @@
...
+
+        /**
+         * Create a profile credentials provider using the configuration applied to this builder.
+         * @return A new profile credentials provider.
+         */
+        ProfileCredentialsProvider build();
+    }
+
+    static final class BuilderImpl implements Builder {
+        private ClientBootstrap clientBootstrap = new ClientBootstrap(null, null);
```
bretambrose, line 110: Do this at provider construction time not as a default initializer.  As it stands, instantiating this class will always cause an event loop group and host resolver to be created lazily, which isn't desirable.

src/test/java/software/amazon/awssdk/crt/test/CredentialsProviderTest.java
ianbotsf: The underlying C code reads the profile immediately upon creation of the provider. I believe it's looking for a role ARN to determine whether the provider will be STS-based or not. When it finds no default profile, it [throws an error](https://github.com/awslabs/aws-c-auth/blob/a5dbb0cb7fccb6405b16245f4be7cbd2225496bd/source/credentials_provider_profile.c#L392-L399) which gets wrapped as this CrtRuntimeException.

src/main/java/software/amazon/awssdk/crt/auth/credentials/ProfileCredentialsProvider.java
ianbotsf: Will do!

src/test/java/software/amazon/awssdk/crt/test/CredentialsProviderTest.java
ianbotsf: Good idea, I'll modify the `ValidProfile` test to check for real credentials.

src/test/java/software/amazon/awssdk/crt/test/CredentialsProviderTest.java
bretambrose: Hmm, so it does

bretambrose:
Merged in https://github.com/awslabs/aws-crt-java/pull/329

PR #323, opened 2021-06-02
Sigv4a

bretambrose:
* Sigv4a implementation


src/main/java/software/amazon/awssdk/crt/CrtResource.java
```diff
@@ -69,9 +72,11 @@ public String toString() {
         }
 
         public CrtResource getWrapper() { return wrapper; }
+
+        public void setNativeHandle(long handle) { nativeHandle = handle; }
     }
 
-    private static final ConcurrentHashMap<Long, ResourceInstance> NATIVE_RESOURCES = new ConcurrentHashMap<>();
+    private static final HashMap<Long, ResourceInstance> CRT_RESOURCES = new HashMap<>();
```
justinboswell, line 79: Why isn't this concurrent anymore? Is synchronized an improvement?

src/native/http_request_utils.c
```diff
@@ -29,6 +30,11 @@ static int s_aws_input_stream_seek(
     enum aws_stream_seek_basis basis) {
     struct aws_http_request_body_stream_impl *impl = stream->impl;
 
+    if (!impl->is_valid) {
+        return aws_raise_error(AWS_ERROR_HTTP_INVALID_BODY_STREAM);
+        ;
```
justinboswell, line 35: this seems like it could go :)

profile.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
```
justinboswell, line 1: Is this intentional? This script is nowhere near usable by anyone who doesn't like pain :)

src/main/java/software/amazon/awssdk/crt/CrtResource.java
bretambrose: Unsure, not sure about any of the changes to CrtResource to be honest, need to go dig through the history and see how/why they ended up here.

profile.sh
bretambrose: No idea what this is!

src/main/java/software/amazon/awssdk/crt/CrtResource.java
bretambrose: Seems like it was less about safety and more about correctness.  We iterate that hash map to dump debug data and it could be getting side-affected at the same time and we really want a consistent snapshot.

profile.sh
bretambrose: Somehow got pulled in with s3-related stuff.  Can delete.

justinboswell (approved):

src/main/java/software/amazon/awssdk/crt/CrtResource.java
alexw91: Synchronizing a standard `HashMap` will have worse performance than using a `ConcurrentHashMap`. Synchronization will lock the entire HashMap any time any thread is adding or removing any time from the Map, and will cause contention issues if multiple threads are making many reads/writes simultaneously. Whereas ConcurrentHashMap only locks a very small inner subset of the Map during reads/writes, and makes it much less likely that two threads will create contention issues if both are making modifications simultaneously.

 - https://web.archive.org/web/20210122220134/https://www.ibm.com/developerworks/java/library/j-jtp08223/index.html
 - https://stackoverflow.com/questions/1291836/concurrenthashmap-vs-synchronized-hashmap
 - https://www.geeksforgeeks.org/difference-between-concurrenthashmap-and-synchronizedhashmap/

bretambrose: The map is used in ways that require a lock beyond just single hash table operations.  Logging a *consistent* snapshot of unreleased resources is pretty important in debugging.  Also, all accesses are predicated on debugNativeObjects which is not on by default.

PR #326, opened 2021-06-10
S3 Client Retry Options

rccarper:
*Description of changes:*
* Adding bindings for retry-strategy options.  Bindings for the retry strategy itself so that it can be used outside of the S3 Client is intended for a future pass.
* Adding retry-strategy options to S3Client/S3NativeClient.


src/main/java/software/amazon/awssdk/crt/io/ExponentialBackoffRetryOptions.java
```diff
@@ -0,0 +1,118 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+package software.amazon.awssdk.crt.io;
+
+import software.amazon.awssdk.crt.CrtResource;
+
+public class ExponentialBackoffRetryOptions extends CrtResource {
```
bretambrose, line 10: I feel like options is misleading.  Aren't these strategy bindings?

bretambrose, line 10: Also, I would strongly consider not even making these a crt resource and instead create down in the s3 client C binding constructor.  These seem like settings only, I don't see why you'd ever want to share them between more than one object, and deriving from CrtResource is a serious "commitment" to ref-counting properly and the more we can avoid doing so the better off we'll be issue-wise.

src/test/java/software/amazon/awssdk/crt/test/S3ClientTest.java
```diff
@@ -97,6 +92,29 @@ public void testS3ClientCreateDestroy() {
         }
     }
 
+    @Test
```
bretambrose, line 95: Can we add some tests that verify that options configurations actually behave as expected?

src/main/java/software/amazon/awssdk/crt/io/ExponentialBackoffRetryOptions.java
rccarper: These are bindings for the options structures right now.  The actual strategy bindings are intended to be added later.

https://github.com/awslabs/aws-c-io/blob/main/include/aws/io/retry_strategy.h#L101

rccarper: They are no longer CRT resources!

src/test/java/software/amazon/awssdk/crt/test/S3ClientTest.java
rccarper: Added some verification that the options are being passed correctly into the retry strategy.  I think validating further behavior of the retry strategy is probably out of scope in the CRT, but open to chat about it.

bretambrose (approved):

PR #332, opened 2021-06-16
Keep alive timeout MS -> Secs

TingDaoK:
Or, we should just set it to Sec instead Ms. Since it makes not sense to use Ms as it will be cast to uint_16


bretambrose:
Why are we changing to seconds?  Are the other CRTs using seconds there?

TingDaoK:
> Why are we changing to seconds? Are the other CRTs using seconds there?

Two things I think:
1. The keep alive time out in C is uint_16 in secs. So, the ms here will be cast to secs as int. So, it really doesn't help to use ms.
2. All the other languages use Secs instead
  1. https://github.com/awslabs/aws-crt-nodejs/blob/main/lib/native/mqtt.ts#L75-L81
  2. https://github.com/awslabs/aws-crt-python/blob/main/awscrt/mqtt.py#L204-L207
  3. https://github.com/awslabs/aws-crt-cpp/blob/main/include/aws/crt/mqtt/MqttClient.h#L183

bretambrose (approved):

PR #337, opened 2021-06-22
S3 md5

TingDaoK:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/s3/S3ClientOptions.java
+++ b/src/main/java/software/amazon/awssdk/crt/s3/S3ClientOptions.java
@@ -14,9 +14,19 @@ public class S3ClientOptions {
     private long partSize;
     private double throughputTargetGbps;
     private int maxConnections;
+    /**
+     * For multi-part upload, content-md5 will be calculated if the
+     * computeContentMd5 is set to true.
+     *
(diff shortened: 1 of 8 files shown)


src/main/java/software/amazon/awssdk/crt/s3/S3ClientOptions.java
```diff
@@ -20,10 +20,20 @@
     private long partSize;
     private double throughputTargetGbps;
     private int maxConnections;
+    /**
+     * For multi-part upload, content-md5 will be calculated if the
+     * computeContentMd5 is set to true.
+     *
+     * For single-part upload, keep the content-md5 in the initial request
```
rccarper, line 27: I think this comment is maybe a little confusing?

For single-part upload, we'll leave the md5 header as-is if it was specified, but should also be calculating it if computeContentMd5 is true and it's not there.

rccarper (approved):
Fix 'n ship.

src/main/java/software/amazon/awssdk/crt/s3/S3ClientOptions.java
TingDaoK: Oh, I copy/pasted it from C. Hmmm, so we should update the comments there as well.

PR #339, opened 2021-06-22
Null handle for delegate Credential provider

TingDaoK:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/native/credentials_provider.c
+++ b/src/native/credentials_provider.c
@@ -377,7 +377,7 @@ static int s_credentials_provider_delegate_get_credentials(
     jbyteArray java_session_token =
         (*env)->GetObjectField(env, java_credentials, credentials_properties.session_token_field_id);
 
-    if (java_access_key_id = NULL || java_secret_access_key == NULL) {
+    if (java_access_key_id == NULL || java_secret_access_key == NULL) {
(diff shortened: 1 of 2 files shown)


src/native/credentials_provider.c
```diff
@@ -379,7 +379,11 @@ static int s_credentials_provider_delegate_get_credentials(
 
     struct aws_byte_cursor access_key_id = aws_jni_byte_cursor_from_jbyteArray_acquire(env, java_access_key_id);
     struct aws_byte_cursor secret_access_key = aws_jni_byte_cursor_from_jbyteArray_acquire(env, java_secret_access_key);
-    struct aws_byte_cursor session_token = aws_jni_byte_cursor_from_jbyteArray_acquire(env, java_session_token);
+    struct aws_byte_cursor session_token;
```
rccarper, line 382: Is it worth doing similar checks for the access_key_id and the secret_access_key?  (Maybe not the same exact behavior, but something that prevents a crash if there's a null.)

TingDaoK: Good call! Should throw an exception if one of them is null. Updated

rccarper (approved):

PR #346, opened 2021-06-30
Put logs in CI

TingDaoK:
*Issue #, if available:*

*Description of changes:*
- Always get error level log enabled, and print it out after CI
- Lower the number of requests we made for concurrent request test, since it's throttled by server, and failed occasionally. 

### DIFF
--- a/builder.json
+++ b/builder.json
@@ -7,8 +7,7 @@
         "mvn -P continuous-integration -B compile"
     ],
     "test_steps": [
-        "mvn -B test -DredirectTestOutputToFile=true -DforkCount=0 -DrerunFailingTestsCount=5",
-        "cat log.txt"
+       "aws-crt-java-test"
     ],
     "imports": [
         "JDK8"
@@ -39,8 +38,7 @@
                 "mvn -P continuous-integration -B compile"
             ],
             "!test_steps": [
-                "mvn -B test -DforkCount=0 -DrerunFailingTestsCount=5",
-                "cat log.txt"
+                "aws-crt-java-test"
             ],
             "architectures": {
                 "armv6": {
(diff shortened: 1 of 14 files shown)


s3-native-client/src/test/java/com/amazonaws/s3/S3NativeClientTest.java
```diff
@@ -55,6 +56,7 @@
     @BeforeClass
     public static void haveAwsCredentials() {
         Assume.assumeTrue(areAwsCredentialsAvailable());
+        Log.initLoggingToFile(Log.LogLevel.Error, "log.txt");
```
bretambrose, line 59: Couldn't we do this in a before-traited function in AwsClientTestFixture?

builder.json
```diff
@@ -7,7 +7,7 @@
         "mvn -P continuous-integration -B compile"
     ],
     "test_steps": [
-        "mvn -B test -DredirectTestOutputToFile=true -DforkCount=0 -DrerunFailingTestsCount=5"
+       "aws-crt-java-test"
```
bretambrose, line 10: What is this?

TingDaoK: that's using the python script to run test. https://github.com/awslabs/aws-crt-java/pull/346/files#diff-ce63a05ed0ed2cf8fc220bbf3cf6fc06e4034fae5fe8be8404060bfaf1bf0d98
So, in order to print the log out, we need to continue the command even if the first cmd failed. Get around it by two different `os.system()`.

bretambrose: Is there a builder version bump needed?  The script is using underscores, this is using dashes.

TingDaoK: Nope, the builder will find the script based on the name https://github.com/awslabs/aws-crt-java/pull/346/files#diff-ce63a05ed0ed2cf8fc220bbf3cf6fc06e4034fae5fe8be8404060bfaf1bf0d98R18

bretambrose (approved):

PR #355, opened 2021-07-13
Connection acquire native reference from weak to strong

bretambrose:

### DIFF
--- a/src/native/http_connection_manager.c
+++ b/src/native/http_connection_manager.c
@@ -362,6 +362,9 @@ static void s_on_http_conn_acquisition_callback(
         jni_error_code);
 
     AWS_FATAL_ASSERT(!aws_jni_check_and_clear_exception(env));
+
+    (*env)->DeleteGlobalRef(env, binding->java_acquire_connection_future);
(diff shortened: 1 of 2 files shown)


src/native/http_connection_manager.c
```diff
@@ -353,21 +353,15 @@ static void s_on_http_conn_acquisition_callback(
...
+    (*env)->CallStaticVoidMethod(
+        env,
+        http_client_connection_properties.http_client_connection_class,
+        http_client_connection_properties.on_connection_acquired_method_id,
+        binding->java_acquire_connection_future,
+        (jlong)binding,
+        jni_error_code);
 
-        AWS_FATAL_ASSERT(!aws_jni_check_and_clear_exception(env));
-        (*env)->DeleteLocalRef(env, acquire_connection_future);
```
graebm, line 367: trivial: worth deleting `binding->java_acquire_connection_future` after firing the callback? we don't need it anymore, right?

graebm (approved):

src/native/http_connection_manager.c
bretambrose: Good point.

PR #356, opened 2021-07-14
Crc

ilevyor:
*Issue #, if available:*

*Description of changes:*


src/main/java/software/amazon/awssdk/crt/checksums/Crc.java
```diff
@@ -0,0 +1,23 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+package software.amazon.awssdk.crt.checksums;
+
+import software.amazon.awssdk.crt.CrtResource;
+
+public abstract class Crc extends CrtResource {
```
DavidOgunsAWS, line 9: nitpicking maybe: should this class be called `Checksum` or `Checksums` then the methods being called crc32/crc32c make sense.

The "checksums" label redundancy between the namespace and class name is more common than having a redundancy in a class name and method name.

src/main/java/software/amazon/awssdk/crt/checksums/Crc.java
```diff
@@ -0,0 +1,23 @@
...
+
+    /* java built in implementation of crc32 and crc32c returns a long, this can cause some confusion if not handled
+       properly as crc32 and crc32c only result in 32 bits. be careful when mixing the built in implementation with this
+       one */
+    public static int crc32(byte[] input) {
```
TingDaoK, line 14: why not expose `previous`?

ilevyor: native function is public, I'm overloading here to make it an optional argument.

DavidOgunsAWS (approved):

src/main/java/software/amazon/awssdk/crt/checksums/Crc.java
bretambrose, line 9: Don't make this a CrtResource.  CrtResource is mostly for instantiable objects that have an associated native object where the lifetimes need to be delicately managed.

joviegas:
Can we make it to implement Checksum as in 

https://docs.oracle.com/javase/7/docs/api/java/util/zip/Checksum.html

something like https://docs.oracle.com/javase/7/docs/api/java/util/zip/CRC32.html

Its very difficult to use the current implementation as its an abstract class , Why is it implemented as an abstract class ?

Can we please have specific classes like 
CRC32 implements  java.nio.ByteBuffer.Checksum

and 

CRC32C implements  java.nio.ByteBuffer.Checksum

PR #357, opened 2021-07-21
Crc range support

ilevyor:
*Issue #, if available:*

*Description of changes:*


src/main/java/software/amazon/awssdk/crt/checksums/Checksums.java
```diff
@@ -0,0 +1,30 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+package software.amazon.awssdk.crt.checksums;
+
+import software.amazon.awssdk.crt.CrtResource;
+
+public abstract class Checksums {
+
+    /* java built in implementation of crc32 and crc32c returns a long, this can cause some confusion if not handled
```
bretambrose, line 11: Can we match the built in on the Java side instead of just a warning comment?  In particular, longs for values (and "previous") down/back into C as well, and only convert and fatal assert against 32 bit bounds.  Otherwise it seems like this is just going to constantly blow up in people's faces.

ilevyor: My one concern is how this interfaces with other sdks. Using Java's built in function fails on tests with string literals that pass with every other language. So if Java's mask looks different from every other sdk we should enforce consistency no?

bretambrose: It sounds like we haven't gotten to the bottom of this then.  Are we really saying the Java standard library implements crc wrong?

ilevyor: This is what you have to do to get the results to match (where crt is what you get when returning an int)
`    @Test
    public void testCrc32cValues() {
        S3Client.logNativeResources();
        byte[] values = new byte[31];
        for (byte i = 0; i < 31; i++) {
            values[i] = i;
        }
        long res = (long)Checksums.crc32(values) & 0xffffffffL  ;
        CRC32 crc32 = new CRC32();
        crc32.update(values);
        final long actualValue = crc32.getValue();
        System.out.println("CRTs " +Integer.toBinaryString((int)res));
        System.out.println("Javas " +Integer.toBinaryString((int) actualValue));
        assertEquals(actualValue, res);
    }`

ilevyor: alternativly casting the long results into ints makes everything work, so I think we should be nudging java to use ints rather than using longs, as otherwise I think we will run into checksum mismatch errors across sdks down the line. CRC32 has (shocker) 32 bit output.

bretambrose: That looks  like the bad conversion I was referring to originally.  If you took the uint32_t from aws_checksums_crc32 converted it to a long (in C) and then returned the long to Java, it would match without having to do bit twiddling.

uint32 -> int64 does not do a sign extension because there's no sign and the destination is guaranteed to hold the correct source value, and so the bit pattern is exactly the same except the upper 32 bits are zero.  But if you do uint32 -> int32 -> int64 then sign extension on int32 -> int64 will fill the upper 32 bits with 1s if the original uint32_t had the high order bit set.  And that's why you end up having to mask the upper 32 bits off there.

src/main/java/software/amazon/awssdk/crt/checksums/Checksums.java
```diff
@@ -18,6 +19,24 @@ public static int crc32c(byte[] input) {
         return crc32c(input, 0, 0, input.length);
     }
 
+    public static int crc32(int input, int prev) {
```
bretambrose, line 22: What are these for?

src/main/java/software/amazon/awssdk/crt/checksums/Checksums.java
ilevyor: That makes sense, I can make that change, but that still raises the question of weather we want to have a 64 bit output, when all the other sdks write 32 bit outputs? isn't this setting the java team up for failure when it comes to to write the checksum to the message trailer?

src/main/java/software/amazon/awssdk/crt/checksums/Checksums.java
ilevyor: Java team asked that we add functionality to pass in ints so that they have an easier time matching the java checksum interface, and this was easier than arguing.

src/main/java/software/amazon/awssdk/crt/checksums/Checksums.java
bretambrose: Well, we're matching the contract of the Java standard library, and to be honest, I think our implementation should literally be an implementation of the Checksum interface: https://docs.oracle.com/javase/8/docs/api/java/util/zip/Checksum.html

I don't think it's unreasonable to expect them to know a crc is 32 bits.

ilevyor: I have no Idea why Java returns a long, they do all their computation with ints, and return this bitmasked long for ??? reason.
https://github.com/himnay/java7-sourcecode/blob/master/java/util/zip/CRC32.java#L84-L89

src/main/java/software/amazon/awssdk/crt/checksums/Checksums.java
bretambrose: I don't think this is matching what they're asking for though.  The update is for bytes not ints.  And honestly let's just implement the checksum interface so that they (and anyone else who ever uses this) don't have to do it themselves.

joviegas:
It would be great if these Classes are [Cloneable](https://docs.oracle.com/javase/9/docs/api/java/lang/Cloneable.html) , this can be done by providing clone() Method.
In SDK Java V2 we need to clone checksum when we do mark and reset.

Example -->
https://github.com/aws/aws-sdk-java-v2/blob/584ccb59e770177aeaa4c3b6bda4e24015b8ece9/core/sdk-core/src/main/java/software/amazon/awssdk/core/checksums/Md5Checksum.java#L50-L84

We will need a this  when we get into testing retries for Java Sdk V2.

src/main/java/software/amazon/awssdk/crt/checksums/CRC32.java
```diff
@@ -0,0 +1,56 @@
...
+        }
+        value = crc32(b, value, off, len);
+    }
+
+    public void update(byte[] b) {
+        value = crc32(b, value, 0, b.length);
+    }
+
+    public void update(int b) {
+        byte[] buf = {(byte)(b & 0x000000ff)};
```
bretambrose, line 51: Can we throw an exception here if  (b < 0 || b > 255)?

src/main/java/software/amazon/awssdk/crt/checksums/CRC32C.java
```diff
@@ -0,0 +1,56 @@
...
+        }
+        value = crc32c(b, value, off, len);
+    }
+
+    public void update(byte[] b) {
+        value = crc32c(b, value, 0, b.length);
+    }
+
+    public void update(int b) {
+        byte[] buf = {(byte)(b & 0x000000ff)};
```
bretambrose, line 51: Same here

src/main/java/software/amazon/awssdk/crt/checksums/CRC32.java
ilevyor: what error should I throw? IlligalArgument?

bretambrose (approved):

PR #359, opened 2021-07-26
Fix memory leak and check it from CI

TingDaoK:
*Issue #, if available:*

- Memory check never actually works for tests!

*Description of changes:*

- Add native memory check for every test
- Fixed several memory leak
- Fixed the tests that not clean up correctly
- Update submodule for another memory leak fix and crash fix

*TODO:*
- use sba_allocator as default allocator for all the native stuff. Some weird crash happens, need more investigation. But, as it doesn't improve the performance dramatically for now, we can save it for later


src/main/java/software/amazon/awssdk/crt/CRT.java
```diff
@@ -251,23 +252,30 @@ private static void jvmInit() {
...
      */
     public static long nativeMemory() {
         return awsNativeMemory();
     }
 
-    private static native long awsNativeMemory();
+    /**
+     * Memory Tracer dump when trace level is 2
+     */
+    public static native void awsMemTracerDump();
```
TingDaoK, line 266: I just added this function in this file

graebm: The dump isn't visible unless logging is enabled at Trace level. Tell users that.

Also remind them how to set the correct system properties, don't assume they've read it in the docs for some other function

Also make the name consistent with the existing public "nativeMemory()" function

```
/**
 * Dump info to logs about all memory currently allocated by native resources.
 * The following system properties must be set to see a dump:
 * aws.crt.memory.tracing must be 1 or 2
 * aws.crt.log.level must be "Trace"
 */
public static void dumpNativeMemory()

File: src/native/aws_signing.c [hunk 2]
```diff
@@ -69,6 +69,10 @@ static void s_cleanup_callback_data(struct s_aws_sign_request_callback_data *cal
     }
 
     if (callback_data->native_request) {
+        struct aws_input_stream *input_stream = aws_http_message_get_body_stream(callback_data->native_request);
+        if (input_stream != NULL) {
+            aws_input_stream_destroy(input_stream);
```
graebm, line 74: Is it ALWAYS the case that the request here owns the input stream?

src/native/crt.c
```diff
@@ -31,12 +31,12 @@
 /* 0 = off, 1 = bytes, 2 = stack traces, see aws_mem_trace_level */
 int g_memory_tracing = 0;
 static struct aws_allocator *s_init_allocator(void) {
+    struct aws_allocator *allocator = aws_default_allocator();
+    // struct aws_allocator *sba_allocator = aws_small_block_allocator_new(allocator, true);
```
graebm, line 35: don't commit commented-out code

src/native/crt.c
```diff
@@ -229,8 +229,10 @@ static void s_jni_atexit_strict(void) {
 
     if (g_memory_tracing) {
         struct aws_allocator *tracer_allocator = aws_jni_get_allocator();
-        aws_mem_tracer_destroy(tracer_allocator);
+        s_allocator = aws_mem_tracer_destroy(tracer_allocator);
     }
+    // aws_small_block_allocator_destroy(s_allocator);
```
graebm, line 234: don't commit commented-out code

src/native/crt.c
```diff
@@ -229,8 +229,10 @@ static void s_jni_atexit_strict(void) {
 
     if (g_memory_tracing) {
         struct aws_allocator *tracer_allocator = aws_jni_get_allocator();
-        aws_mem_tracer_destroy(tracer_allocator);
+        s_allocator = aws_mem_tracer_destroy(tracer_allocator);
```
graebm, line 232: trivial:
~~s_allocator =~~ aws_mem_tracer_destroy(tracer_allocator);
it's weird that you're setting s_allocator here. there's no reason to do this. It gets set NULL at the end of the function.

src/main/java/software/amazon/awssdk/crt/CRT.java
```diff
@@ -176,29 +178,24 @@ private static void loadLibraryFromJar() {
 
             // load the shared lib from the temp path
             System.load(libTempPath.toString());
-        }
-        catch (CrtRuntimeException crtex) {
```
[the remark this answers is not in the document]
TingDaoK: Nope, we don't have format checker for java

src/native/crt.c
TingDaoK: Still trying to debug what's happening with the sba allocator and it leads to a crash sometimes.

src/native/aws_signing.c
TingDaoK: I think so. The native request is created for signing. (callback_data owns the native request) https://github.com/awslabs/aws-crt-java/blob/main/src/native/aws_signing.c#L381-L382.
And the native request owns the body stream https://github.com/awslabs/aws-crt-java/blob/main/src/native/http_request_utils.c#L326-L332

src/test/java/software/amazon/awssdk/crt/test/CredentialsProviderTest.java
```diff
@@ -213,8 +213,7 @@ public void testCreateDestroyProfile_InvalidProfile() throws IOException {
         ProfileCredentialsProvider.Builder builder = ProfileCredentialsProvider.builder()
                 .withConfigFileNameOverride(confPath.toString()).withCredentialsFileNameOverride(credsPath.toString());
 
-        try {
-            builder.build();
+        try (ProfileCredentialsProvider provider = builder.build()) {
```
justinboswell, line 216: HA

src/test/java/software/amazon/awssdk/crt/test/CrtTestFixture.java
```diff
@@ -33,6 +35,11 @@ public final CrtTestContext getContext() {
 
     @Before
     public void setup() {
+        if (System.getProperty("aws.crt.aws_trace_log_per_test") != null) {
```
justinboswell, line 38: Isn't this the same code as the fixture?

justinboswell (approved):
Fix n ship

src/test/java/software/amazon/awssdk/crt/test/CrtTestFixture.java
TingDaoK: yep, we have two fixtures. One for the S3 native client test, the other one is for the regular CRT test

src/test/java/software/amazon/awssdk/crt/test/CredentialsProviderTest.java
TingDaoK: https://github.com/awslabs/aws-crt-java/pull/359/files#diff-7a1d883433da59ad94de2b6989b5ae4042739a36ab35dbbd80baf1c2ccf52d71R68-R71 this is actually another fix for it

justinboswell: This is the correct fix, should always use try-with-resource if we can

PR #365, opened 2021-08-12
Publish android

TingDaoK:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/android/crt/build.gradle
+++ b/android/crt/build.gradle
@@ -197,37 +197,6 @@ afterEvaluate {
                 }
                 version = android.defaultConfig.versionName
             }
-
-            debug(MavenPublication) {
-                from components.debug
-
-                groupId = 'software.amazon.awssdk.crt'
-                artifactId = 'aws-crt-android'
-                pom {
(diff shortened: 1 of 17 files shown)


android/crt/build.gradle
```diff
@@ -188,16 +172,31 @@ afterEvaluate {
...
+                }
                 version = android.defaultConfig.versionName
             }
-
-            debug(MavenPublication) {
-                from components.debug
-
-                groupId = 'software.amazon.awssdk.crt'
-                artifactId = 'aws-crt-android'
-                version = android.defaultConfig.versionName + '-SNAPSHOT'
```
TingDaoK, line 199: The way we publish doesn't support snapshot yet, so, remove the debug publish which does nothing. Haven't got deep into it. The whole idea for this PR is to get it work first, we can come back later, when/if we want to completely move to gradle.

aajtodd: I believe Ian said the same thing that snapshots don't work. We don't really need to publish snapshots anyway.

aajtodd (approved):

Issue #334, opened 2021-06-18
android libs not bundled


aajtodd:
CRT fails to init on android (emulator or otherwise):

```
W/System.err: Unable to unpack AWS CRT lib: java.io.IOException: Unable to open library in jar for AWS CRT: /android/x86/libaws-crt-jni.so
    java.io.IOException: Unable to open library in jar for AWS CRT: /android/x86/libaws-crt-jni.so
```

I don't see any android platform dir in the release jar (neither `0.12.5` or latest `0.13.1`):

```
> jar -xf ~/Downloads/aws-crt-0.13.1.jar                                                                                                                                                                                    
> ls
META-INF                com                     libaws-c-event-stream.a libs2n.a                linux                   osx                     software                windows
```

aajtodd:
closed by #367. Android support is provided by a new artifact `aws-crt-android`.

PR #367, opened 2021-08-16
Fix Android support

aajtodd:
*Issue #, if available:*
#334 

*Description of changes:*

* Fix Android build
    * Update Gradle and Android plugin to latest
    * Update to LTS version of NDK
    * Drop the Android API requirement from 26 -> 24 which required removing `java.nio.File/Path` APIs and few other minor API changes.
* Add an `android` job to CI which builds the Android project. This will also run the lint checks to verify API compatibility with the declared `minSdk` version.  
    * This runs directly on Ubuntu runner for now. I created an Android docker image in [aws-crt-builder:corretto](https://github.com/awslabs/aws-crt-builder/tree/corretto). When we figure out the issues with Docker and GitHub we can integrate those changes.
* (conditionally) enables the Android build based on if you have configured the Android SDK. This means most developers (and CI) will not need to configure Android to contribute to `aws-crt-java`. CI will handle the rest.
* Publish the Android `.aar` as a separate publication (it will get it's own maven coordinates at `software.amazon.awssdk.crt:aws-crt-android`
    * This uses Gradle to do the publication which has some complications with the Nexus publishing plugin. Currently we use maven to build and release. The maven nexus plugin has support to do separate publish/closeAndRelease tasks. This is currently a missing feature in the Gradle plugin. Instead we publish and close the staged nexus repo from Gradle and then use maven to do the release of both publications.

### DIFF
--- a/android/crt/build.gradle
+++ b/android/crt/build.gradle
@@ -1,8 +1,6 @@
 import java.util.regex.Pattern
 
 apply plugin: 'com.android.library'
-//apply plugin: 'kotlin-android'
-//apply plugin: 'kotlin-android-extensions'
 
 // Before configuring, make sure libcrypto is installed
 preBuild {
@@ -96,11 +94,13 @@ android {
     buildTypes {
         debug {
             versionNameSuffix = gitVersionTag()
+            buildConfigField("String", "VERSION_NAME", "\"" + gitVersionName() + "\"")
         }
         release {
             minifyEnabled false
             proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
             versionNameSuffix ""
+            buildConfigField("String", "VERSION_NAME", "\"" + gitVersionName() + "\"")
         }
     }
 
@@ -122,7 +122,6 @@ build.dependsOn preBuild
 
 dependencies {
     implementation 'androidx.appcompat:appcompat:1.1.0'
-    implementation 'androidx.core:core-ktx:1.2.0'
     androidTestImplementation 'androidx.test.ext:junit:1.1.1'
     androidTestImplementation 'androidx.test:runner:1.2.0'
     androidTestImplementation 'androidx.test:monitor:1.2.0@aar'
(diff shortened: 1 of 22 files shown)


android/crt/build.gradle
```diff
@@ -115,18 +116,21 @@ android {
     compileOptions {
         sourceCompatibility = 1.8
         targetCompatibility = 1.8
+        coreLibraryDesugaringEnabled true
```
justinboswell, line 119: Can you put a comment here on what this does, or link to docs? I need to know EVERYTHING about whatever "desugaring" is

src/main/java/software/amazon/awssdk/crt/CRT.java
```diff
@@ -150,19 +151,30 @@ private static void loadLibraryFromJar() {
             String prefix = "AWSCRT_" + new Date().getTime();
             String libraryName = System.mapLibraryName(CRT_LIB_NAME);
             String libraryPath = "/" + getOSIdentifier() + "/" + getArchIdentifier() + "/" + libraryName;
-            Path libTempPath = Files.createTempFile(tmpdirPath, prefix, libraryName);
+
+            File tempSharedLib = File.createTempFile(prefix, libraryName, tmpdirFile);
+            System.out.println(tempSharedLib.getAbsolutePath());
```
justinboswell, line 156: This line can go

justinboswell (approved):
Fix-n-shipperooni

TingDaoK (approved):

android/crt/build.gradle
aajtodd: Added a comment. It allows targeting a subset of Java 8 API's on older versions of Android. The Android tooling will add a backport library behind the scenes if necessary. For a library this really just gets around lint issues since the desugaring should happen when building an APK not an aar. 

See:
* https://jakewharton.com/d8-library-desugaring/
* https://developer.android.com/studio/write/java8-support#library-desugaring
* https://developer.android.com/studio/write/java8-support-table

PR #383, opened 2021-09-10
Bind PKCS#11 TLS options to C

graebm:
Design decisions:
- rename ~Pkcs11TlsOptions~ -> `TlsContextPkcs11Options` for consistency with C
- `TlsContextPkcs11Options` is a `CrtResource`, but doesn't actually "carry around" its own native resource.
    - There's no real benefit in tying the lifetime of the native object to the java object, so don't
    - But being a `CrtResource`, lets it fit nicely into the rest of our `CrtResource` "ecosystem":
      - It can properly reference other resources.
      - It can be declared in the same try-blocks as other resources. etc

### DIFF
--- a/src/native/tls_context_options.c
+++ b/src/native/tls_context_options.c
@@ -9,6 +9,7 @@
 #include <aws/io/tls_channel_handler.h>
 
 #include "crt.h"
+#include "tls_context_pkcs11_options.h"
 
 /* Have to wrap the native struct so we can manage string lifetime */
 struct jni_tls_ctx_options {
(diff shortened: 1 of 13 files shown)


src/native/tls_context_options.c
```diff
@@ -125,7 +131,41 @@ jlong JNICALL Java_software_amazon_awssdk_crt_io_TlsContextOptions_tlsContextOpt
...
+            goto on_error;
+        }
+
+        if (aws_tls_ctx_options_init_client_mtls_with_pkcs11(&tls->options, allocator, tls->pkcs11_options)) {
+            aws_jni_throw_runtime_exception(env, "aws_tls_ctx_options_init_client_mtls_with_pkcs11 failed");
+            goto on_error;
+        }
+    }
+#if defined(__APPLE__)
+    else if (jni_pkcs12_path && jni_pkcs12_password) {
```
graebm, line 147: I moved this Apple/pkcs12 block of code up as a bug-fix. It used to be further down.

This fix ensures we do the `init()` calls before any of the `set()` calls on `aws_tls_ctx_options`.

bretambrose (approved):

PR #386, opened 2021-09-24
Set release target version to JDK 8, when building with JDK > 8.

millems:
This fixes an issue where customers on JDK 8 would get a NoSuchMethodException for ByteBuffer.position.


TingDaoK (approved):
Thank you for fixing it!

justinboswell (approved):

PR #391, opened 2021-10-11
sts assume role provider

kggilmer:
*Issue #, if available:*
https://github.com/awslabs/aws-sdk-kotlin/issues/19

*Description of changes:*
* Adds native code and Java class to expose STS assume role credential provider from `aws-c-auth` to Java

*Testing Done*
* Added unit test for failure case
* Attempted to create environment for positive test case:
```
* Add a role to test AWS account with S3 access
* In test, pass either static or profile credential provider to user in test account
* Specify test role in roleArn parameter

Expected:
See that credential provider would return valid generated credentials mapping to test role.  

Actual:
 Debug logging inside of CRT states that HTTP 403 (forbidden) returned from service
```
Unsure of which part of my test setup is incorrect.  Looking for documented test cases or other info leading me to a successful test outcome.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,6 +39,9 @@ To send us a pull request, please:
 GitHub provides additional document on [forking a repository](https://help.github.com/articles/fork-a-repo/) and
 [creating a pull request](https://help.github.com/articles/creating-a-pull-request/).
 
+### Formatting Changes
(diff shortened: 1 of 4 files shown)


justinboswell:
Windows warnings and clang-format need to be fixed, but otherwise looks good to go!

PR #398, opened 2021-10-20
HTTP2 Phase 1

TingDaoK:
- Phase 1 of http2 binding.
- Some of the functions are declared, but not implemented as it's already complicated to review
- And the http message and headers are still on the way
- Connection manager will not do anything special for http2 right now. 


graebm (approved):

src/main/java/software/amazon/awssdk/crt/http/HttpClientConnection.java
```diff
@@ -23,6 +25,42 @@
...
+            enumMapping.put(HTTP_1_1.getValue(), HTTP_1_1);
+            enumMapping.put(HTTP_2.getValue(), HTTP_2);
+
+            return enumMapping;
+        }
+
+        public int getValue() {
+            return protocolVersion;
+        }
+    }
```
TingDaoK, line 63: Since it's used outside connection, feel like it's better to move it out as a separate file in http/

src/main/java/software/amazon/awssdk/crt/http/Http2ClientConnection.java
```diff
@@ -0,0 +1,176 @@
...
+     *
+     * If `manualWindowManagement` is false, this call will have no effect. The
+     * connection maintains its flow-control windows such that no back-pressure is
+     * applied and data arrives as fast as possible.
+     *
+     * If you are not connected, this call will have no effect.
+     *
+     * Crashes when the connection is not http2 connection. The limit of the Maximum
+     * Size is 2**31 - 1. If the increment size cause the connection flow window
+     * exceeds the Maximum size, this call will result in the connection lost.
```
dagnir, line 144: Can it just throw a validation exception instead of losing the connection?

src/main/java/software/amazon/awssdk/crt/http/Http2ClientConnection.java
```diff
@@ -0,0 +1,176 @@
...
+     *
+     * @param increment_size The size to increment for the connection's flow control
+     *                       window
+     */
+    public void updateConnectionWindow(long incrementSize) {
```
dagnir, line 149: might be nice to return the new window size

src/main/java/software/amazon/awssdk/crt/http/Http2Stream.java
```diff
@@ -0,0 +1,30 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+package software.amazon.awssdk.crt.http;
+
+import software.amazon.awssdk.crt.CrtRuntimeException;
+
+public class Http2Stream extends HttpStream {
```
dagnir, line 10: How does sending stream level frames work, for example stream window size. Will that be supported?

src/main/java/software/amazon/awssdk/crt/http/Http2ClientConnection.java
```diff
@@ -0,0 +1,176 @@
...
+            final byte[] debugData) {
+        throw new CrtRuntimeException("Unimplemented");
+    }
+
+    public void sendGoAway(final Http2ErrorCode Http2ErrorCode, final boolean allowMoreStreams) {
+        throw new CrtRuntimeException("Unimplemented");
+    }
+
+    /**
+     * Increment the connection's flow-control window to keep data flowing.
```
dagnir, line 127: In the manual management mode, how does the user know whenever the window size shrinks so they can call this method?

src/main/java/software/amazon/awssdk/crt/http/HttpClientConnection.java
```diff
@@ -23,6 +25,42 @@
...
+        HTTP_2(3);
+
+        private int protocolVersion;
+        private static Map<Integer, ProtocolVersion> enumMapping = buildEnumMapping();
+
+        ProtocolVersion(int value) {
+            protocolVersion = value;
+        }
+
+        public static ProtocolVersion getEnumValueFromInteger(int value) {
```
dagnir, line 43: nit: not sure how closely you try to follow the Java SDK conventions but this is normally named `fromValue`

src/main/java/software/amazon/awssdk/crt/http/Http2Stream.java
dagnir: Is there a way to get the connection from a stream?

dagnir: What do the `ByteBuffers` returned to the caller represent? Is it just the raw payload of the `DATA` frames?

src/main/java/software/amazon/awssdk/crt/http/Http2ConnectionSetting.java
```diff
@@ -0,0 +1,77 @@
...
+    public ID getId() {
+        return id;
+    }
+
+    /**
+     * HTTP/2 connection settings.
+     *
+     * value is limited from 0 to UINT32_MAX (RFC-7540 6.5.1)
+     */
+    public Http2ConnectionSetting(ID id, long value) {
```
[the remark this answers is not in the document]
TingDaoK: Sure, we can have getters/setters for this class. A good call! But I am not sure what you mean by "have a single `HttpConnectionSettings` object".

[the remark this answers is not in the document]
TingDaoK: Good call! Will do!

src/main/java/software/amazon/awssdk/crt/http/Http2ClientConnection.java
```diff
@@ -0,0 +1,176 @@
...
+     * Send a SETTINGS frame. SETTINGS will be applied locally when SETTINGS ACK is
+     * received from peer.
+     *
+     * @param settings The array of settings to change. Note: each setting has its
+     *                 boundary.
+     *
+     * @return When this future completes without exception, the peer has
+     *         acknowledged the settings and the change has been applied.
+     */
+    public CompletableFuture<Void> changeSettings(final Http2ConnectionSetting settings[]) {
```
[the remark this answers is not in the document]
TingDaoK: > Might have missed it, but there a way to specify the SETTINGS values that are sent on connection negotiation?

You didn't miss it... We haven't exposed that yet. We have not implement the connection pool and the initial settings I am thinking about should be exposed with the connection pool. So, you cannot really set that for now.

src/main/java/software/amazon/awssdk/crt/http/Http2ClientConnection.java
TingDaoK: We have an async API under the C level. So, we don't really know the new window size until the thread under the hood invoke. Image a case we receiving data while user updating the window. Or, multiple thread update the windows size at the same time. So, nothing much we can do to prevent this happen before head. From C level, we decided to just close the connection.

src/main/java/software/amazon/awssdk/crt/http/Http2ClientConnection.java
TingDaoK: ditto

src/main/java/software/amazon/awssdk/crt/http/Http2ClientConnection.java
```diff
@@ -0,0 +1,176 @@
...
+/**
+ * This class wraps aws-c-http to provide the basic HTTP/2 request/response
+ * functionality via the AWS Common Runtime.
+ *
+ * Http2ClientConnection represents a single connection to a HTTP/2 service
+ * endpoint.
+ *
+ * This class is not thread safe and should not be called from different
+ * threads.
+ */
```
dagnir, line 21: We should discuss this API in depth. It seems to me like this represents an H2 connection and not a stream.

From a usage perspective, this doesn't look like the right level of abstraction for Java or the other SDKs. I think the CRT should manage the stream aspect internally so that each SDK doesn't need to implement it in their own libraries.

i.e.  the return value of calling `connectionManager.acquireConnection()` should be an HTTP/2 stream under the hood and not a physical connection.

If a user wants to make 2 concurrent requests, they might do something like

```
HttpConnection conn1 = connectionManager.acquireConnection().join();
conn1.makeRequest(...).activate();

HttpConnection conn2 = connectionManager.acquireConnection().join();
conn1.makeRequest(...).activate();
```

It seems to me like in the current version of the code, this results in two HTTP/2 connections being used, rather than both requests being interleaved on on the same connection over two different streams.

src/main/java/software/amazon/awssdk/crt/http/Http2Stream.java
TingDaoK: - Stream window size is already supported as a generic function for both http1 stream and http2 stream: https://github.com/awslabs/aws-crt-java/blob/main/src/main/java/software/amazon/awssdk/crt/http/HttpStream.java#L60
- We don't support getting the connection from stream, but I am okay to support it
- I am not sure about what `ByteBuffers` you are referring to

src/main/java/software/amazon/awssdk/crt/http/Http2ClientConnection.java
TingDaoK: that's a good callout! I'll need to think more about it. We currently don't have any feedback when windows shrinks, user will need to track the window themselves, which may not be a good option, especially the connection manager will handle all the connections.

src/main/java/software/amazon/awssdk/crt/http/Http2ClientConnection.java
```diff
@@ -0,0 +1,176 @@
...
+     *                         be acknowledged and the GOAWAY's Last-Stream-ID will
+     *                         be set to a max value. If false, new peer-initiated
+     *                         streams will be ignored and the GOAWAY's
+     *                         Last-Stream-ID will be set to the latest acknowledged
+     *                         stream.
+     * @param debugData        Optional debug data to send. Size must not exceed
+     *                         16KB. null is acceptable to not include debug data.
+     */
+    public void sendGoAway(final Http2ErrorCode Http2ErrorCode, final boolean allowMoreStreams,
+            final byte[] debugData) {
```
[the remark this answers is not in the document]
TingDaoK: We actually prefer to use `byte[]` than `ByteBuffer`. It helps us to improve the performance without needs to go back to the JNI boundary to fetch data. Also for the fragmentation and lifetime, which make `byte[]` easier for JNI to handle.

PR #403, opened 2021-10-28
HTTP2 Phase 2

TingDaoK:
- Phase 2 of http2 binding.
  - Http2 stream support
  - http2 connections done

#### TODO:

  - http message and headers are still on the way


src/main/java/software/amazon/awssdk/crt/http/Http2ClientConnection.java
```diff
@@ -65,15 +68,19 @@ public Http2ClientConnection(long connectionBinding) {
      * @return When this future completes without exception, the peer has
      *         acknowledged the settings and the change has been applied.
      */
-    public CompletableFuture<Void> changeSettings(final Http2ConnectionSetting settings[]) {
-        throw new CrtRuntimeException("Unimplemented");
+    public CompletableFuture<Void> changeSettings(final List<Http2ConnectionSetting> settings) {
+        CompletableFuture<Void> completionFuture = new CompletableFuture<>();
+        http2ClientConnectionChangeSettings(getNativeHandle(), completionFuture,
```
JonathanHenson, line 73: Can’t pass futures to JNI and support Android, also this is a coupling we don’t need. I understand if for consistency, I guess, but this should be a functor object, not a future. You can always have a futures api that sits on top of that.

TingDaoK: Oh, I see the code around uses `AsyncCallback` class. The signer is the only place we use Future directly. I can use the `AysncCallback` class for sure

TingDaoK: https://github.com/awslabs/aws-crt-java/pull/367/files#diff-210ac8ee910f94aacc1e516387a4a2304b24cd5787216d51add1486a9e736a37R62 Seems like we are targetting 24 as minimal requirement for andriod. So, should be good?

graebm (approved):
fix & ship

src/main/java/software/amazon/awssdk/crt/http/Http2ClientConnection.java
```diff
@@ -26,19 +30,9 @@
...
-        STREAM_CLOSED(5),
-        FRAME_SIZE_ERROR(6),
-        REFUSED_STREAM(7),
-        CANCEL(8),
-        COMPRESSION_ERROR(9),
-        CONNECT_ERROR(10),
-        ENHANCE_YOUR_CALM(11),
-        INADEQUATE_SECURITY(12),
-        HTTP_1_1_REQUIRED(13);
+        PROTOCOL_ERROR(1), INTERNAL_ERROR(2), FLOW_CONTROL_ERROR(3), SETTINGS_TIMEOUT(4), STREAM_CLOSED(5),
```
dagnir, line 33: How are these errors surfaced back to the caller?

Curious about things where the peer might send an error in the middle of a request like in a `GOAWAY`

src/test/java/software/amazon/awssdk/crt/test/Http2ClientConnectionTest.java
```diff
@@ -47,6 +52,65 @@ public void testHttp2ConnectionGetVersion() throws Exception {
...
+                Assert.assertTrue(conn.getVersion() == EXPECTED_VERSION);
+                List<Http2ConnectionSetting> settings = new ArrayList<Http2ConnectionSetting>();
+                settings.add(new Http2ConnectionSetting(Http2ConnectionSetting.ID.ENABLE_PUSH, 0));
+                settings.add(new Http2ConnectionSetting(Http2ConnectionSetting.ID.ENABLE_PUSH, 0));
+                conn.changeSettings(settings);
+            }
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+
```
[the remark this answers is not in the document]
TingDaoK: good catch, that's a bug. I forgot to wait for the future to finish.

src/main/java/software/amazon/awssdk/crt/http/Http2ClientConnection.java
TingDaoK: >@TODO: bindings for getters of local/remote setting and goaway.
https://github.com/awslabs/aws-crt-java/pull/398/files#diff-f99e160a2ce01eefcfa64cf86b4611d7eae8cf2851f37627682b7334f2e9a269R154
If it's something we want, I'll add the binding soon.

TingDaoK: We also have `on_goaway_received` callback in C, but it will be bound after the connection pool http2 support.

src/test/java/software/amazon/awssdk/crt/test/Http2ClientConnectionTest.java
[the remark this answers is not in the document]
TingDaoK: good catch, that's a bug.

src/main/java/software/amazon/awssdk/crt/http/Http2ClientConnection.java
```diff
@@ -147,7 +164,34 @@ public void sendGoAway(final Http2ErrorCode Http2ErrorCode, final boolean allowM
...
+    @Override
+    public Http2Stream makeRequest(HttpRequest request, HttpStreamResponseHandler streamHandler)
+            throws CrtRuntimeException {
+        if (isNull()) {
+            throw new IllegalStateException("Http2ClientConnection has been closed, can't make requests on it.");
+        }
+
+        Http2Stream stream = http2ClientConnectionMakeRequest(getNativeHandle(), request.marshalForJni(),
+                request.getBodyStream(), new HttpStreamResponseHandlerNativeAdapter(streamHandler));
+        return stream;
```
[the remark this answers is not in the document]
TingDaoK: I haven't tested it, but I think we can support this case.

PR #406, opened 2021-11-02
HTTP/2 Request (HTTP/2 Phase3)

TingDaoK:
*Issue #, if available:*

*Description of changes:*
- Basically bare minimal for http/2 request, user is responsible to make sure the requests are following the protocol requirement, otherwise, the other side will reject the request.
- Use http/2 request can help the request not going through the transform from http/1 request to http/2
- And HTTP/2 Request will **not apply** those auto fix we did for http/1 request, which like converting headers to lower case, and get host out, etc..
- Any suggestions around will be super appreciated, feeling bad about my OOD....

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/http/Http2ClientConnection.java
+++ b/src/main/java/software/amazon/awssdk/crt/http/Http2ClientConnection.java
@@ -29,9 +29,19 @@ public class Http2ClientConnection extends HttpClientConnection {
      * (RFC-7540 7).
      */
     public enum Http2ErrorCode {
-        PROTOCOL_ERROR(1), INTERNAL_ERROR(2), FLOW_CONTROL_ERROR(3), SETTINGS_TIMEOUT(4), STREAM_CLOSED(5),
-        FRAME_SIZE_ERROR(6), REFUSED_STREAM(7), CANCEL(8), COMPRESSION_ERROR(9), CONNECT_ERROR(10),
-        ENHANCE_YOUR_CALM(11), INADEQUATE_SECURITY(12), HTTP_1_1_REQUIRED(13);
+        PROTOCOL_ERROR(1),
+        INTERNAL_ERROR(2),
+        FLOW_CONTROL_ERROR(3),
+        SETTINGS_TIMEOUT(4),
+        STREAM_CLOSED(5),
+        FRAME_SIZE_ERROR(6),
+        REFUSED_STREAM(7),
+        CANCEL(8),
+        COMPRESSION_ERROR(9),
+        CONNECT_ERROR(10),
+        ENHANCE_YOUR_CALM(11),
+        INADEQUATE_SECURITY(12),
+        HTTP_1_1_REQUIRED(13);
 
         private int errorCode;
 
--- a/src/main/java/software/amazon/awssdk/crt/http/Http2ClientConnection.java
+++ b/src/main/java/software/amazon/awssdk/crt/http/Http2ClientConnection.java
@@ -29,19 +29,9 @@ public class Http2ClientConnection extends HttpClientConnection {
      * (RFC-7540 7).
      */
     public enum Http2ErrorCode {
-        PROTOCOL_ERROR(1),
-        INTERNAL_ERROR(2),
-        FLOW_CONTROL_ERROR(3),
-        SETTINGS_TIMEOUT(4),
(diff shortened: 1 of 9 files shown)


src/native/http_request_utils.c
```diff
@@ -214,31 +214,39 @@ int aws_marshal_http_headers_to_dynamic_buffer(
...
+    if (!aws_byte_cursor_read_be32(request_blob, &version)) {
+        aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+    }
+    return version;
+}
+
 static inline int s_unmarshal_http_request(struct aws_http_message *message, struct aws_byte_cursor *request_blob) {
     uint32_t field_len = 0;
+    if (aws_http_message_get_protocol_version(message) != AWS_HTTP_VERSION_2) {
+        if (!aws_byte_cursor_read_be32(request_blob, &field_len)) {
```
graebm, line 228: add comment like "HTTP/1 puts method and path first, but those are just headers in HTTP/2"

src/native/http_request_utils.c
```diff
@@ -282,7 +290,12 @@ int aws_apply_java_http_request_changes_to_native_request(
...
-    result = s_unmarshal_http_request(message, &marshalled_cur);
+    enum aws_http_version version = s_unmarshal_http_request_to_get_version(&marshalled_cur);
+    if (version != aws_http_message_get_protocol_version(message)) {
+        result = aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+    } else {
+        result = s_unmarshal_http_request(message, &marshalled_cur);
+    }
     (*env)->ReleasePrimitiveArrayCritical(env, marshalled_request, marshalled_request_data, 0);
 
     if (jni_body_stream) {
```
graebm, line 301: look into this.
It's a little weird that we continue to do stuff after `result` is AWS_OP_ERR

graebm (approved):
fix & ship

src/main/java/software/amazon/awssdk/crt/http/Http2ClientConnection.java
```diff
@@ -183,7 +184,7 @@ public void updateConnectionWindow(long incrementSize) {
      *         the user thread making this request when it's done.
      */
     @Override
-    public Http2Stream makeRequest(HttpRequest request, HttpStreamResponseHandler streamHandler)
+    public Http2Stream makeRequest(HttpRequestBase request, HttpStreamResponseHandler streamHandler)
```
dagnir, line 187: Take `Http2Request` here?

src/main/java/software/amazon/awssdk/crt/http/Http2Request.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package software.amazon.awssdk.crt.http;
+
+import software.amazon.awssdk.crt.http.HttpClientConnection.ProtocolVersion;
+
+/**
+ * Represents a single Client Request to be sent on a HTTP connection
+ */
+public class Http2Request extends HttpRequestBase {
```
dagnir, line 13: When calling `makeRequest` with a normal `HttpRequest` like this:

```
HttpRequest request = new HttpRequest("GET", "/");
http2Connection.makeRequest(request, ...);
```

Does CRT do all of the conversions for me into the HTTP/2 request; for example setting the `:method` and `:path` headers?

dagnir: I just did some testing against Kinesis and it doesn't look like it :(. Can this be supported? Again, the goal I think is that `HttpRequest` can be constructed and executed without the user having to worry whether they're working with H2 or HTTP/1.1 under the hood.

TingDaoK: We do support converting a HTTP/1 request to HTTP/2, as long as you created a HTTP/2 connection. I guess you are not getting a HTTP/2 connection.

src/main/java/software/amazon/awssdk/crt/http/Http2ClientConnection.java
TingDaoK: That's an override. The native code has different function for `http2ClientConnectionMakeRequest` vs `httpClientConnectionMakeRequest`, which makes sure if you makes a request from a Http2Connection, Http2Stream will be returned.

PR #409, opened 2021-11-04
H2 elasticurl

TingDaoK:
*Description of changes:*
- HTTP2 Java Elasticurl
- Rename ProtocolVersion to HttpVersion and put it into a separate class


pom.xml
```diff
@@ -278,6 +278,18 @@
...
+            <artifactId>commons-cli</artifactId>
+            <version>1.4</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-compress</artifactId>
+            <version>1.19</version>
+            <scope>test</scope>
+        </dependency>
```
graebm, line 292: Does this introduce dependencies to all of aws-crt-java?

If so, is there any way we can structure this so the dependencies are only taken on by Elasticurl?
Like maybe Elasticurl should move out of tests, so it can have its own pom.xml file with its own dependencies?

src/test/java/software/amazon/awssdk/crt/test/Elasticurl.java
```diff
@@ -0,0 +1,362 @@
...
+        cliOpts.addOption(Option.builder("o").longOpt("output").hasArg().argName("file")
+                .desc("dumps content-body to FILE instead of stdout.").build());
+        cliOpts.addOption(Option.builder("t").longOpt("trace").hasArg().argName("file")
+                .desc("dumps logs to FILE instead of stderr.").build());
+        cliOpts.addOption(Option.builder("p").longOpt("alpn").hasArgs().argName("str")
+                .desc("protocol for ALPN. May be specified multiple times.").build());
+        cliOpts.addOption(null, "http1_1", false, "HTTP/1.1 connection required.");
+        cliOpts.addOption(null, "http2", false, "HTTP/2 connection required.");
+        cliOpts.addOption(Option.builder("v").longOpt("verbose").hasArg().argName("str")
+                .desc("logging level (ERROR|INFO|DEBUG|TRACE) default is none.").build());
```
graebm, line 91: trivial: there's "WARN" too. not sure if the other Elasticurls support that

integration-testing/java_elasticurl_runner.py
```diff
@@ -0,0 +1,45 @@
+import sys
+import os
+import subprocess
+import shlex
+import subprocess
+
+TIMEOUT = 100
+# Runner for elasticurl integration tests
```
graebm, line 8: this runner thing is clever ... very clever

pom.xml
TingDaoK: The test dependency will not be included during runtime. Should not be a concern when we ship aws-crt-java to customers. But, a really good call!
I also tested the previous release, the junit and mockito are not included as well:
```
jdeps -s aws-crt-0.15.9.jar 
aws-crt-0.15.9.jar -> java.base 
```

src/test/java/software/amazon/awssdk/crt/test/Elasticurl.java
TingDaoK: Nope, all the others don't have `WARN`...

PR #410, opened 2021-11-05
Signable trailer

ilevyor:
add bindings for aws-c-auths implementation of sigv4/a trailing headers.


src/main/java/software/amazon/awssdk/crt/http/HttpRequest.java
```diff
@@ -123,22 +125,21 @@ public HttpRequestBodyStream getBodyStream() {
...
      */
     public byte[] marshalForJni() {
         int size = 0;
         size += BUFFER_INT_SIZE + method.length();
         size += BUFFER_INT_SIZE + encodedPath.length();
-        size += (BUFFER_INT_SIZE * 2) * headers.size();
 
-        for(HttpHeader header : headers) {
+        for (HttpHeader header : headers) {
             if (header.getNameBytes().length > 0) {
```
TingDaoK, line 141: That feels weird... We silently rule out the empty headers for users, which I don't think we should do... We should just let it pass by and fail hard from either http level or the other side.

So, maybe just remove this check???

src/native/aws_signing.c
```diff
@@ -249,6 +252,14 @@ static void s_aws_chunk_signing_complete(struct aws_signing_result *result, int
     s_cleanup_callback_data(callback_data);
 }
 
+static void s_aws_chunk_signing_complete(struct aws_signing_result *result, int error_code, void *userdata) {
+    s_aws_chunk_like_signing_complete(result, error_code, userdata);
+}
+
+static void s_aws_trailing_headers_signing_complete(struct aws_signing_result *result, int error_code, void *userdata) {
+    s_aws_chunk_like_signing_complete(result, error_code, userdata);
+}
```
TingDaoK, line 261: Trivial: maybe instead of this two functions, just pass the `s_aws_chunk_like_signing_complete` to both chunk and trailing headers signing.

src/main/java/software/amazon/awssdk/crt/auth/signing/AwsSigner.java
```diff
@@ -100,18 +110,38 @@
...
+     * Signs a body chunk according to the supplied signing configuration
+     * 
+     * @param chunkBody         stream of bytes that make up the chunk
+     * @param previousSignature the signature of the previous component of the
+     *                          request: either the request itself for the first
+     *                          chunk, or the previous chunk otherwise
+     * @param config            signing configuration
+     * @return future which will contain a signing result, which provides easier
+     *         access to all signing-related result properties
+     */
```
TingDaoK, line 123: Update the copied/pasted comments

src/native/http_request_utils.c
```diff
@@ -347,6 +372,36 @@ struct aws_http_message *aws_http_request_new_from_java_http_request(
     return NULL;
 }
 
+struct aws_http_headers *aws_http_headers_new_from_java_http_headers(JNIEnv *env, jbyteArray marshalled_headers) {
+    const char *exception_message = NULL;
+    struct aws_http_headers *headers = aws_http_headers_new(aws_jni_get_allocator());
+    if (headers == NULL) {
+        aws_jni_throw_runtime_exception(env, "aws_http_headers_new_from_java_http_headers: Unable to allocate request");
```
TingDaoK, line 379: Well, we don't fail from allocation anymore, but seems like it can still fail from arraylist init. Maybe just change the error message here.

src/native/aws_signing.c
```diff
@@ -492,6 +503,74 @@ void JNICALL Java_software_amazon_awssdk_crt_auth_signing_AwsSigner_awsSignerSig
...
+
+    jint jvmresult = (*env)->GetJavaVM(env, &callback_data->jvm);
+    AWS_FATAL_ASSERT(jvmresult == 0);
+
+    callback_data->java_signing_result_future = (*env)->NewGlobalRef(env, java_signing_result_future);
+    AWS_FATAL_ASSERT(callback_data->java_signing_result_future != NULL);
+
+    callback_data->trailing_headers = aws_http_headers_new_from_java_http_headers(env, marshalled_headers);
+    if (callback_data->trailing_headers == NULL) {
+        aws_jni_throw_runtime_exception(env, "Error building headers");
```
TingDaoK, line 531: The `aws_http_headers_new_from_java_http_headers` already throw exceptions on failure, do we need to throw another one here? I am actually not sure about it....

src/native/aws_signing.c
```diff
@@ -492,6 +503,74 @@ void JNICALL Java_software_amazon_awssdk_crt_auth_signing_AwsSigner_awsSignerSig
...
+        goto on_error;
+    }
+
+    callback_data->java_previous_signature = (*env)->NewGlobalRef(env, java_previous_signature);
+    callback_data->previous_signature = aws_jni_byte_cursor_from_jbyteArray_acquire(env, java_previous_signature);
+
+    callback_data->original_message_signable = aws_signable_new_trailing_headers(
+        allocator, callback_data->trailing_headers, callback_data->previous_signature);
+    if (callback_data->original_message_signable == NULL) {
+        aws_jni_throw_runtime_exception(env, "Failed to create signable from chunk data");
```
TingDaoK, line 550: exception message is saying chunk data, we are doing trailing headers right?

src/native/aws_signing.c
```diff
@@ -492,6 +503,74 @@ void JNICALL Java_software_amazon_awssdk_crt_auth_signing_AwsSigner_awsSignerSig
...
+    }
+
+    /* Sign the native request */
+    if (aws_sign_request_aws(
+            allocator,
+            callback_data->original_message_signable,
+            (struct aws_signing_config_base *)&signing_config,
+            s_aws_trailing_headers_signing_complete,
+            callback_data)) {
+        aws_jni_throw_runtime_exception(env, "Failed to initiate signing process for Chunk");
```
TingDaoK, line 561: Same, chunk seems like a copy/paste to me

TingDaoK (commented):
I know our CRT java don't have a assigned formatter, which is annoying... Can you try not formatting the document, which makes tons of the change coming from the formatting and makes me losing the target I need to review at.

src/main/java/software/amazon/awssdk/crt/http/HttpRequest.java
ilevyor: This is from an old PR, mentioned you in relevant slack conversation. Just fixed a bug where headers size was incorrectly calculated.

src/native/aws_signing.c
ilevyor: Bret suggested this, so that it would be clear I was not accidentally passing the same function to both trailing headers callback, and chunk call back.

src/native/aws_signing.c
ilevyor: good question, @bretambrose do you have thoughts on this?

bretambrose: Can probably drop this throw.

src/main/java/software/amazon/awssdk/crt/http/HttpRequest.java
```diff
@@ -25,41 +26,42 @@
...
+     * object, because accessing this struct from JNI is too slow.
      *
-     *  Requests are marshalled as follows:
+     * Requests are marshalled as follows:
      *
-     *  each string field is:
-     *  [4-bytes BE] [variable length bytes specified by the previous field]
-     *  Each request is then:
-     *  [method][path][header name-value pairs]
+     * each string field is: [4-bytes BE] [variable length bytes specified by the
```
bretambrose, line 60: All this reformatting makes it quite a bit harder to actually see the changes and in some cases reduces the readability (like here).  We don't really do any active formatting on the Java/JS/Python side as far as I know.

If we want to do that, it should be a separate PR.

src/native/http_request_utils.c
```diff
@@ -347,6 +372,36 @@ struct aws_http_message *aws_http_request_new_from_java_http_request(
...
+    const size_t marshalled_headers_length = (*env)->GetArrayLength(env, marshalled_headers);
+
+    jbyte *marshalled_headers_data = (*env)->GetPrimitiveArrayCritical(env, marshalled_headers, NULL);
+    struct aws_byte_cursor marshalled_cur =
+        aws_byte_cursor_from_array((uint8_t *)marshalled_headers_data, marshalled_headers_length);
+    int result = s_unmarshal_http_headers(headers, &marshalled_cur);
+    (*env)->ReleasePrimitiveArrayCritical(env, marshalled_headers, marshalled_headers_data, 0);
+
+    if (result) {
+        exception_message = "aws_http_headers_new_from_java_http_headers: Invalid marshalled headers data.";
```
bretambrose, line 391: can we just throw the exception here instead?

bretambrose (approved):

PR #412, opened 2021-11-08
[S3] credential provider override request-level

TingDaoK:
[S3] credential provider override request-level


src/native/s3_client.c
```diff
@@ -340,14 +340,25 @@ JNIEXPORT jlong JNICALL Java_software_amazon_awssdk_crt_s3_S3Client_s3ClientMake
...
     struct aws_allocator *allocator = aws_jni_get_allocator();
     struct aws_s3_client *client = (struct aws_s3_client *)jni_s3_client;
+    struct aws_credentials_provider *credentials_provider = (struct aws_credentials_provider *)jni_credentials_provider;
+    struct aws_signing_config_aws *signing_config = NULL;
+    bool success = false;
+    if (credentials_provider) {
+        struct aws_byte_cursor region = aws_jni_byte_cursor_from_jbyteArray_acquire(env, jni_region);
+        signing_config = aws_mem_calloc(allocator, 1, sizeof(struct aws_signing_config_aws));
+        aws_s3_init_default_signing_config(signing_config, region, credentials_provider);
+        aws_jni_byte_cursor_from_jbyteArray_release(env, jni_region, region);
```
cmello, line 360: Do we need to copy the `region` before releasing?

TingDaoK: good catch!

cmello (approved):

PR #422, opened 2021-12-22
JNI binding for CRT directory traversal

cmello:
*Issue #, if available:*

*Description of changes:*


ilevyor (approved):

src/native/directory_traversal.c
```diff
@@ -0,0 +1,85 @@
...
+    jobject handler;
+};
+
+static bool s_on_directory_entry(const struct aws_directory_entry *entry, void *user_data) {
+
+    struct directory_traversal_callback_ctx *ctx = user_data;
+    JNIEnv *env = ctx->env;
+
+    jstring path = aws_jni_string_from_cursor(env, &entry->path);
+    jstring relativePath = aws_jni_string_from_cursor(env, &entry->relative_path);
```
graebm, line 36: I was going to say we should check that these succeeded

BUT 

looks like it can only fail if we're out-of-memory
and we don't deal with that anymore
so maybe add an AWS_FATAL_ASSERT to aws_jni_string_from_cursor()
and add a comment in crt.h that the function cannot fail

src/native/directory_traversal.c
```diff
@@ -0,0 +1,85 @@
...
+        ctx->handler,
+        directory_traversal_handler_properties.on_directory_entry_method_id,
+        path,
+        relativePath,
+        (jboolean)entry->file_type & AWS_FILE_TYPE_DIRECTORY,
+        (jboolean)entry->file_type & AWS_FILE_TYPE_SYM_LINK,
+        (jboolean)entry->file_type & AWS_FILE_TYPE_FILE,
+        (jlong)entry->file_size);
+
+    return (bool)callback_result;
```
graebm, line 49: kinda sucks that an exception gets swallowed here, and just stops iteration like nothing went wrong. might be a good idea to modify aws-c-common's implementation so the callback returns an int and we can alert when something goes wrong.

we'd need to do something else to stop, either a special "safe" error like AWS_ERROR_STOP or a separate bool out-param somewhere

ALSO: use aws_jni_check_and_clear_exception() to check whether an exception happened, and properly clear it

src/native/directory_traversal.c
```diff
@@ -0,0 +1,85 @@
...
+    JNIEnv *env,
+    jclass jni_class,
+    jstring path,
+    jboolean recursive,
+    jobject handler) {
+    (void)jni_class;
+
+    struct aws_allocator *allocator = aws_jni_get_allocator();
+
+    const char *path_c_str = (*env)->GetStringUTFChars(env, path, NULL);
```
graebm, line 62: we have a helper function aws_jni_new_string_from_jstring()
also, need to check result for NULL
not sure if you ned to raise a java exception before returning, or if it's already thrown by the underlying JNI function

src/main/java/software/amazon/awssdk/crt/io/DirectoryTraversal.java
```diff
@@ -0,0 +1,32 @@
+package software.amazon.awssdk.crt.io;
+
+/**
+ * This class wraps the directory traversal implementation provided by the CRT.
+ */
+public final class DirectoryTraversal {
```
graebm, line 6: can we mark this "experimental?" or "internal use only"?

CMakeLists.txt
```diff
@@ -196,7 +196,10 @@ if (NOT ANDROID)
         set(JAVA_AWT_LIBRARY NotNeeded)
         set(JAVA_AWT_INCLUDE_PATH NotNeeded)
     endif()
+    file(TO_CMAKE_PATH "$ENV{CMAKE_JAVA_HOME}" CMAKE_JAVA_HOME)
+    set(JAVA_HOME "${CMAKE_JAVA_HOME}") # required because gradle sets JAVA_HOME to the JRE path
```
TingDaoK, line 200: I think we use maven to build an release, why gradle will be invoked here?

cmello: Looks like the CI calls it from here? https://github.com/awslabs/aws-crt-java/blob/main/builder.json#L39 . Definitely it is causing [some checks to fail (e.g. linux-compat al2-x64)](https://github.com/awslabs/aws-crt-java/runs/4657370335?check_suite_focus=true)

I'll move this to a distinct PR and address the comments there, thanks!

CMakeLists.txt
```diff
@@ -196,7 +196,10 @@ if (NOT ANDROID)
...
     message(STATUS "JAVA_HOME=$ENV{JAVA_HOME}")
+    message(STATUS "CMAKE_JAVA_HOME=${CMAKE_JAVA_HOME}")
```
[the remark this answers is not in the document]
cmello: Moved this change to PR #423 so we can address the build issue separately

src/main/java/software/amazon/awssdk/crt/io/DirectoryTraversalHandler.java
```diff
@@ -0,0 +1,28 @@
...
+     * @param isFile entry corresponds to a file.
+     * @param fileSize size of the file.
+     * @return
+     */
+    boolean onDirectoryEntry(final String path,
+                             final String relativePath,
+                             boolean isDirectory,
+                             boolean isSymLink,
+                             boolean isFile,
+                             long fileSize);
```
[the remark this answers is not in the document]
cmello: Fixed, changed to single object instead of N args.

src/main/java/software/amazon/awssdk/crt/io/DirectoryTraversal.java
```diff
@@ -0,0 +1,32 @@
...
+     * will only iterate the contents of the directory, but will not descend into any directories it encounters.
+     *
+     * If recursive is set to true, the traversal is performed post-order, depth-first
+     * (for practical reasons such as deleting a directory that contains subdirectories or files).
+     *
+     * @param path directory to traverse.
+     * @param recursive true to recurse the entire directory, false will only iterate the path specified
+     * @param handler callback to invoke for each file or directory found during the traversal.
+     */
+    public static void traverse(final String path, boolean recursive, final DirectoryTraversalHandler handler) {
```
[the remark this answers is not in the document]
cmello: The `java.util.function.Consumer<T>` looks like a more generic option, but it returns `void`. In order to cancel the iteration the implementer would need to throw an exception which doesn't look good. I think the ideal API would be `Stream<T>` , but as this is an intermediate API consumed by TransferManager, I don't think we should add the overhead at this layer.

src/native/directory_traversal.c
```diff
@@ -0,0 +1,85 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+#include "crt.h"
+#include "java_class_ids.h"
+#include "retry_utils.h"
+#include <aws/common/file.h>
+#include <aws/common/string.h>
+#include <jni.h>
```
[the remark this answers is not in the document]
cmello: Thanks, removed!

src/native/directory_traversal.c
```diff
@@ -0,0 +1,85 @@
...
+
+struct directory_traversal_callback_ctx {
+    JNIEnv *env;
+    jobject handler;
+};
+
+static bool s_on_directory_entry(const struct aws_directory_entry *entry, void *user_data) {
+
+    struct directory_traversal_callback_ctx *ctx = user_data;
+    JNIEnv *env = ctx->env;
```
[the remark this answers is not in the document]
cmello: Added a comment, thanks! That callback is synchronous, that's why `aws_jni_get_thread_env()` is not necessary

src/native/directory_traversal.c
cmello: Changed as suggested, thanks!

src/native/directory_traversal.c
cmello: If an exception is thrown by the user callback, the traversal is cancelled. Cancelling the traversal either via the user callback returning `false` or throwing, results in `DirectoryTraversal.traverse()` throwing an exception to notify user about incomplete results. This behavior is checked by the unit tests.

Added the missing `aws_jni_check_and_clear_exception()`. Curiously it was working before because the return value was `false` if an exception was thrown from inside the callback.

src/native/directory_traversal.c
cmello: Beautiful, thanks!

Looked into other usages of the helper function, need to raise the Java exception before returning.

src/main/java/software/amazon/awssdk/crt/io/DirectoryTraversal.java
cmello: Added the @deprecated annotation and corresponding documentation.

CMakeLists.txt
cmello: Addressed in PR #423

src/native/directory_traversal.c
graebm: Oh hey, this is a synchronous callback, we can actually preserve the java exception!

instead of calling `aws_jni_check_and_clear_exception(env)`, which clears the exception, just check if the exception occurred via `(*env)->ExceptionCheck(env)`

and if an exception occurred, then stop the iteration by returning false.

and then in the `Java_software_amazon_awssdk_crt_io_DirectoryTraversal_crtTraverse()`, the JNI exception will still be set. I think that when that function returns, since the JNI exception is still set, you'll see the exception propagate!

Does that make sense? typing quick before the end of the day

src/native/directory_traversal.c
```diff
@@ -0,0 +1,104 @@
...
+    struct directory_traversal_callback_ctx *ctx = user_data;
+
+    /* this callback is synchronous, therefore no need to use aws_jni_get_thread_env() */
+    JNIEnv *env = ctx->env;
+
+    jobject directory_entry_object = (*env)->NewObject(
+        env,
+        directory_entry_properties.directory_entry_class,
+        directory_entry_properties.directory_entry_constructor_method_id);
+    if (aws_jni_check_and_clear_exception(env) || directory_entry_object == NULL) {
```
graebm, line 26: similarly, DON'T clear the exception, just stop iterating

src/native/directory_traversal.c
cmello: Good idea, thanks! :-)

src/native/directory_traversal.c
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    struct directory_traversal_callback_ctx ctx = {
+        .env = env,
+        .handler = handler,
+    };
+
+    struct aws_allocator *allocator = aws_jni_get_allocator();
+    if (aws_directory_traverse(allocator, path_str, (bool)recursive, s_on_directory_entry, &ctx)) {
+        aws_jni_throw_runtime_exception(env, "Directory traversal failed");
```
graebm, line 100: ugh. I just realized aws_directory_traverse() raises a C error if the user returns `false` to stop iteration. So, with the code as-is. If a Java exception is raised, that leads to `s_on_directory_entry` returning `false`, and `aws_directory_traverse()` raises a C error, and then we throw a new java exception here. I assume the new exception erases the existence of the one that came from the callback? The original exception was probably more useful to the user than one that just says "directory traversal failed". 

maybe do like:
```
if (aws_directory_traverse(...)) {
    /* If there's already a Java exception being thrown from the callback, then we don't need to throw another*/
    if (!(*env)->ExceptionCheck(env)) {
        aws_jni_throw_runtime_exception(env, "Directory traversal failed");
    }
}
```

src/test/java/software/amazon/awssdk/crt/test/DirectoryTraversalTest.java
```diff
@@ -0,0 +1,250 @@
...
+    public void testTraverseDirectoryCallbackThrowingException() throws Exception {
+
+        try (final DirectoryStructureHelper directoryStructure = new DirectoryStructureHelper()) {
+            // traverse dir using CRT
+            Set<String> entries = new HashSet<>();
+
+            DirectoryTraversal.traverse(directoryStructure.getRootDirectory(), false, new DirectoryTraversalHandler() {
+                @Override
+                public boolean onDirectoryEntry(final DirectoryEntry directoryEntry) {
+                    throw new RuntimeException("THROWN BY TEST CALLBACK");
```
graebm, line 242: with the existing code, what happens to `RuntimeException("THROWN BY TEST CALLBACK")`?
is it:
1) `RuntimeException("THROWN BY TEST CALLBACK")` is raised?
2) `RuntimeException("AWS_ERROR_OPERATION_INTERUPTED")` is raised?
3) `RuntimeException("AWS_ERROR_OPERATION_INTERUPTED")` which wraps `RuntimeException("THROWN BY TEST CALLBACK")` is raised?

I know I'd personally prefer 1 or 3, so I can see the actual reason for the failure

cmello: The current behavior is 1. I just added an assertion in the unit test for this particular check. Thanks!

graebm (approved):

PR #423, opened 2022-01-03
Fix JAVA_HOME used by gradle cmakeBuild task

cmello:
gradle is changing the JAVA_HOME path to be the nested JRE path instead of the java home set by the builder.
Then the build fails because the JRE path doesn't contain the JNI include headers and libraries.

This commit creates an optional override to the JAVA_HOME used
by the cmakeBuild task.

*Issue #, if available:*

*Description of changes:*


CMakeLists.txt
```diff
@@ -196,6 +196,12 @@ if (NOT ANDROID)
         set(JAVA_AWT_LIBRARY NotNeeded)
         set(JAVA_AWT_INCLUDE_PATH NotNeeded)
     endif()
+    if (DEFINED ENV{OVERRIDE_JAVA_HOME})
+        # This override might be set by builder to avoid using the JAVA_HOME modified by gradle
+        file(TO_CMAKE_PATH "$ENV{OVERRIDE_JAVA_HOME}" OVERRIDE_JAVA_HOME)
```
graebm, line 201: I wish we knew better why this was happening, but I'm ok with this as a workaround to get us unstuck

graebm (approved):

TingDaoK (approved):
I agree that's a workaround we can deliver first and see can we remove gradle later

PR #425, opened 2022-01-07
Regen docs to remove old jquery ref

bretambrose:


sbSteveK (approved):

graebm:
if the next person to update docs has an older version of java on their machine, this old jquery ref will come back :|

I have a dream that we get some kind of github actions automation to generate these docs. We can just have to commit docs to their own branch, instead of `main` so we don't need to worry about getting PRs approved, or accidentally breaking `main`

PR #430, opened 2022-01-20
Update aws-c-s3 and add CopyObject to meta request enum

cmello:
* aws-c-s3 submodule updated to v0.1.31
* add CopyObject meta request type to enum
* added javadoc

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/s3/S3MetaRequestResponseHandler.java
+++ b/src/main/java/software/amazon/awssdk/crt/s3/S3MetaRequestResponseHandler.java
@@ -3,14 +3,42 @@ package software.amazon.awssdk.crt.s3;
 import java.nio.ByteBuffer;
 import software.amazon.awssdk.crt.http.HttpHeader;
 
+/**
+ * Interface called by native code to provide S3MetaRequest responses.
+ */
 public interface S3MetaRequestResponseHandler {
+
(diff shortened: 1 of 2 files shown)


src/main/java/software/amazon/awssdk/crt/s3/S3MetaRequestResponseHandler.java
```diff
@@ -3,14 +3,42 @@
...
     default int onResponseBody(ByteBuffer bodyBytesIn, long objectRangeStart, long objectRangeEnd) {
         return 0;
     }
 
+    /**
+     * Invoked when the entire meta request execution is complete.
+     *
+     * @param errorCode The CRT error code
+     * @param responseStatus statusCode of the HTTP response
+     * @param errorPayload body of the error response
```
TingDaoK, line 40: Maybe clarify it could be null, if the entire meta request finishes without error

TingDaoK (approved):
fix & ship.

src/main/java/software/amazon/awssdk/crt/s3/S3MetaRequestResponseHandler.java
cmello: Fixed, thanks!

Issue #432, opened 2022-01-20
CRT TLS context


aceeric:
Problem: The current CRT class `S3Client` does not support HTTP access to an S3 bucket. The code always establishes a TLS context - either by using the one provided to it, or by creating a default one. This then later results in the selection of HTTPS over 443. We want to use the AWS SDK + CRT in a Kubernetes service-meshed environment in which the endpoints are HTTP, and the mesh handles TLS.

Specifically:

In the CRT code, the code in `src/native/s3_client.c` in the function `JNIEXPORT jlong JNICALL Java_software_amazon_awssdk_crt_s3_S3Client_s3ClientNew` has code:
```
    ...
    struct aws_s3_client_config client_config = {
        .max_active_connections_override = max_connections,
        .region = region,
        .client_bootstrap = client_bootstrap,
        .tls_connection_options = tls_options,
        .signing_config = &signing_config,
        .part_size = (size_t)part_size,
        .throughput_target_gbps = throughput_target_gbps,
        .retry_strategy = retry_strategy,
        .shutdown_callback = s_on_s3_client_shutdown_complete_callback,
        .shutdown_callback_user_data = callback_data,
        .compute_content_md5 = compute_content_md5 ? AWS_MR_CONTENT_MD5_ENABLED : AWS_MR_CONTENT_MD5_DISABLED,
    };
    ....
```
In the `aws_s3_client_config` struct there is a member `enum aws_s3_meta_request_tls_mode tls_mode;` which is *not initialized* in the above block. Therefore, by default, this struct member has value zero which maps to enum `AWS_MR_TLS_ENABLED`. This enum is referenced in `source/s3_client.c` and has the effect of *always* enabling TLS.

In order to change this to support HTTP or HTTPS, I propose to submit a PR as follows:

1) Add a parameter `boolean tlsEnabled` to the JNI method `s3ClientNew` signature
2) Inside the JNI C code for that function, initialize the aforementioned struct member to enable or disable TLS on the basis of the parameter value. The code in `source/s3_client.c` already anticipates this so - no changes are required there
3) Modify `S3ClientOptions` and `S3Client` to also accommodate `tlsEnabled`

So this  change is localized in the CRT Java/JNI code, and doesn't touch any of the git sub-modules. We've already proofed this in-house. Does the idea of this PR seem acceptable? If so I will submit for review. Thanks.

JonathanHenson:
[...]

Issue #431, opened 2022-01-20
Custom ports in the S3Client


aceeric:
This relates to the AWS Java SDK V2 pull request: https://github.com/aws/aws-sdk-java-v2/pull/2976

The PR relates to using the SDK `TransferManager` to parallelize downloads of large objects. In our testing, this parallelization results in large objects downloading up to 5X faster than non-parallel downloads. The pull request refers to using the CRT `S3Client` class. In looking at CRT code, the `S3Client` class makes its way down into the JNI layer and ultimately winds up resolving a connection to either 80 or 443 (hard-coded.)

We would like to use the SDK `TransferManager` functionality in a Kubernetes environment with an S3-compliant object store (e.g. Minio) and would like the flexibility to use different ports both inside the cluster and also outside the cluster to support - for example - desktop testing via port-forwarding.

To support this, the CRT `S3Client` and the JNI `s3ClientNew` method (and related lower-level C functions) would need to be modified to allow a custom port, much like the custom endpoint that is currently accepted. This would replace the 80/443 hardcodes.

We've already done some experimentation with this and the change does  not seem too pervasive if localized to the S3Client and below. Before investing the effort in a PR - I wanted to get your views on whether you would accept such a PR - or - if you're already working on changes to support custom ports in the CRT in the future. Thank you.

JonathanHenson:
It SHOULD simply be pulled from the endpoint you pass (e.g. http://dnsname:port). If the port is not honored, it's a bug and we need to fix it.

ron1:
@JonathanHenson Note that in the following test (here)[https://github.com/awslabs/aws-crt-java/blob/7adb0e6ed09fe8fbd28f179e4d1c7bb412da5ddf/src/test/java/software/amazon/awssdk/crt/test/S3ClientTest.java#L37], ENDPOINT is the dnsname of a host. Based on your comment, should it instead be https://dnsname:443 where the :443 suffix is optional? I believe changing the test to include the https protocol in the endpoint results in a dns host resolver error. Am I correct such an error should be considered a bug?

aceeric:
@JonathanHenson As I read the code:

1. In `aws-c-s3`, file `source/s3_client.c` has function `aws_s3_client_make_meta_request` which is called by the AWS SDK V2 Java code as a result of a `getObject` SDK call
[...]

PR #435, opened 2022-01-30
Release v0.15.15

aceeric:
This PR addresses the following two issues:

1. https://github.com/awslabs/aws-crt-java/issues/431
2. https://github.com/awslabs/aws-crt-java/issues/432

This PR supports the ability to perform S3 get and put operations over HTTP as well as HTTPS, *and* with custom ports. This PR depends on changes made to the aws-c-s3 repo. Those changes are at my fork here: https://github.com/aceeric/aws-c-s3/tree/release-v0.1.30

I wasn't sure the best way to submit this PR with the submodule dependency so I'm starting this way for now.

The approach in this PR adds a URI into the `S3MetaRequestOptions` class. This field is evaluated by the `S3Client.makeMetaRequest` method, and propagated to the JNI and C layers below where it influences how the connection is created. I left the default behavior for the `S3Client` in which a TLS context is always created for the client - either explicitly by the caller - or implicitly by the CRT. In the meta request though - if the URI is "http" then that TLS Context is simply excluded in the call to `aws_s3_client_make_meta_request`.

I considered two alternate approaches, but elected to submit this one. The alternate approaches were:

1) Accept a port and scheme in the `S3MetaRequestOptions` class. Reason: The URI field is somewhat redundant with the `HttpRequest` host header. However, splitting out the scheme and port individually makes the `S3MetaRequestOptions` class a little busier
2) Modify the CRT `HttpRequest` to look more like the AWS SDK `SdkHttpRequest` class, which accepts protocol, host, and port. But that change would have a huge footprint on the CRT

In the end - I went with this approach as a starter. I also modified the `S3ClientTest` class to accept region and endpoint as env vars to support my testing. I elected to make these changes to the latest tagged release. So I realize I'm a little behind. If you accept this PR I can rebase on the latest of this and the submodule and re-submit.

### DIFF
--- a/src/native/s3_client.c
+++ b/src/native/s3_client.c
@@ -50,7 +50,6 @@ JNIEXPORT jlong JNICALL Java_software_amazon_awssdk_crt_s3_S3Client_s3ClientNew(
     jbyteArray jni_endpoint,
     jlong jni_client_bootstrap,
     jlong jni_tls_ctx,
-    jboolean use_tls,
     jlong jni_credentials_provider,
     jlong part_size,
     jdouble throughput_target_gbps,
@@ -126,8 +125,6 @@ JNIEXPORT jlong JNICALL Java_software_amazon_awssdk_crt_s3_S3Client_s3ClientNew(
         .region = region,
         .client_bootstrap = client_bootstrap,
         .tls_connection_options = tls_options,
-        // TODO WHETHER TLS IS INITAILIZED OR NOT DOESNT FORCE/PREVENT TLS LATER SO REMOVE?
-        .tls_mode = use_tls ? AWS_MR_TLS_ENABLED : AWS_MR_TLS_DISABLED,
         .signing_config = &signing_config,
         .part_size = (size_t)part_size,
         .throughput_target_gbps = throughput_target_gbps,
@@ -394,7 +391,7 @@ JNIEXPORT jlong JNICALL Java_software_amazon_awssdk_crt_s3_S3Client_s3ClientMake
         .body_callback = s_on_s3_meta_request_body_callback,
         .finish_callback = s_on_s3_meta_request_finish_callback,
         .shutdown_callback = s_on_s3_meta_request_shutdown_complete_callback,
-        .use_tls = use_tls ? true : false,
+        .tls_mode = use_tls ? AWS_MR_TLS_ENABLED : AWS_MR_TLS_DISABLED,
         .port = port};
 
     struct aws_s3_meta_request *meta_request = aws_s3_client_make_meta_request(client, &meta_request_options);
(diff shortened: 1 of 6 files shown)


TingDaoK:
Do you also update the submodule [aws-c-s3](https://github.com/awslabs/aws-c-s3)? Can you open a PR to that as well?

aceeric:
@TingDaoK Hello - per your request please see PR: https://github.com/awslabs/aws-c-s3/pull/174

aceeric:
@TingDaoK Hello - I updated this PR and the corresponding one here: https://github.com/awslabs/aws-c-s3/pull/174 to pass the URI into the C code.

src/native/s3_client.c
```diff
@@ -2,6 +2,7 @@
  * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
  * SPDX-License-Identifier: Apache-2.0.
  */
+#include "aws/io/uri.h"
```
TingDaoK, line 5: can you use the same pattern to include the dependency? So, something like `#include <aws/io/uri.h>`

src/native/s3_client.c
```diff
@@ -404,6 +406,14 @@ JNIEXPORT jlong JNICALL Java_software_amazon_awssdk_crt_s3_S3Client_s3ClientMake
         aws_s3_init_default_signing_config(signing_config, region, credentials_provider);
     }
 
+    struct aws_uri endpoint;
+    AWS_ZERO_STRUCT(endpoint);
+    if (jni_endpoint != NULL) {
+        struct aws_byte_cursor endpoint_str = aws_jni_byte_cursor_from_jbyteArray_acquire(env, jni_endpoint);
+        AWS_FATAL_ASSERT(aws_uri_init_parse(&endpoint, allocator, &endpoint_str) == AWS_OP_SUCCESS);
```
TingDaoK, line 413: Well, I think we can handle this error. That should not crash the program. 
```suggestion
        if (aws_uri_init_parse(&endpoint, allocator, &endpoint_str)) {
            aws_jni_throw_runtime_exception(env, "S3Client.aws_s3_client_make_meta_request: failed to parse endpoint");
            aws_jni_byte_cursor_from_jbyteArray_release(env, jni_endpoint, endpoint_str);
            goto done;
        }
```

Also, for proper clean up, you need to move the endpoint part after `request_message` created, as we clean up the resources from `done` tag

src/native/s3_client.c
aceeric: ok done

src/native/s3_client.c
aceeric: done - handled the error  very slightly differently to call `aws_jni_byte_cursor_from_jbyteArray_release` only once

TingDaoK (approved):
looks good! Thank you!

aceeric:
I noticed the Windows build was failing. It was complaining about ref to `meta_request`. I build on Linux no issues. Even so I moved the declaration to the top of the function so I don't see a way the Windows compiler can complain about it now

aceeric:
@TingDaoK Hello - Thank you for your help - is there anything else needed from me to get this into the main branch?

TingDaoK:
Nope, thank you. I merged for you, let me know if you need a new release to use it from maven.

aceeric:
Super - thank you. If you can release to Maven we would appreciate it because we have some in-house code that presently uses a patched JAR and it would be better to use the official Maven release version. Thanks.

PR #436, opened 2022-01-31
Stream manager

TingDaoK:
- Java binding of Http2Stream manager


dagnir:
How does ALPN factor into this design if the customer wants to let the SDK decide at runtime?

TingDaoK:
> How does ALPN factor into this design if the customer wants to let the SDK decide at runtime?

Can you clarify a bit more? I am not sure I understand what SDK to decide here. So, this stream manager only support HTTP/2. You can set ALPN from tls settings passing in when the stream manager created.

src/main/java/software/amazon/awssdk/crt/http/Http2StreamManager.java
```diff
@@ -0,0 +1,274 @@
...
+    /**
+     * Request a Http2Stream from StreamManager.
+     *
+     * @param request
+     * @param streamHandler
+     * @return A future for a Http2Stream that will be completed when the stream is
+     *         acquired.
+     * @throws CrtRuntimeException
+     */
+    public CompletableFuture<Http2Stream> acquireStream(Http2Request request,
```
dagnir, line 180: Can we get an overload that takes a normal HttpRequest?

src/main/java/software/amazon/awssdk/crt/http/Http2StreamManager.java
```diff
@@ -0,0 +1,274 @@
...
+
+        int maxConnections = options.getMaxConnections();
+        if (maxConnections <= 0) {
+            throw new IllegalArgumentException("Max Connections must be greater than zero.");
+        }
+        int maxConcurrentStreamsPerConnection = options.getMaxConcurrentStreamsPerConnection();
+        if (maxConcurrentStreamsPerConnection <= 0) {
+            throw new IllegalArgumentException("Max Concurrent Streams Per Connection must be greater than zero.");
+        }
+        int idealConcurrentStreamsPerConnection = options.getIdealConcurrentStreamsPerConnection();
```
dagnir, line 84: What does this do?

src/native/http2_stream_manager.c
```diff
@@ -0,0 +1,393 @@
...
+    int use_tls = (jni_tls_ctx != 0);
+
+    struct aws_tls_connection_options tls_conn_options;
+    AWS_ZERO_STRUCT(tls_conn_options);
+
+    if (use_tls) {
+        aws_tls_connection_options_init_from_ctx(&tls_conn_options, tls_ctx);
+        aws_tls_connection_options_set_server_name(&tls_conn_options, allocator, &endpoint);
+        /* Make sure we use h2 alpn list to create HTTP/2 connection. */
+        aws_tls_connection_options_set_alpn_list(&tls_conn_options, allocator, "h2");
```
TingDaoK, line 151: Should we set the ALPN or should we just rely on the user?

TingDaoK:
> > Can you clarify a bit more? I am not sure I understand what SDK to decide here. So, this stream manager only support HTTP/2. You can set ALPN from tls settings passing in when the stream manager created.
> 
> If the user creates an SDK client and just wants the SDK to offer both versions (ALPN list = [http/1.1, h2]) and let the service to decide the protocol for them. `HttpClientConnectionManager` is used for HTTP/1.1 and `Http2StreamManager` HTTP/2. It seems like the SDK can't support that use case because the manager to use depends on the protocol.

For now, if you use [http/1.1, h2] and stream manager got HTTP/1.1 connection, the stream acquiring will fail with exception that the expected version of connection cannot be acquired. Maybe SDK can catch that exception and use HTTP/1.1 client instead? I previously think HTTP/1.1 doesn't need stream manager, as it doesn't really have stream. But, I can add the support for HTTP/1.1 connection to our stream manager, which will basically just a wrap around the connection manager if needed.

TingDaoK:
Decided to add another level from Java (CRT Java probably) to adapt both HTTP/1.1 and HTTP/2.
Which will be:
- Have a connection manager under the hood, and acquire a connection
- Check the connection version.
- If HTTP/1.1, make request on the connection, not release the connection until the request done
- If HTTP/2, create HTTP2 stream manager and release the connection manager under the hood, use the API from stream manager.

src/main/java/software/amazon/awssdk/crt/http/Http2StreamManager.java
TingDaoK: Added documents to `HttpStreamManagerOptions`

src/test/java/software/amazon/awssdk/crt/test/Http2StreamManagerTest.java
```diff
@@ -0,0 +1,201 @@
...
+            throws Exception {
+        URI uri = new URI(endpoint);
+        HttpHeader[] requestHeaders = new HttpHeader[] {
+                new HttpHeader(":method", method),
+                new HttpHeader(":path", path),
+                new HttpHeader(":scheme", uri.getScheme()),
+                new HttpHeader(":authority", uri.getHost()),
+                new HttpHeader("content-length", Integer.toString(requestBody.getBytes(UTF8).length))
+        };
+        Http2Request request = new Http2Request(requestHeaders, null);
```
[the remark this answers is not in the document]
TingDaoK: The body is empty...

TingDaoK:
The uniformed stream manager is created in a following PR [here](https://github.com/awslabs/aws-crt-java/pull/444)

dagnir (approved):
:shipit:

src/main/java/software/amazon/awssdk/crt/http/Http2StreamManager.java
```diff
@@ -0,0 +1,282 @@
...
+    private final int maxConcurrentStreamsPerConnection;
+    private final CompletableFuture<Void> shutdownComplete = new CompletableFuture<>();
+
+    /**
+     * Factory function for Http2StreamManager instances
+     *
+     * @param options configuration options
+     * @return a new instance of an Http2StreamManager
+     */
+    public static Http2StreamManager create(HttpStreamManagerOptions options) {
```
graebm, line 39: not a fan of the HTTP/2 manager taking generic "HTTP" options.
every time we try and make a generic all purpose thing we regret it
let's have the HTTP/2 manager take HTTP/2 options
and an HTTP/1 manager can take HTTP/1 options
or there can be some options base class for common params
or if there's no HTTP/1-specific options then HTTP/1 can use the base class and HTTP/2 can use the HTTP/2-specific class

src/main/java/software/amazon/awssdk/crt/http/Http2StreamManager.java
```diff
@@ -0,0 +1,282 @@
...
+                    "Ideal Concurrent Streams Per Connection must be greater than zero and smaller than max.");
+        }
+
+        int port = options.getPort();
+        if (port == -1) {
+            port = uri.getPort();
+            /* Pick a default port based on the scheme if one wasn't set */
+            if (port == -1) {
+                if (HTTP.equals(uri.getScheme())) {
+                    port = DEFAULT_HTTP_PORT;
```
graebm, line 96: we don't properly do plaintext HTTP/2, we only do it with "prior knowledge", right?

is this error caught further down the pipe in C? will the error be clearly raised to the user?

graebm (commented):
I'll look more tomorrow

src/native/aws_signing.c
```diff
@@ -74,10 +74,6 @@ static void s_cleanup_callback_data(struct s_aws_sign_request_callback_data *cal
     }
 
     if (callback_data->native_request) {
-        struct aws_input_stream *input_stream = aws_http_message_get_body_stream(callback_data->native_request);
-        if (input_stream != NULL) {
-            aws_input_stream_destroy(input_stream);
-        }
```
graebm, line 80: get this bugfix into main ASAP

src/native/http2_stream_manager.c
```diff
@@ -0,0 +1,403 @@
...
+};
+
+static void s_destroy_manager_binding(struct aws_http2_stream_manager_binding *binding) {
+    if (binding == NULL) {
+        return;
+    }
+
+    JNIEnv *env = aws_jni_get_thread_env(binding->jvm);
+    if (binding->java_http2_stream_manager != NULL) {
+        (*env)->DeleteWeakGlobalRef(env, binding->java_http2_stream_manager);
```
graebm, line 57: Bret overhauled how we get env, check out usage of: aws_jni_acquire_thread_env()

oh. in that overhaul we stopped helper functions from grabbing `env` when we could have passed in `env` as a function param. You should probably just pass `env` in as a function param

src/native/http2_stream_manager.c
```diff
@@ -0,0 +1,403 @@
...
+    }
+
+    aws_mem_release(aws_jni_get_allocator(), binding);
+}
+
+static void s_on_http_conn_manager_shutdown_complete_callback(void *user_data) {
+
+    struct aws_http2_stream_manager_binding *binding = (struct aws_http2_stream_manager_binding *)user_data;
+    JNIEnv *env = aws_jni_get_thread_env(binding->jvm);
```
graebm, line 66: again, use aws_jni_acquire_thread_env() instead, and check the results

ohhh. maybe you haven't merged from main in a while

src/native/http2_stream_manager.c
```diff
@@ -0,0 +1,403 @@
...
+
+    AWS_LOGF_DEBUG(AWS_LS_HTTP_STREAM_MANAGER, "Java Stream Manager Shutdown Complete");
+    jobject java_http2_stream_manager = (*env)->NewLocalRef(env, binding->java_http2_stream_manager);
+    if (java_http2_stream_manager != NULL) {
+        (*env)->CallVoidMethod(env, java_http2_stream_manager, http2_stream_manager_properties.onShutdownComplete);
+
+        AWS_FATAL_ASSERT(!aws_jni_check_and_clear_exception(env));
```
graebm, line 73: why fatal assert? why not just clear the exception and move along

src/native/http2_stream_manager.c
```diff
@@ -0,0 +1,403 @@
...
+    (void)jni_class;
+
+    struct aws_client_bootstrap *client_bootstrap = (struct aws_client_bootstrap *)jni_client_bootstrap;
+    struct aws_socket_options *socket_options = (struct aws_socket_options *)jni_socket_options;
+    struct aws_tls_ctx *tls_ctx = (struct aws_tls_ctx *)jni_tls_ctx;
+    struct aws_http2_stream_manager_binding *binding = NULL;
+    struct aws_allocator *allocator = aws_jni_get_allocator();
+
+    if (!client_bootstrap) {
+        aws_jni_throw_runtime_exception(env, "ClientBootstrap can't be null");
```
graebm, line 114: aws_jni_throw_illegal_argument_exception() seems more appropriate

src/native/http2_stream_manager.c
```diff
@@ -0,0 +1,403 @@
...
+        return (jlong)NULL;
+    }
+
+    if (!socket_options) {
+        aws_jni_throw_runtime_exception(env, "SocketOptions can't be null");
```
graebm, line 119: aws_jni_throw_illegal_argument_exception() seems more appropriate

src/native/http2_stream_manager.c
```diff
@@ -0,0 +1,403 @@
...
+    size_t num_initial_settings = marshalled_len / 2;
+    struct aws_http2_setting *initial_settings =
+        num_initial_settings ? aws_mem_calloc(allocator, num_initial_settings, sizeof(struct aws_http2_setting)) : NULL;
+
+    jlong *marshalled_settings = (*env)->GetLongArrayElements(env, java_marshalled_settings, NULL);
+    for (size_t i = 0; i < num_initial_settings; i++) {
+        jlong id = marshalled_settings[i * 2];
+        initial_settings[i].id = id;
+        jlong value = marshalled_settings[i * 2 + 1];
+        initial_settings[i].value = (uint32_t)value;
```
graebm, line 135: trivial: you should probably do `(uint32_t)` cast for id, to satisfy all compilers of the world

and add a comment that we already checked in java that these fit in uint32_t

src/native/http2_stream_manager.c
```diff
@@ -0,0 +1,403 @@
...
+    }
+
+    struct aws_byte_cursor endpoint = aws_jni_byte_cursor_from_jbyteArray_acquire(env, jni_endpoint);
+
+    if (jni_port <= 0 || 65535 < jni_port) {
+        aws_jni_throw_runtime_exception(env, "Port must be between 1 and 65535");
```
graebm, line 141: aws_jni_throw_illegal_argument_exception() probably better

src/native/http2_stream_manager.c
```diff
@@ -0,0 +1,403 @@
...
+        goto cleanup;
+    }
+
+    if (jni_max_conns <= 0) {
+        aws_jni_throw_runtime_exception(env, "Max Connections must be > 0");
```
graebm, line 146: aws_jni_throw_illegal_argument_exception() probably better

or we could validate a lot of this stuff in some common package-private java function before passing it into C

src/native/http2_stream_manager.c
```diff
@@ -0,0 +1,403 @@
...
+    manager_options.max_concurrent_streams_per_connection = (size_t)jni_max_concurrent_streams_per_connection;
+
+    if (use_tls) {
+        manager_options.tls_connection_options = &tls_conn_options;
+    }
+
+    struct aws_http_connection_monitoring_options monitoring_options;
+    AWS_ZERO_STRUCT(monitoring_options);
+    if (jni_monitoring_throughput_threshold_in_bytes_per_second >= 0 &&
+        jni_monitoring_failure_interval_in_seconds >= 2) {
```
graebm, line 198: why 2?

src/native/http2_stream_manager.c
```diff
@@ -0,0 +1,403 @@
...
+        (*env)->CallVoidMethod(
+            env, callback_data->java_async_callback, async_callback_properties.on_failure, crt_exception);
+        (*env)->DeleteLocalRef(env, crt_exception);
+        aws_http_stream_binding_destroy(env, callback_data->stream_binding);
+    } else {
+        callback_data->stream_binding->native_stream = stream;
+        jobject j_http_stream =
+            aws_java_http_stream_from_native_new(env, callback_data->stream_binding, AWS_HTTP_VERSION_2);
+        if (!j_http_stream) {
+            jobject crt_exception = aws_jni_new_crt_exception_from_error_code(env, aws_last_error());
```
graebm, line 301: if `aws_java_http_stream_from_native_new()` fails it doesn't set an aws-error-code
it raises a java exception

src/native/http2_stream_manager.c
```diff
@@ -0,0 +1,403 @@
...
+            callback_data->stream_binding->java_http_stream_base = (*env)->NewGlobalRef(env, j_http_stream);
+            (*env)->CallVoidMethod(
+                env,
+                callback_data->java_async_callback,
+                async_callback_properties.on_success_with_object,
+                callback_data->stream_binding->java_http_stream_base);
+            (*env)->DeleteLocalRef(env, j_http_stream);
+        }
+    }
+    AWS_FATAL_ASSERT(!aws_jni_check_and_clear_exception(env));
```
graebm, line 318: why FATAL_ASSERT?

src/native/http2_stream_manager.c
```diff
@@ -0,0 +1,403 @@
...
+    jbyteArray marshalled_request,
+    jobject jni_http_request_body_stream,
+    jobject jni_http_response_callback_handler,
+    jobject java_async_callback) {
+    (void)jni_class;
+    struct aws_http2_stream_manager_binding *sm_binding = (struct aws_http2_stream_manager_binding *)jni_stream_manager;
+    struct aws_http2_stream_manager *stream_manager = sm_binding->stream_manager;
+
+    if (!stream_manager) {
+        aws_jni_throw_runtime_exception(env, "Stream Manager can't be null");
```
graebm, line 335: aws_jni_throw_illegal_argument_exception()

src/native/aws_signing.c
TingDaoK: it's not a bug yet. We haven't release the ownership from user. https://github.com/awslabs/aws-crt-java/pull/436/files#diff-fd13a85f62a17285ea21b70c9e3c83e367b6cc4034d28274ba2ef4d9ea6f4ed0R390

src/native/http2_stream_manager.c
TingDaoK: updated to get the exception using `ExceptionOccurred`

src/native/crt.h
```diff
@@ -55,6 +55,15 @@ void aws_jni_throw_illegal_argument_exception(JNIEnv *env, const char *msg, ...)
  ******************************************************************************/
 bool aws_jni_check_and_clear_exception(JNIEnv *env);
 
+/*******************************************************************************
+ * Get the exception that is pending on the stack and clears it if any.
+ * If an exception was pending, it is cleared and returned. Otherwise, return NULL.
+ *
+ * @return true the exception was pending, NULL otherwise. If it returns the exception
```
graebm, line 62: trivial: ~true the exception was pending~ -> the exception that was pending

src/native/crt.c
```diff
@@ -265,6 +265,15 @@ bool aws_jni_check_and_clear_exception(JNIEnv *env) {
     return exception_pending;
 }
 
+jthrowable aws_jni_get_and_clear_exception(JNIEnv *env) {
+    jthrowable exception = (*env)->ExceptionOccurred(env);
+    if (exception) {
+        (*env)->ExceptionDescribe(env);
```
graebm, line 271: Probably don't want to call `ExceptionDescribe()` which prints things to `stderr`. It made sense for `aws_jni_check_and_clear_exception()` to do this, since we're throwing out the exception there, and there's no other way to see that anything weird happened.

But presumably with this new function we're going to actually DO something with the exception

src/native/crt.c
```diff
@@ -265,6 +265,15 @@ bool aws_jni_check_and_clear_exception(JNIEnv *env) {
...
+        (*env)->ExceptionClear(env);
+    }
+    return exception;
```
graebm, line 274: this is probably going to lead to user-after-free

I assume ExceptionClear() will lead to the exception being garbage-collected, but we're returning its pointer.

This code worked in `aws_jni_check_and_clear_exception()` because we were just returning a `bool` of whether or not the pointer was non-null

src/main/java/software/amazon/awssdk/crt/http/Http2StreamManagerOptions.java
```diff
@@ -0,0 +1,159 @@
...
+     *
+     * Note:
+     *    1. the windowSize of connection manager will be ignored, as the initial flow-control window size for HTTP/2 stream
+     *       is controlled by the initial settings.
+     *    2. The expectedHttpVersion will also be ignored.
+     *
+     * @param connectionManagerOptions The connection manager options for the underlying connection manager
+     * @return this
+     */
+    public Http2StreamManagerOptions withConnectionManagerOptions(HttpClientConnectionManagerOptions connectionManagerOptions) {
```
graebm, line 140: Should we denote that this is required? (how do we usually indicate that something is required in an options class?)

src/main/java/software/amazon/awssdk/crt/http/Http2StreamManagerOptions.java
```diff
@@ -0,0 +1,159 @@
...
+        this.connectionManagerOptions = connectionManagerOptions;
+        return this;
+    }
+
+    /**
+     * Validate the stream manager options are valid to use. Throw exceptions if not.
+     */
+    public void validateOptions() {
+        connectionManagerOptions.validateOptions();
```
graebm, line 149: probably want to explicitly check this for null?

the way this code is now, users will just get a NullPointerException, but it's probably nicer to get an exception saying that connection manager is required

graebm (approved):
fix & ship

PR #437, opened 2022-02-01
fix conn pool leak when acquire future is cancelled

aajtodd:
*Issue #, if available:*
see https://github.com/awslabs/aws-sdk-kotlin/issues/511

*Description of changes:*
Check if the connection acquire future is actually completed by the callback. If not (likely due to cancellation) then return it immediately to the pool otherwise it will leak and future acquisitions will fail/hang.


src/test/java/software/amazon/awssdk/crt/test/HttpClientConnectionManagerTest.java
```diff
@@ -5,11 +5,7 @@
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.concurrent.Callable;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
+import java.util.concurrent.*;
```
bretambrose, line 8: Very minor and probably a result of IDE settings, but I think we have a preference for keeping imports explicit and not wild-carded.

bretambrose (approved):

src/test/java/software/amazon/awssdk/crt/test/HttpClientConnectionManagerTest.java
aajtodd: yup IDE auto import settings, will fix

PR #438, opened 2022-02-01
windows CI fix

TingDaoK:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ env:
 
 jobs:
   linux-compat:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


bretambrose (commented):
Primary issue is that windows CD builds for the CRT (which use 2015) will have no validation before we actually make a release.

TingDaoK:
> Primary issue is that windows CD builds for the CRT (which use 2015) will have no validation before we actually make a release.

hmmm, I am not removing the windows vc14 check. I just use the 2019, which still has vs14 available to run the CI. So, I think we still have 2015 covered by CI before CD, right?

graebm (approved):

PR #442, opened 2022-02-08
android check fix

TingDaoK:
fix the issue when we run unit test in Android app from Desktop, we will still think we run it from device.

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/CRT.java
+++ b/src/main/java/software/amazon/awssdk/crt/CRT.java
@@ -67,18 +67,11 @@ public final class CRT {
         return value.toLowerCase(Locale.US).replaceAll("[^a-z0-9]+", "");
     }
 
-    private static boolean isAndroid() {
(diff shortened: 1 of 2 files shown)


cmello (approved):

src/main/java/software/amazon/awssdk/crt/CRT.java
```diff
@@ -67,23 +67,11 @@ private static String normalize(String value) {
...
-        } catch (ClassNotFoundException ex) {
-            return false;
-        }
-    }
-
     /**
      * @return a string describing the detected platform the CRT is executing on
      * @throws UnknownPlatformException
      */
     public static String getOSIdentifier() throws UnknownPlatformException {
```
aajtodd, line 74: this is still being used on L105 to explicitly check for `android`

aajtodd (commented):
I think this is probably fine to remove since the current logic will always fail. Your first approach was also probably fine though.

I'll leave it up to CRT team which one you want to move forward with.

PR #443, opened 2022-02-11
Connection builder simplification

TwistedTwigleg:
*Description of changes:*

Make a few small changes to reduce the boiler plate needed for creating an MQTT connection using the MQTT builder without breaking backwards compatibility. Now it is possible to make a connection without needing to manually create a HostResolver, EventLoopGroup, or ClientBootstrap, this can all be done automatically by the builder (and is automatically freed when done).

This change makes the necessary CRT binding related changes to allow for simplification. These changes are primarily adding new constructors that, when used, will use static default presets of classes for ease of use and simplicity for the customer.

Changes:
* Added new constructors to `X509CredentialsProvider`, `ClientBootstrap`, and `MqttClient`.
* Adds a static default ClientBootstrap
  * Similar in use and API to the static default EventLoopGroup and HostResolver
  * Functions for getting and releasing added
* Added code to the MQTT connection class and configuration to detect if static classes are used. If static classes are used, the class will automatically free them
  * This combined with the Java V2 SDK changes, makes much easier to get a connection without needing to manually manage the EventLoopGroup, HostResolver, or ClientBootstrap.

__________

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.*


src/main/java/software/amazon/awssdk/crt/auth/credentials/X509CredentialsProvider.java
```diff
@@ -48,6 +48,21 @@ public X509CredentialsProviderBuilder withClientBootstrap(ClientBootstrap client
             return this;
         }
 
+        /**
```
bretambrose, line 51: I think it would be better to remove this and have the null bootstrap no longer be an error in the provider constructor and instead use the default there.

src/main/java/software/amazon/awssdk/crt/io/ClientBootstrap.java
```diff
@@ -70,6 +81,36 @@ private void onShutdownComplete() {
 
     public CompletableFuture<Void> getShutdownCompleteFuture() { return shutdownComplete; }
 
+    /**
+     * Releases the static ClientBootstrap, if it exists.  Primarily intended for tests that use the static
+     * default ClientBootstrap, before they call waitForNoResources().
+     */
+    public static void releaseStaticDefault() {
+        synchronized (EventLoopGroup.class) {
```
bretambrose, line 89: Wrong class here

src/main/java/software/amazon/awssdk/crt/io/ClientBootstrap.java
```diff
@@ -17,6 +17,17 @@
 
     private final CompletableFuture<Void> shutdownComplete = new CompletableFuture<>();
 
+    /**
+     * Creates a new ClientBootstrap with a default HostResolver and EventLoopGroup.
+     * Most applications will only ever need one instance of this.
+     */
+    public ClientBootstrap() throws CrtRuntimeException {
```
bretambrose, line 24: I'm not sure this is a useful public API.  Consider making it private since the only one who should ever instantiate is the static GetOrCreate

src/main/java/software/amazon/awssdk/crt/mqtt/MqttClientConnection.java
```diff
@@ -118,6 +121,18 @@ public MqttClientConnection(MqttConnectionConfig config) throws MqttException {
     @Override
     protected void releaseNativeHandle() {
         mqttClientConnectionDestroy(getNativeHandle());
+        
+        if (config != null) {
```
bretambrose, line 125: This shouldn't be necessary.

src/main/java/software/amazon/awssdk/crt/mqtt/MqttConnectionConfig.java
```diff
@@ -44,6 +44,11 @@
     private HttpProxyOptions proxyOptions;
     private Consumer<WebsocketHandshakeTransformArgs> websocketHandshakeTransform;
 
+    /* static defaults (for auto cleanup) */
+    private boolean cleanStaticDefaultClientBootstrap = false;
```
bretambrose, line 48: This shouldn't be necessary

src/main/java/software/amazon/awssdk/crt/mqtt/MqttConnectionConfig.java
```diff
@@ -512,6 +517,57 @@ public void setWebsocketHandshakeTransform(Consumer<WebsocketHandshakeTransformA
         return websocketHandshakeTransform;
     }
 
+    /**
+     * Sets whether the MQTT connection will automatically clean the resources for the static default ClientBootstrap
+     * when clearing its own resources.
+     * @param useStaticBootstrap whether to clean the static ClientBootstrap
+     */
+    public void setCleanStaticDefaultClientBootstrap(boolean useStaticBootstrap) {
```
bretambrose, line 525: Shouldn't be needed

src/main/java/software/amazon/awssdk/crt/mqtt/MqttConnectionConfig.java
```diff
@@ -539,6 +595,10 @@ public MqttConnectionConfig clone() {
             clone.setHttpProxyOptions(getHttpProxyOptions());
             clone.setWebsocketHandshakeTransform(getWebsocketHandshakeTransform());
 
+            clone.setCleanStaticDefaultClientBootstrap(getCleanStaticDefaultClientBootstrap());
```
bretambrose, line 598: Shouldn't be needed.

bretambrose (changes requested):
CrtResource.waitForNoResources should close the static bootstrap as well

TwistedTwigleg:
Thanks for the reviews! Will merge into `main`.

PR #444, opened 2022-02-21
unified stream manager

TingDaoK:
*Issue #, if available:*

*Description of changes:*


src/main/java/software/amazon/awssdk/crt/http/HttpStreamManager.java
```diff
@@ -0,0 +1,213 @@
...
+
+    private HttpClientConnectionManagerOptions getConnManagerOptFromStreamManagerOpt(HttpStreamManagerOptions options) {
+
+        HttpClientConnectionManagerOptions connManagerOptions = new HttpClientConnectionManagerOptions();
+        connManagerOptions.withClientBootstrap(options.getClientBootstrap())
+                .withSocketOptions(options.getSocketOptions())
+                .withTlsContext(options.getTlsContext())
+                .withWindowSize(options.getWindowSize())
+                .withUri(options.getUri())
+                .withMaxConnections(options.getMaxConnections())
```
TingDaoK, line 46: instead of max connections, maybe something like JAVA SDK, using max concurrency that will different between protocols. https://github.com/aws/aws-sdk-java-v2/blob/master/http-clients/aws-crt-client/src/main/java/software/amazon/awssdk/http/crt/AwsCrtAsyncHttpClient.java#L267-L272

src/main/java/software/amazon/awssdk/crt/http/HttpStreamManager.java
```diff
@@ -0,0 +1,213 @@
...
+                }
+            }
+        });
+    }
+
+    /**
+     * Get the protocol version the manager runs on.
+     */
+    public HttpVersion getHttpVersion() throws Exception {
+        if (this.h2StreamManager != null && this.connectionManager != null) {
```
TingDaoK, line 93: Let's think about the case if server have two different host that host both http/1.1 and http/2. And, they give different protocol overtime. How will we handle it? Error out? Or probably handle it someway.

src/main/java/software/amazon/awssdk/crt/http/HttpStreamManager.java
```diff
@@ -0,0 +1,213 @@
...
+     * Factory function for HttpStreamManager instances
+     *
+     * @param options configuration options
+     * @return a new instance of an HttpStreamManager
+     */
+    public static HttpStreamManager create(HttpStreamManagerOptions options) {
+        return new HttpStreamManager(options);
+    }
+
+    private HttpClientConnectionManagerOptions getConnManagerOptFromStreamManagerOpt(HttpStreamManagerOptions options) {
```
TingDaoK, line 38: Expose initial http2 settings to the stream manager.

dagnir (approved):

src/main/java/software/amazon/awssdk/crt/http/HttpStreamManager.java
TingDaoK: exposed

PR #449, opened 2022-03-03
Fix signing failure not popping up

TingDaoK:
*Issue #, if available:*

- We clear the exception raised at the end of the signing functions, which will cover the failure and causing the signing to hang forever.
- When delegate credential provider failed to get the credential, memory leak will happen.
- We need to stop the test if one of them failed to make it easier to allocate the failure test.

*Description of changes:*


sbSteveK (approved):

src/test/java/software/amazon/awssdk/crt/test/SigningTest.java
```diff
@@ -190,6 +192,50 @@ public void testSigningSuccess() throws Exception {
         }
     }
 
+
+    @Test
+    public void testSigningException() throws Exception {
```
graebm, line 197: bless you for adding tests 😇

graebm (approved):

PR #458, opened 2022-03-17
checksum callback

ilevyor:
*Issue #, if available:*

*Description of changes:*


src/test/java/software/amazon/awssdk/crt/test/S3ClientTest.java
```diff
@@ -353,6 +356,180 @@ public long getLength() {
...
+                Assert.assertEquals(Integer.valueOf(0), onGetFinishedFuture.get());
+            }
+        } catch (InterruptedException | ExecutionException ex) {
+            Assert.fail(ex.getMessage());
+        }
+    }
+
+
+    @Test
+    public void testS3GetChecksums() {
```
graebm, line 474: Do you have any tests that exercise the failure path, where validate-checksums fails?

src/main/java/software/amazon/awssdk/crt/s3/S3ResponseContext.java
```diff
@@ -0,0 +1,17 @@
...
+    public final S3ChecksumAlgorithm checksumAlgorithm;
+    public final boolean didValidateChecksum;
+
+    public S3ResponseContext(final int errorCode, final int responseStatus, final byte[] errorPayload, final S3ChecksumAlgorithm checksumAlgorithm, final boolean didValidateChecksum) {
+        this.errorCode = errorCode;
+        this.responseStatus = responseStatus;
+        this.errorPayload = errorPayload;
+        this.checksumAlgorithm = checksumAlgorithm;
+        this.didValidateChecksum = didValidateChecksum;
+    }
```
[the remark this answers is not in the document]
ilevyor: I don't think it makes sense to have a builder pattern or setters for this class, as the purpose of S3RespoonseContext is to be a parameter for the onFinished callback. So a java user really only needs to access the data provided from the CRT, no need to create a context, or set the variables.

src/main/java/software/amazon/awssdk/crt/s3/S3ResponseContext.java
```diff
@@ -0,0 +1,17 @@
+package software.amazon.awssdk.crt.s3;
+
+public class S3ResponseContext {
```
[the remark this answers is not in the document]
ilevyor: This class needs to be public, as user has this a parameter for the onFinished callback. I originally set all the fields as public final, as they should only be set by the CRT when calling the callback, updated them to private with getters as I think zoe said this is more idiomatic.

src/main/java/software/amazon/awssdk/crt/s3/S3ResponseContext.java
```diff
@@ -0,0 +1,33 @@
...
+    private final int errorCode;
+    private final int responseStatus;
+    private final byte[] errorPayload;
+    private final ChecksumAlgorithm checksumAlgorithm;
+    private final boolean didValidateChecksum;
+
+    public S3ResponseContext(final int errorCode, final int responseStatus, final byte[] errorPayload, final ChecksumAlgorithm checksumAlgorithm, final boolean didValidateChecksum) {
```
graebm, line 10: make this constructor package-private

src/main/java/software/amazon/awssdk/crt/s3/ChecksumAlgorithm.java
```diff
@@ -0,0 +1,48 @@
+package software.amazon.awssdk.crt.s3;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public enum ChecksumAlgorithm {
+
+    NONE(0),
+
+    CRC32_C(1),
```
graebm, line 10: trivial: I usually see this written as "32C" without the underscore.

src/main/java/software/amazon/awssdk/crt/s3/S3ResponseContext.java
```diff
@@ -0,0 +1,33 @@
...
+        this.errorCode = errorCode;
+        this.responseStatus = responseStatus;
+        this.errorPayload = errorPayload;
+        this.checksumAlgorithm = checksumAlgorithm;
+        this.didValidateChecksum = didValidateChecksum;
+    }
+
+    public int getErrorCode () {
+        return this.errorCode;
```
graebm, line 19: Have an instance of `CrtS3RuntimeException` here instead of giving people a raw int

src/main/java/software/amazon/awssdk/crt/s3/S3ResponseContext.java
```diff
@@ -0,0 +1,33 @@
...
+    }
+    public int getResponseStatus () {
+        return this.responseStatus;
```
graebm, line 22: It's confusing what this value will be if the operation fails (say, due to connection failures where no actual responses ever came back)

Will it be -1? 0?
Is this something the user should only query when the exception is non-null?

src/main/java/software/amazon/awssdk/crt/s3/S3ResponseContext.java
```diff
@@ -0,0 +1,33 @@
...
+    }
+    public byte[] getErrorPayload () {
+        return this.errorPayload;
```
graebm, line 25: we can get rid of errorPayload as its own member if we use CrtS3RuntimeException

src/main/java/software/amazon/awssdk/crt/s3/S3ResponseContext.java
```diff
@@ -0,0 +1,33 @@
...
+    }
+    public ChecksumAlgorithm getChecksumAlgorithm () {
+        return this.checksumAlgorithm;
```
graebm, line 28: similar to the feedback on getResponseStatus... It's confusing what these checksum values will be if the operation fails. Like, will checksumAlgorithm be NONE if things fail without getting any responses, but then it may be CRC32 if we fail later on?

Would the user actually care about checksum algorithm if there's a failure?

src/main/java/software/amazon/awssdk/crt/s3/ChecksumAlgorithm.java
zoewangg: I just checked the S3 model, it actually uses `CRC32C`, not sure why Java SDK codegen adds underscore 🤷🏼‍♀️  , so it might be better to remove the underscore here, but either way is fine with me.

```
public enum ChecksumAlgorithm {
    CRC32("CRC32"),

    CRC32_C("CRC32C"),

    SHA1("SHA1"),

    SHA256("SHA256")
```

src/main/java/s3NativeClient/com/amazonaws/s3/S3NativeClient.java
```diff
@@ -199,11 +199,11 @@ public int onResponseBody(ByteBuffer bodyBytesIn, long objectRangeStart, long ob
...
 
             @Override
-            public void onFinished(int errorCode, int responseStatus, byte[] errorPayload) {
+            public void onFinished(S3ResponseContext context) {
                 CrtS3RuntimeException ex = null;
                 try {
-                    if (errorCode != CRT.AWS_CRT_SUCCESS) {
-                        ex = new CrtS3RuntimeException(errorCode, responseStatus, errorPayload);
+                    if (context.errorCode != CRT.AWS_CRT_SUCCESS) {
+                        ex = new CrtS3RuntimeException(context.errorCode, context.responseStatus, context.errorPayload);
```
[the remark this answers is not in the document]
ilevyor: https://github.com/awslabs/aws-c-s3/blob/dd2dd34c7730cc0dc124fe7a873331c96dc90ba2/include/aws/s3/s3_client.h#L295-L317

src/test/java/software/amazon/awssdk/crt/test/S3ClientTest.java
ilevyor: No, and there isn't an obvious way that jumps out at me, need to think about this.

src/main/java/software/amazon/awssdk/crt/s3/S3ResponseContext.java
ilevyor: Looking through the code I believe the callback isn't called in the event a connection timeout, but not positive.

ilevyor: I ran the `s_test_s3_put_single_part_fail_object_inputstream_fail_reading` test in aws-c-s3 through the debugger, and the response status is left as 0 for this kind of failure.

src/main/java/software/amazon/awssdk/crt/s3/S3Client.java
```diff
@@ -64,11 +64,11 @@ public S3MetaRequest makeMetaRequest(S3MetaRequestOptions options) {
         if (options.getCredentialsProvider() != null) {
             credentialsProviderNativeHandle = options.getCredentialsProvider().getNativeHandle();
         }
-
         URI endpoint = options.getEndpoint();
+        int checksumAlgorithm = options.getChecksumAlgorithm() != null ? options.getChecksumAlgorithm().getNativeValue() : 0;
```
TingDaoK, line 68: do we need this? Will C part just parse null as zero if you pass null into C? (I am not sure about the answer though, just curious)

src/main/java/software/amazon/awssdk/crt/s3/S3MetaRequestResponseHandler.java
```diff
@@ -34,12 +34,12 @@ default int onResponseBody(ByteBuffer bodyBytesIn, long objectRangeStart, long o
...
     /**
      * Invoked when the entire meta request execution is complete.
-     *
-     * @param errorCode The CRT error code
-     * @param responseStatus statusCode of the HTTP response
-     * @param errorPayload body of the error response. Can be null if the request completed successfully
+     * @param context a wrapper object containing the following fields
+     * errorCode The CRT error code
+     * responseStatus statusCode of the HTTP response
+     * errorPayload body of the error response. Can be null if the request completed successfully
```
TingDaoK, line 40: remove these out dated documentation

src/main/java/s3NativeClient/com/amazonaws/s3/S3NativeClient.java
```diff
@@ -199,11 +199,11 @@ public int onResponseBody(ByteBuffer bodyBytesIn, long objectRangeStart, long ob
...
 
             @Override
-            public void onFinished(int errorCode, int responseStatus, byte[] errorPayload) {
+            public void onFinished(S3FinishedResponseContext context) {
                 CrtS3RuntimeException ex = null;
                 try {
-                    if (errorCode != CRT.AWS_CRT_SUCCESS) {
-                        ex = new CrtS3RuntimeException(errorCode, responseStatus, errorPayload);
+                    if (context.getErrorCode() != CRT.AWS_CRT_SUCCESS) {
+                        ex = new CrtS3RuntimeException(context.getErrorCode(), context.getResponseStatus(), context.getErrorPayload());
```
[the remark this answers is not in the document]
TingDaoK: Also, if it's an internal failure (Checksum mismatch or failed to read from file or something), the response status and error payload is not informative.

src/main/java/software/amazon/awssdk/crt/s3/S3Client.java
ilevyor: As written this would be a type error as s3ClientMakeMetaRequest takes an int as an argument. This is how we handle enums to jni, we extract the int that the java level. It seems other people do this differently though, not sure what the tradeoffs are though. https://community.oracle.com/tech/developers/discussion/1549037/passing-enums-as-arguments-within-a-jni-function

src/main/java/software/amazon/awssdk/crt/s3/S3MetaRequestResponseHandler.java
ilevyor: I just added this documentation, forgot to add a couple of points though.

src/main/java/software/amazon/awssdk/crt/s3/S3MetaRequestOptions.java
```diff
@@ -88,6 +90,33 @@ public MetaRequestType getMetaRequestType() {
         return metaRequestType;
     }
 
+    /*
+     * Specify the checksum algorithm to use use for put requests, if unset no checksum will be calculated.
```
zoewangg, line 94: What's the default value?

zoewangg (approved):

src/main/java/software/amazon/awssdk/crt/s3/S3FinishedResponseContext.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    /*
+     * If the request didn't receive a response due to a connection 
+     * failure or some othe issue the response status will be 0.
+     */
+    public int getResponseStatus () {
+        return this.responseStatus;
+    }
+    public byte[] getErrorPayload () {
```
TingDaoK, line 29: add some documentation here

src/main/java/s3NativeClient/com/amazonaws/s3/S3NativeClient.java
```diff
@@ -199,11 +199,11 @@ public int onResponseBody(ByteBuffer bodyBytesIn, long objectRangeStart, long ob
             }
 
             @Override
-            public void onFinished(int errorCode, int responseStatus, byte[] errorPayload) {
+            public void onFinished(S3FinishedResponseContext context) {
                 CrtS3RuntimeException ex = null;
                 try {
-                    if (errorCode != CRT.AWS_CRT_SUCCESS) {
-                        ex = new CrtS3RuntimeException(errorCode, responseStatus, errorPayload);
+                    if (context.getErrorCode() != CRT.AWS_CRT_SUCCESS) {
```
TingDaoK, line 205: Just throw out ideas: I think we can throw out `CrtRuntimeException` if it's not because the server side reject the request. And only throw out `CrtS3RuntimeException` when the error body is informative (server side rejects our request).

src/main/java/software/amazon/awssdk/crt/s3/S3MetaRequestResponseHandler.java
TingDaoK:
```
     * errorCode The CRT error code
     * responseStatus statusCode of the HTTP response
     * errorPayload body of the error response. Can be null if the request completed successfully
     * checksumAlgorithm, the algorith used to validate the Body, None if not validated
     * isChecksumValidated which is true if the response was validated.
```
Move those docs to the `S3FinishedResponseContext` instead. They are details about `S3FinishedResponseContext`.

src/main/java/software/amazon/awssdk/crt/s3/S3MetaRequestOptions.java
```diff
@@ -88,6 +90,33 @@ public MetaRequestType getMetaRequestType() {
...
+        return checksumAlgorithm;
+    }
+
+    /*
+     * validateChecksum will cause the client to compare a streamed calculation of the objects 
+     * checksum to a remotely stored checksum assigned to the object if one exists.
+     * the checksumValidated field passed in parameter of the finished callback will inform
+     * the user if validation occured. A mismatch will result in a AWS_ERROR_S3_RESPONSE_CHECKSUM_MISMATCH error
+     */
+    public S3MetaRequestOptions withValidateChecksum(boolean validateChecksum) {
```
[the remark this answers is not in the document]
ilevyor: NONE edited to clarify

src/main/java/s3NativeClient/com/amazonaws/s3/S3NativeClient.java
```diff
@@ -199,12 +200,15 @@ public int onResponseBody(ByteBuffer bodyBytesIn, long objectRangeStart, long ob
...
 
             @Override
-            public void onFinished(int errorCode, int responseStatus, byte[] errorPayload) {
-                CrtS3RuntimeException ex = null;
+            public void onFinished(S3FinishedResponseContext context) {
+                CrtRuntimeException ex = null;
                 try {
-                    if (errorCode != CRT.AWS_CRT_SUCCESS) {
-                        ex = new CrtS3RuntimeException(errorCode, responseStatus, errorPayload);
+                    if (context.getResponseStatus() > 299){
```
DmitriyMusatkin, line 206: Why 300+? 3xx are redirection codes and are generally not considered errors. not sure if we get any 3xx from s3.

src/main/java/software/amazon/awssdk/crt/s3/S3Client.java
DmitriyMusatkin: do we have an enum value for 0? this is a bit hard to read. 

Can also do something like Optional.ofNullable(getAlgo).map(Algo::getNativeValue).orElse(0)

DmitriyMusatkin (approved):

TingDaoK:
removed requirement of `aws-crt-java-sdk-java-test` for merging, as we need java sdk to pick up the api break change. We can merge this PR now.

Issue #311, opened 2021-04-24
Java 9+ Compatibility


MikeDombo:
https://github.com/awslabs/aws-crt-java/blob/652a6f853a35cd064dbe1dbff4aedee197991b37/src/native/java_class_ids.c#L204,L210

The ByteBuffer methods won't be able to be found on Java 9+ due to Java's changing the return type to be `ByteBuffer` instead of the superclass `Buffer`

You may need to update the code to look for what it looks for now, and then if that fails try to look for the other version.

bretambrose:
I'm running the samples without issue with jdk11 targeting jre11.  Is there something additional that needs to be done to surface this?  It appears that Buffer still has the method in question (and is probably the exact same thing really) so is it possible that the jni lookup is actually mapping to the Buffer method and everything still works?

MikeDombo:
I hit this issue when building with graal native image for jdk 11.

jmklix:
Closing: duplicate of this FR - https://github.com/awslabs/aws-crt-java/issues/780

PR #464, opened 2022-03-24
Native image support

timmattison:
This addresses issue #311.

I have added support for native image that has been tested with GraalVM 22.0.0.2 on Java 17.

The final commit on this PR builds a native image application to print out the operating system name from CRT. This application will build but fail to run without the added configuration. With the new configuration files added it runs as expected.

This has not been tested with every CRT feature so some testing needs to be done to see if the configuration files are complete.


maschnetwork:
@timmattison this is great - it helped me a lot getting it running. In the meantime there were some new classes added so I had to fix the jni-config. I can confirm this is working for a sample application that writes items to DynamoDB. You can find the full version with the latest CRT Client version here: https://github.com/maschnetwork/aws-lambda-graalvm-crt

[Here](https://github.com/maschnetwork/aws-lambda-graalvm-crt/tree/main/request-writer/src/main/resources/META-INF/native-image/software.amazon.awssdk/aws-crt-client) you can find the updated jni-config and the resource-config which only includes the linux client.

To be able to get the new jni-config I had to do the following things:

1. Clone this repo
2. Install dependencies (cmake) & GraalVM
3. Add an [App.class ](https://github.com/awslabs/aws-crt-java/blob/e243be1675c915e9781d4694719e93d80043decd/src/main/java/software/amazon/awssdk/crt/App.java)that you provided to be able to execute the native image command
4. Run `maven clean package -DskipTests=true` to get the uber-jar 
5. Run  `java -agentlib:native-image-agent=config-output-dir=./graal-output -cp target/aws-crt-1.0.0-SNAPSHOT.jar software.amazon.awssdk.crt.App` to get the new configs

PR #472, opened 2022-04-14
Misc eventstream-related fixes

bretambrose:
* Fix a global ref leak to ServerConnectionHandler
* Fix a bunch of potential ref leaks in error paths
* Fix potential local ref leaks when creating Java objects from an event loop thread (where they won't get automatically cleaned up on return since we're not returning to Java)
* Fix early drop of ServerListener's downstream references before it actually gets cleaned up
* Add a bunch of null connection/continuation checks to the JNI entrypoints of eventstream functions from Java
* remove an unused http helper function that wasn't cleaning up references properly


TwistedTwigleg (approved):
I took a look through the changes and it looks good to me! 👍

MikeDombo (approved):

PR #478, opened 2022-04-27
Jni env safety

bretambrose:
* Adds check-and-return to all logic that depends on JNI env retrieval
* Adds check and skip on JNI env retrieval to event loop group shutdown thread exit callback
* Fixes static client bootstrap early shutdown issue


MikeDombo (approved):

src/native/aws_signing.c
```diff
@@ -54,6 +54,9 @@ struct s_aws_sign_request_callback_data {
 static void s_cleanup_callback_data(struct s_aws_sign_request_callback_data *callback_data) {
 
     JNIEnv *env = aws_jni_get_thread_env(callback_data->jvm);
+    if (env == NULL) {
```
TingDaoK, line 57: Is that the same logic as Mike did for python? Like when JVM has gone, just silently exist? If so, can we add some comments? Like https://github.com/awslabs/aws-crt-python/blob/main/source/auth_credentials.c#L578

bretambrose: A little reluctant to copy-and-paste the same comment 60 times.

TingDaoK: Can we add the reason behind at the PR description at least?

TingDaoK (approved):
fix or not and ship :shipit:

src/native/aws_signing.c
bretambrose: I added the comments but still am not thrilled with the idea of every get-ptr-check-ptr-use-ptr requiring a comment on check ptr.

src/native/event_loop_group.c
```diff
@@ -49,17 +49,26 @@ static void s_event_loop_group_cleanup_completion_callback(void *user_data) {
     (*jvm)->AttachCurrentThread(jvm, (void **)&temp_env, NULL);
     env = temp_env;
```
TwistedTwigleg, line 50: Should we also check the result of `AttachCurrentThread` against `JNI_OK`, in case an error that is not an no-op occurs?

src/native/event_stream_rpc_client.c
```diff
@@ -81,6 +89,9 @@ static void s_on_connection_shutdown(
 
     struct connection_callback_data *callback_data = user_data;
     JNIEnv *env = aws_jni_get_thread_env(callback_data->jvm);
+    if (env == NULL) {
+        return;
```
TwistedTwigleg, line 93: Super minor: Missing comment on not crashing.

TwistedTwigleg (approved):
Looks good! 👍

src/native/event_loop_group.c
bretambrose: I don't think it gives you an environment unless it succeeds.

TwistedTwigleg: I looked and I think you are right, it looks like it just returns an error and will keep `env` as null if there is an error.
It should be good then as is.

MikeDombo (approved):

PR #479, opened 2022-04-29
Support for Apple M1

ilevyor:
*Issue #, if available:*

*Description of changes:*

Edited pom.xml to build for mac-arm chip and added corresponding build script.


pom.xml
```diff
@@ -99,6 +99,17 @@
                 <cmake.cflags/>
             </properties>
         </profile>
+        <profile>
+            <id>mac-arm-make</id>
+                <activation>
+                    <property>
+                        <name>build_mac_m1</name>
```
graebm, line 106: call it "arm64" not "m1"

$100 says there's an "m2" in the next year or so. They're running out of Pro Max Ultra adjectives.

pom.xml
```diff
@@ -99,6 +99,17 @@
...
+                    </property>
+                </activation>
```
graebm, line 108: should there be activation properties set here so that it's the default profile on an M1 mac
```
                <os>
                    <family>mac</family>
                    <arch>arm64</arch>
                </os>
```

otherwise, the existing `mac-make` profile is going to be picked by default, since it specifies `<family>mac</family>`, right?

codebuild/cd/deploy-snapshot.yml
```diff
@@ -27,6 +27,7 @@ phases:
       - cp -rv $CODEBUILD_SRC_DIR_linux_x64/dist/* $CODEBUILD_SRC_DIR/aws-crt-java/target/cmake-build/
       - cp -rv $CODEBUILD_SRC_DIR_linux_x86/dist/* $CODEBUILD_SRC_DIR/aws-crt-java/target/cmake-build/
       - cp -rv $CODEBUILD_SRC_DIR_osx_x64/* $CODEBUILD_SRC_DIR/aws-crt-java/target/cmake-build/
+      - cp -rv $CODEBUILD_SRC_DIR_osx_arm64/* $CODEBUILD_SRC_DIR/aws-crt-java/target/cmake-build/
```
graebm, line 30: trivial: let's modify the existing `osx-x64-build.sh` shell script to explicitly choose the profile that will build for x64

graebm (approved):
approved if we can run the pipeline one more time successfully
and check that our artifacts are working if you install them on M1

PR #482, opened 2022-05-06
Remove event loop group shutdown callback 

bretambrose:
* Remove event loop group shutdown callback that was doing nothing useful and likely causing rare crashes when the JVM shuts down first (the async elg cleanup thread was not attached to the JVM and thus would not keep it alive).


bretambrose:
On second thought, this was not the right thing to do.

PR #483, opened 2022-05-10
Refactor CRT JVM shutdown safety

bretambrose:
* Refactors all JNIEnv usage to be within the scope of a read-write lock based on JVM status

The problem: if CRT resources are not precisely tracked and handled by a complicated ref-counting API, native CRT threads can outlive the JVM while still performing requested work.  When the native thread attempts to callback to Java in the time period between JVM shutdown and process shutdown, the cached JVM pointer is potentially garbage (it's not just a const vtable).  Dereferencing the invalid JVM pointer would lead to crashes on shutdown.

The solution: track the state of the JVM internally using the CRT's init as the introduction and a JVM shutdown hook invocation as the end of validity.  Change the JNIEnv retrieval API to use locks to guarantee that the JVM will remain valid for the duration of the JNIEnv usage.  Use a read-write lock to minimize contention and performance degradation: only CRT init and JVM shutdown ever take the writer lock, all other uses use a reader lock.  Under normal operation, performance and concurrency should not be affected other than the cost of taking an uncontested reader lock.

The multiple-JVM use case is not yet supported, but the changes here are compatible with it should it eventually be needed.
Multi-JVM support would require:
* ref-counting CRT init/shutdown
* per-JVM class/method/member id tracking (java_class_ids.h)

### DIFF
--- a/.builder/actions/aws_crt_java_test.py
+++ b/.builder/actions/aws_crt_java_test.py
@@ -10,7 +10,6 @@ class AWSCrtJavaTest(Builder.Action):
         env.shell.setenv('AWS_CRT_MEMORY_TRACING', '2')
         actions = []
 
-        os.system("echo TESTING")
         all_test_result = os.system("mvn -P continuous-integration -B test -DredirectTestOutputToFile=true -DreuseForks=false \
             -DrerunFailingTestsCount=5 -Daws.crt.memory.tracing=2 -Daws.crt.debugnative=true")
 
--- a/.builder/actions/aws_crt_java_test.py
+++ b/.builder/actions/aws_crt_java_test.py
@@ -12,7 +12,7 @@ class AWSCrtJavaTest(Builder.Action):
 
         os.system("echo TESTING")
         if os.system("mvn -P continuous-integration -B test -DredirectTestOutputToFile=true -DreuseForks=false \
-            -Daws.crt.memory.tracing=2 -Daws.crt.debugnative=true"):
+            -DrerunFailingTestsCount=5 -Daws.crt.memory.tracing=2 -Daws.crt.debugnative=true"):
             # Failed
             actions.append("exit 1")
         os.system("cat log.txt")
--- a/.builder/actions/aws_crt_java_test.py
+++ b/.builder/actions/aws_crt_java_test.py
@@ -10,6 +10,7 @@ class AWSCrtJavaTest(Builder.Action):
         env.shell.setenv('AWS_CRT_MEMORY_TRACING', '2')
         actions = []
 
+        os.system("echo TESTING")
         if os.system("mvn -B test -DredirectTestOutputToFile=true \
             -DrerunFailingTestsCount=5 -DskipAfterFailureCount=1 -DreuseForks=false \
             -Daws.crt.memory.tracing=2 -Daws.crt.debugnative=true"):
--- a/.builder/actions/aws_crt_java_test.py
+++ b/.builder/actions/aws_crt_java_test.py
@@ -10,7 +10,7 @@ class AWSCrtJavaTest(Builder.Action):
         env.shell.setenv('AWS_CRT_MEMORY_TRACING', '2')
         actions = []
 
-        if os.system("mvn -B test -DredirectTestOutputToFile=true -DforkCount=0 \
+        if os.system("mvn -B test -DredirectTestOutputToFile=true \
             -DrerunFailingTestsCount=5 -DskipAfterFailureCount=1 -DreuseForks=false \
             -Daws.crt.memory.tracing=2 -Daws.crt.debugnative=true"):
             # Failed
--- a/.builder/actions/aws_crt_java_test.py
+++ b/.builder/actions/aws_crt_java_test.py
@@ -11,8 +11,7 @@ class AWSCrtJavaTest(Builder.Action):
         actions = []
 
         os.system("echo TESTING")
-        if os.system("mvn -B test -DredirectTestOutputToFile=true \
-            -DrerunFailingTestsCount=5 -DskipAfterFailureCount=1 -DreuseForks=false \
+        if os.system("mvn -P continuous-integration -B test -DredirectTestOutputToFile=true -DreuseForks=false \
             -Daws.crt.memory.tracing=2 -Daws.crt.debugnative=true"):
             # Failed
             actions.append("exit 1")
--- a/.builder/actions/aws_crt_java_test.py
+++ b/.builder/actions/aws_crt_java_test.py
@@ -11,7 +11,7 @@ class AWSCrtJavaTest(Builder.Action):
         actions = []
 
         if os.system("mvn -B test -DredirectTestOutputToFile=true -DforkCount=0 \
-            -DrerunFailingTestsCount=5 -DskipAfterFailureCount=1 \
+            -DrerunFailingTestsCount=5 -DskipAfterFailureCount=1 -DreuseForks=false \
             -Daws.crt.memory.tracing=2 -Daws.crt.debugnative=true"):
             # Failed
             actions.append("exit 1")
--- a/.builder/actions/aws_crt_java_test.py
+++ b/.builder/actions/aws_crt_java_test.py
@@ -11,8 +11,14 @@ class AWSCrtJavaTest(Builder.Action):
         actions = []
 
         os.system("echo TESTING")
-        if os.system("mvn -P continuous-integration -B test -DredirectTestOutputToFile=true -DreuseForks=false \
-            -DrerunFailingTestsCount=5 -Daws.crt.memory.tracing=2 -Daws.crt.debugnative=true"):
+        all_test_result = os.system("mvn -P continuous-integration -B test -DredirectTestOutputToFile=true -DreuseForks=false \
+            -DrerunFailingTestsCount=5 -Daws.crt.memory.tracing=2 -Daws.crt.debugnative=true")
+
+        env.shell.setenv('AWS_CRT_SHUTDOWN_TESTING', '1')
+        shutdown_test_result = os.system("mvn -P continuous-integration -B test -DredirectTestOutputToFile=true -DreuseForks=false \
+            -Daws.crt.memory.tracing=2 -Daws.crt.debugnative=true -Dtest=ShutdownTest")
+
+        if shutdown_test_result or all_test_result:
             # Failed
             actions.append("exit 1")
         os.system("cat log.txt")
--- a/src/native/aws_signing.c
+++ b/src/native/aws_signing.c
@@ -51,16 +51,14 @@ struct s_aws_sign_request_callback_data {
     struct aws_credentials *credentials;
 };
 
-static void s_cleanup_callback_data(struct s_aws_sign_request_callback_data *callback_data) {
-
-    /********** JNI ENV ACQUIRE **********/
-    JNIEnv *env = aws_jni_acquire_thread_env(callback_data->jvm);
-    if (env != NULL) {
-        /* If we can't get an environment, then the JVM is probably shutting down.  Don't crash. */
+static void s_cleanup_callback_data(struct s_aws_sign_request_callback_data *callback_data, JNIEnv *env) {
+    if (callback_data == NULL || env == NULL) {
         return;
     }
 
-    (*env)->DeleteGlobalRef(env, callback_data->java_signing_result_future);
+    if (callback_data->java_signing_result_future != NULL) {
+        (*env)->DeleteGlobalRef(env, callback_data->java_signing_result_future);
+    }
 
     if (callback_data->java_original_request != NULL) {
         (*env)->DeleteGlobalRef(env, callback_data->java_original_request);
@@ -108,9 +106,6 @@ static void s_cleanup_callback_data(struct s_aws_sign_request_callback_data *cal
     if (callback_data->java_previous_signature != NULL) {
         (*env)->DeleteGlobalRef(env, callback_data->java_previous_signature);
     }
-
-    aws_jni_release_thread_env(callback_data->jvm, env);
-    /********** JNI ENV RELEASE **********/
 
     aws_mem_release(aws_jni_get_allocator(), callback_data);
 }
@@ -246,10 +241,11 @@ static void s_aws_request_signing_complete(struct aws_signing_result *result, in
 
 done:
 
-    aws_jni_release_thread_env(callback_data->jvm, env);
-    /********** JNI ENV RELEASE **********/
+    JavaVM *jvm = callback_data->jvm;
+    s_cleanup_callback_data(callback_data, env);
 
-    s_cleanup_callback_data(callback_data);
+    aws_jni_release_thread_env(jvm, env);
+    /********** JNI ENV RELEASE **********/
 }
 
 static void s_aws_chunk_like_signing_complete(struct aws_signing_result *result, int error_code, void *userdata) {
@@ -272,10 +268,11 @@ static void s_aws_chunk_like_signing_complete(struct aws_signing_result *result,
 
 done:
 
-    aws_jni_release_thread_env(callback_data->jvm, env);
-    /********** JNI ENV RELEASE **********/
+    JavaVM *jvm = callback_data->jvm;
+    s_cleanup_callback_data(callback_data, env);
 
-    s_cleanup_callback_data(callback_data);
+    aws_jni_release_thread_env(jvm, env);
+    /********** JNI ENV RELEASE **********/
 }
 
 static void s_aws_chunk_signing_complete(struct aws_signing_result *result, int error_code, void *userdata) {
@@ -396,7 +393,6 @@ void JNICALL Java_software_amazon_awssdk_crt_auth_signing_AwsSigner_awsSignerSig
     jobject java_signing_result_future) {
 
     (void)jni_class;
-    (void)env;
 
     struct aws_allocator *allocator = aws_jni_get_allocator();
     struct s_aws_sign_request_callback_data *callback_data =
@@ -455,7 +451,7 @@ void JNICALL Java_software_amazon_awssdk_crt_auth_signing_AwsSigner_awsSignerSig
 
 on_error:
 
-    s_cleanup_callback_data(callback_data);
+    s_cleanup_callback_data(callback_data, env);
 }
 
 JNIEXPORT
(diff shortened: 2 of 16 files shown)


src/native/crt.c
```diff
@@ -47,6 +49,151 @@ struct aws_allocator *aws_jni_get_allocator() {
     return s_allocator;
 }
 
+void s_detach_jvm_from_thread(void *user_data) {
+    AWS_LOGF_DEBUG(AWS_LS_COMMON_GENERAL, "s_detach_jvm_from_thread invoked");
+    JavaVM *jvm = user_data;
+    (*jvm)->DetachCurrentThread(jvm);
```
graebm, line 55: Seems like a crash can still happen here. Shouldn't we check here if the jvm has already exited before we try and use it?

src/native/crt.c
```diff
@@ -47,6 +49,151 @@ struct aws_allocator *aws_jni_get_allocator() {
...
+read locks on a shared rw lock.  Only when the JVM shutdown hook calls into native is there read-write contention.
+ */
+static struct aws_rw_lock s_jvm_table_lock = AWS_RW_LOCK_INIT;
+static struct aws_hash_table *s_jvms = NULL;
+
+static void s_jvm_table_add_jvm_for_env(JNIEnv *env) {
+    aws_rw_lock_wlock(&s_jvm_table_lock);
+
+    if (s_jvms == NULL) {
+        s_jvms = aws_mem_calloc(aws_default_allocator(), 1, sizeof(struct aws_hash_table));
```
graebm, line 119: trivial: add comment about using default allocator so it doesn't show up in memory tracking stats. right? that's why you're using `aws_default_allocator()`?

src/native/crt.c
```diff
@@ -47,6 +49,151 @@ struct aws_allocator *aws_jni_get_allocator() {
...
+        AWS_FATAL_ASSERT(
+            AWS_OP_SUCCESS ==
+            aws_hash_table_init(s_jvms, aws_default_allocator(), 1, aws_hash_ptr, aws_ptr_eq, NULL, NULL));
+    }
+
+    JavaVM *jvm = NULL;
+    jint jvmresult = (*env)->GetJavaVM(env, &jvm);
+    AWS_FATAL_ASSERT(jvmresult == 0 && jvm != NULL);
+
+    AWS_FATAL_ASSERT(AWS_OP_SUCCESS == aws_hash_table_put(s_jvms, jvm, NULL, NULL));
```
graebm, line 129: trivial: should we do a sanity check that the JVM isn't already in the table?

src/native/crt.c
```diff
@@ -212,40 +359,38 @@ struct aws_string *aws_jni_new_string_from_jstring(JNIEnv *env, jstring str) {
...
-            return NULL;
-        }
-        /* This should only happen in event loop threads, the JVM main thread attachment is
-         * managed by the JVM, so we only need to clean up event loop thread attachments */
-        AWS_FATAL_ASSERT(AWS_OP_SUCCESS == aws_thread_current_at_exit(s_detach_jvm_from_thread, (void *)jvm));
-    }
+/* clang-format off */
+static struct aws_error_info s_crt_errors[] = {
+    AWS_DEFINE_ERROR_INFO_CRT(
+        AWS_ERROR_JAVA_CRT_JVM_DESTROYED,
```
graebm, line 367: 🍻
good work
every CRT should be defining its own errors

src/test/java/software/amazon/awssdk/crt/test/ShutdownTest.java
```diff
@@ -0,0 +1,52 @@
...
+            return HttpClientConnectionManager.create(options);
+        }
+    }
+
+    @Test
+    public void testShutdownDuringAcquire() throws Exception {
+        Assume.assumeTrue(doShutdownTest());
+
+        HttpClientConnectionManager manager = createConnectionManager(new URI("https://aws-crt-test-stuff.s3.amazonaws.com"));
+        CompletableFuture<HttpClientConnection> connection = manager.acquireConnection();
```
graebm, line 50: add some comments about exactly what this test is doing and why it's so weird

.builder/actions/aws_crt_java_test.py
```diff
@@ -10,9 +10,15 @@ def run(self, env):
...
-        if os.system("mvn -B test -DredirectTestOutputToFile=true -DforkCount=0 \
-            -DrerunFailingTestsCount=5 -DskipAfterFailureCount=1 \
-            -Daws.crt.memory.tracing=2 -Daws.crt.debugnative=true"):
+        os.system("echo TESTING")
+        all_test_result = os.system("mvn -P continuous-integration -B test -DredirectTestOutputToFile=true -DreuseForks=false \
+            -DrerunFailingTestsCount=5 -Daws.crt.memory.tracing=2 -Daws.crt.debugnative=true")
+
+        env.shell.setenv('AWS_CRT_SHUTDOWN_TESTING', '1')
+        shutdown_test_result = os.system("mvn -P continuous-integration -B test -DredirectTestOutputToFile=true -DreuseForks=false \
+            -Daws.crt.memory.tracing=2 -Daws.crt.debugnative=true -Dtest=ShutdownTest")
```
graebm, line 19: add a comment about why this is being run separately

actually ... maybe this is just too weird to be run as a literal unit test?
Would this be better as a `static void main()` function we can run as a standalone application?
Since it's a weird thing that doesn't work properly unless it's the ONLY test running?
It's probably fine for the code to still live under the `test/` directory

and then it would be easier to make this test more complex. Like the python test takes an arg of which step it bails out at. We could do the same thing here, running the app in a loop, bailing out at different steps

.builder/actions/aws_crt_java_test.py
```diff
@@ -10,9 +10,15 @@ def run(self, env):
         env.shell.setenv('AWS_CRT_MEMORY_TRACING', '2')
         actions = []
 
-        if os.system("mvn -B test -DredirectTestOutputToFile=true -DforkCount=0 \
-            -DrerunFailingTestsCount=5 -DskipAfterFailureCount=1 \
-            -Daws.crt.memory.tracing=2 -Daws.crt.debugnative=true"):
+        os.system("echo TESTING")
```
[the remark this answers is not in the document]
bretambrose: That needs to be reverted

src/native/credentials_provider.c
```diff
@@ -654,7 +661,7 @@ struct aws_credentials_provider_get_credentials_callback_data {
 };
 
 static void s_cp_callback_data_clean_up(struct aws_credentials_provider_get_credentials_callback_data *callback_data) {
-    JNIEnv *env = aws_jni_get_thread_env(callback_data->jvm);
+    JNIEnv *env = aws_jni_acquire_thread_env(callback_data->jvm);
```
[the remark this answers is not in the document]
bretambrose: At least in this particular case, there is no re-entrance because in the two places it's used from a potential native thread, it releases before invoking the callback.

We could update the cleanup callback to take the env and release it when called from a function that had acquired the env (rather than had it passed in by Java).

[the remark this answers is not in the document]
bretambrose: Fixing the potential re-entrant issue would likely require thread-local storage.

src/native/aws_signing.c
```diff
@@ -214,7 +216,7 @@ static void s_aws_request_signing_complete(struct aws_signing_result *result, in
 
     struct s_aws_sign_request_callback_data *callback_data = userdata;
 
-    JNIEnv *env = aws_jni_get_thread_env(callback_data->jvm);
+    JNIEnv *env = aws_jni_acquire_thread_env(callback_data->jvm);
     if (env == NULL) {
         /* If we can't get an environment, then the JVM is probably shutting down.  Don't crash. */
         goto done;
```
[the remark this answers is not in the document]
bretambrose: While we won't clean up the JVM objects, we will free the native callback structure.

src/native/credentials_provider.c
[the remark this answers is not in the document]
bretambrose: Documenting discussion: we're going to use TryAcquire to mitigate re-entrancy issues, acknowledging that TryAcquire does not work on Windows 7.  If we get a Windows 7 Java ticket, the plan would be to mitigate either as an option (customer accepts shutdown crash risk) or use a full mutex (only in this particular case) which would be a substantial performance degradation.

src/native/crt.c
```diff
@@ -47,6 +49,151 @@ struct aws_allocator *aws_jni_get_allocator() {
...
+done:
+
+    aws_rw_lock_wunlock(&s_jvm_table_lock);
+}
+
+JNIEnv *aws_jni_acquire_thread_env(JavaVM *jvm) {
+    aws_rw_lock_rlock(&s_jvm_table_lock);
+
+    if (s_jvms == NULL) {
+        aws_raise_error(AWS_ERROR_JAVA_CRT_JVM_DESTROYED);
```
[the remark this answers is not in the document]
bretambrose: Per discussion, going to leave as is for now; wouldn't play nice if we ever have to support the multi-JVM case.

src/native/crt.c
bretambrose: It's not clear what you'd do in that case as it implies (at least currently) init being called multiple times which we don't support.  Could just crash I suppose.

src/native/aws_signing.c
```diff
@@ -53,8 +53,9 @@ struct s_aws_sign_request_callback_data {
 
 static void s_cleanup_callback_data(struct s_aws_sign_request_callback_data *callback_data) {
 
+    /********** JNI ENV ACQUIRE **********/
     JNIEnv *env = aws_jni_acquire_thread_env(callback_data->jvm);
-    if (env == NULL) {
+    if (env != NULL) {
         /* If we can't get an environment, then the JVM is probably shutting down.  Don't crash. */
         return;
```
graebm, line 60: shouldn't it be `if (env == NULL)`?

src/native/crt.c
```diff
@@ -49,10 +49,19 @@ struct aws_allocator *aws_jni_get_allocator() {
...
+
+    /* we don't need this JNIEnv, but this is an easy way to verify the JVM is still valid to use */
+    /********** JNI ENV ACQUIRE **********/
+    JNIEnv *env = aws_jni_acquire_thread_env(jvm);
+    (void)env;
+
     (*jvm)->DetachCurrentThread(jvm);
+
+    aws_jni_release_thread_env(jvm, env);
+    /********** JNI ENV RELEASE **********/
```
graebm, line 64: did you mean to add an if-check?
```
if (env != NULL)
    (*jvm)->DetachCurrentThread(jvm);
```

src/native/crt.c
```diff
@@ -156,7 +168,9 @@ static void s_jvm_table_remove_jvm_for_env(JNIEnv *env) {
 }
 
 JNIEnv *aws_jni_acquire_thread_env(JavaVM *jvm) {
-    aws_rw_lock_rlock(&s_jvm_table_lock);
+    if (aws_rw_lock_try_rlock(&s_jvm_table_lock)) {
+        return NULL;
```
graebm, line 172: trivial: I would add `aws_raise_error(AWS_ERROR_JAVA_CRT_JVM_DESTROYED);`
otherwise the errors is just going to be something like `WAAA_MY_LOCK_FAILED`

also, add comments about why we're using try-lock here, because the reasoning is complicated

src/native/credentials_provider.c
```diff
@@ -654,7 +666,9 @@ struct aws_credentials_provider_get_credentials_callback_data {
 };
 
 static void s_cp_callback_data_clean_up(struct aws_credentials_provider_get_credentials_callback_data *callback_data) {
-    JNIEnv *env = aws_jni_get_thread_env(callback_data->jvm);
+
+    /********** JNI ENV ACQUIRE **********/
+    JNIEnv *env = aws_jni_acquire_thread_env(callback_data->jvm);
```
graebm, line 671: I still think `env` should just be passed to this function as a param.
This is a helper function, called from 2 places, both of which already have `env` available

src/native/aws_signing.c
```diff
@@ -53,8 +53,9 @@ struct s_aws_sign_request_callback_data {
 
 static void s_cleanup_callback_data(struct s_aws_sign_request_callback_data *callback_data) {
 
-    JNIEnv *env = aws_jni_get_thread_env(callback_data->jvm);
-    if (env == NULL) {
+    /********** JNI ENV ACQUIRE **********/
+    JNIEnv *env = aws_jni_acquire_thread_env(callback_data->jvm);
```
graebm, line 57: this is a helper function, not a callback, we could just pass `env` in as an argument

src/native/event_stream_rpc_client.c
```diff
@@ -40,7 +40,8 @@ static void s_destroy_connection_callback_data(struct connection_callback_data *
         return;
     }
 
-    JNIEnv *env = aws_jni_get_thread_env(callback_data->jvm);
+    /********** JNI ENV ACQUIRE **********/
+    JNIEnv *env = aws_jni_acquire_thread_env(callback_data->jvm);
```
graebm, line 44: `s_destroy_connection_callback_data()` is a helper function, not a callback, we could just pass `env` in as an argument

src/native/http_connection_manager.c
```diff
@@ -51,7 +51,8 @@ static void s_destroy_manager_binding(struct http_connection_manager_binding *bi
         return;
     }
 
-    JNIEnv *env = aws_jni_get_thread_env(binding->jvm);
+    /********** JNI ENV ACQUIRE **********/
+    JNIEnv *env = aws_jni_acquire_thread_env(binding->jvm);
```
graebm, line 55: `s_destroy_manager_binding()` is a helper function, not a callback, we could just pass `env` in as an argument

src/native/http_connection_manager.c
```diff
@@ -332,7 +341,8 @@ static void s_destroy_connection_binding(struct aws_http_connection_binding *bin
...
```
graebm, line 345: `s_destroy_connection_binding()` is a helper function, not a callback, we could just pass `env` in as an argument

src/native/mqtt_connection.c
```diff
@@ -101,30 +101,33 @@ static struct mqtt_jni_async_callback *mqtt_jni_async_callback_new(
     struct mqtt_jni_connection *connection,
     jobject async_callback) {
 
-    JNIEnv *env = aws_jni_get_thread_env(connection->jvm);
+    /********** JNI ENV ACQUIRE **********/
+    JNIEnv *env = aws_jni_acquire_thread_env(connection->jvm);
```
graebm, line 105: `mqtt_jni_async_callback_new()` is a helper function, not a callback, we could just pass `env` in as an argument

src/native/mqtt_connection.c
```diff
@@ -101,30 +101,33 @@ static struct mqtt_jni_async_callback *mqtt_jni_async_callback_new(
...
 
     return callback;
 }
 
 static void mqtt_jni_async_callback_destroy(struct mqtt_jni_async_callback *callback) {
     AWS_FATAL_ASSERT(callback && callback->connection);
-    JNIEnv *env = aws_jni_get_thread_env(callback->connection->jvm);
+
+    /********** JNI ENV ACQUIRE **********/
+    JNIEnv *env = aws_jni_acquire_thread_env(callback->connection->jvm);
```
graebm, line 130: `mqtt_jni_async_callback_destroy()` is a helper function, not a callback, we could just pass `env` in as an argument

src/native/mqtt_connection.c
```diff
@@ -202,7 +215,8 @@ static void s_on_connection_interrupted_internal(
     int error_code,
     jobject ack_callback) {
 
-    JNIEnv *env = aws_jni_get_thread_env(connection->jvm);
+    /********** JNI ENV ACQUIRE **********/
+    JNIEnv *env = aws_jni_acquire_thread_env(connection->jvm);
```
graebm, line 219: `s_on_connection_interrupted_internal()` is a helper function, not a callback, we could just pass `env` in as an argument

src/native/mqtt_connection.c
```diff
@@ -538,14 +568,18 @@ static void s_deliver_ack_success(struct mqtt_jni_async_callback *callback) {
     AWS_FATAL_ASSERT(callback->connection);
 
     if (callback->async_callback) {
-        JNIEnv *env = aws_jni_get_thread_env(callback->connection->jvm);
+        /********** JNI ENV ACQUIRE **********/
+        JNIEnv *env = aws_jni_acquire_thread_env(callback->connection->jvm);
```
graebm, line 572: `s_deliver_ack_success()` is a helper function, not a callback, we could just pass env in as an argument

src/native/mqtt_connection.c
```diff
@@ -554,7 +588,8 @@ static void s_deliver_ack_failure(struct mqtt_jni_async_callback *callback, int
...
```
graebm, line 592: `s_deliver_ack_failure()` is a helper function, not a callback, we could just pass env in as an argument

src/native/crt.c
bretambrose: I left the unimplemented error alone in my update.

src/native/crt.h
```diff
@@ -162,6 +162,7 @@ struct aws_byte_cursor aws_jni_byte_cursor_from_direct_byte_buffer(JNIEnv *env,
 struct aws_string *aws_jni_new_string_from_jstring(JNIEnv *env, jstring str);
 
 /*******************************************************************************
+<<<<<<< HEAD
```
MikeDombo, line 165: bad merge

src/native/crt.h
```diff
@@ -175,4 +175,11 @@ JNIEnv *aws_jni_acquire_thread_env(JavaVM *jvm);
  ******************************************************************************/
 void aws_jni_release_thread_env(JavaVM *jvm, JNIEnv *env);
 
+/*******************************************************************************
+ * aws_jni_new_crt_exception_from_error_code - Creates a new jobject from the aws
+ * error code, which is the type of software/amazon/awssdk/crt/CrtRuntimeException.
+ * Reference of the jobject needed to be cleaned up after use.
+ ******************************************************************************/
+jobject aws_jni_new_crt_exception_from_error_code(JNIEnv *env, int error_code);
```
graebm, line 183: OMG how have we gotten this far without already having this function?

graebm (approved):
💦🔌💦
Sloppy Shutdown 4ever

src/native/crt.h
bretambrose: That's all Dengke.

PR #493, opened 2022-05-28
Clean up .dll files on Windows

graebm:
**ISSUE:** On Windows, the .dll file extracted to the temp directory was never being deleted. And repeated runs of the program were filling the disk with old .dll files.

**DIAGNOSIS:** We were using [File.deleteOnExit()](https://docs.oracle.com/javase/8/docs/api/java/io/File.html#deleteOnExit--) to clean up the shared lib, which is working OK on other platforms, but not on Windows. On Windows, files cannot be deleted while they're in use, and it appears that Java doesn't try to "unload" the .dll, so it's apparently impossible for a Java process to delete a .dll that it's using.

I did an experiment with a dead-simple JNI library. I simply loaded it, and called `File.deleteOnExit()`. That .dll didn't get deleted either. So it's not just a problem with our .dll being very complex.

**DESCRIPTION OF CHANGES:** On Windows only, during startup, scan the temp dir for old instances of `AWSCRT_*aws-crt-jni.dll` and try to delete them.

**SIDE-NOTE:**
I realized that on non-Windows platforms, the shared-lib won't be deleted on exit if Java dies suddenly (crash in C, OS terminated process, etc). But on non-Windows platforms you can delete the shared-lib from disk immediately after loading it, you don't need to wait until the process exits. I did experiments on Mac and this works. But I left it out of this PR because this is an emergency fix for windows and I don't want to introduce non-necessary changes.

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/CRT.java
+++ b/src/main/java/software/amazon/awssdk/crt/CRT.java
@@ -149,7 +149,7 @@ public final class CRT {
             }
 
             // Prefix the lib
-            String prefix = "AWSCRT_";
+            String prefix = "AWSCRT_TEST";
             String libraryName = System.mapLibraryName(CRT_LIB_NAME);
             String libraryPath = "/" + getOSIdentifier() + "/" + getArchIdentifier() + "/" + libraryName;
 
--- a/src/main/java/software/amazon/awssdk/crt/CRT.java
+++ b/src/main/java/software/amazon/awssdk/crt/CRT.java
@@ -155,20 +155,19 @@ public final class CRT {
 
             File tempSharedLib = File.createTempFile(prefix, libraryName, tmpdirFile);
 
-			// The lib should be deleted when we're done using it.
-			// Ask Java to try and delete it on exit, we do this immediately
+			// The temp lib file should be deleted when we're done with it.
+			// Ask Java to try and delete it on exit. We call this immediately
 			// so that if anything goes wrong writing the file to disk, or
 			// loading it as a shared lib, it will still get cleaned up.
 			tempSharedLib.deleteOnExit();
 
-
-            // Unfortunately File.deleteOnExit() won't work on Windows,
-			// where files cannot be deleted while they're in use.
-            // Once our .dll is loaded, it can't be deleted by this process.
+            // Unfortunately File.deleteOnExit() won't work on Windows, where
+            // files cannot be deleted while they're in use. On Windows, once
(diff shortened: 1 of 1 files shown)


src/main/java/software/amazon/awssdk/crt/CRT.java
```diff
@@ -149,13 +150,50 @@ private static void extractAndLoadLibrary(String path) {
...
+			tempSharedLib.deleteOnExit();
+
+            // Unfortunately File.deleteOnExit() won't work on Windows, where
+            // files cannot be deleted while they're in use. On Windows, once
+            // our .dll is loaded, it can't be deleted by this process.
+            //
+            // The Windows-only solution to this problem is to scan on startup
+            // for old instances of the .dll and try to delete them. If another
+            // process is still using the .dll, the delete will fail, which is fine.
+            String os = getOSIdentifier();
```
bretambrose, line 171: Let's split this off in a helper function

src/main/java/software/amazon/awssdk/crt/CRT.java
```diff
@@ -149,13 +150,50 @@ private static void extractAndLoadLibrary(String path) {
...
+            if (os == "windows") {
```
bretambrose, line 172: Feel like platform id shouldn't be a string, but that's a larger change.

src/main/java/software/amazon/awssdk/crt/CRT.java
```diff
@@ -149,13 +150,50 @@ private static void extractAndLoadLibrary(String path) {
             }
 
             // Prefix the lib
-            String prefix = "AWSCRT_" + new Date().getTime();
+            String prefix = "AWSCRT_";
```
bretambrose, line 153: nativeLibraryNamePrefix or similar maybe

bretambrose (approved):

sbSteveK (approved):

src/main/java/software/amazon/awssdk/crt/CRT.java
MikeDombo: .equals()?

bretambrose: Another great reason to make it not-a-string!

graebm: I started making it not-a-string but it was turning into a big refactor so going with `.equals()` as suggested by Dombo

PR #510, opened 2022-07-14
(Possible) Mqtt disconnect shutdown hang fix

TwistedTwigleg:
*Description of changes:*

Adds a new function to check if the JNI layer is no longer safe to call and then checks this before trying to perform MQTT operations. This fixes a small hang on Windows in some situations where a promise is waiting to be finished but the JVM is shutting down and no longer safe to call.

________

### DIFF
--- a/src/native/crt.c
+++ b/src/native/crt.c
@@ -217,6 +217,34 @@ void aws_jni_release_thread_env(JavaVM *jvm, JNIEnv *env) {
     }
 }
 
+JNIEnv *aws_jni_shutdown_acquire_thread_env(JavaVM *jvm) {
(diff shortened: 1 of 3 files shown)


bretambrose (commented):
s_jvms isn't valid to be accessed unless the associated rw lock is held

Also I don't think we want to do this overall.  There are cases that this doesn't address and there's no way to address them: if the promise was already created at shutdown time, this doesn't help and it will still never be resolved.  I would rather GG explore a way to conditionally not wait on the promise.

TwistedTwigleg:
Okay, that makes sense. Thank you for reviewing! I'll close this PR then 👍

PR #513, opened 2022-07-16
Custom TLS Key operation support

TwistedTwigleg:
*Description of changes:*

Adds custom key operation support for the TLS handshake when connecting with MQTT, to allow for making your own private key operations.

This PR makes the changes to expose the custom key operation functionality to the CRT so it can be used in Java. It also includes a sample and tests that run in CI.

_______________


src/main/java/software/amazon/awssdk/crt/io/TlsKeyOperation.java
```diff
@@ -0,0 +1,226 @@
...
+ * A class containing a mutual TLS (mTLS) Private Key operation that needs to be performed.
+ * This class is passed to TlsKeyOperationHandler if a custom key operation is set.
+ *
+ * You MUST call either complete(output) or completeExceptionally(exception)
+ * or the TLS connection will hang forever!
+ *
+ * You do NOT need to call close on this CrtResource - it will be called automatically
+ * when you call either complete(output) or completeExceptionally(exception)
+ */
+public final class TlsKeyOperation extends CrtResource {
```
graebm, line 26: I'm worried about this being a CrtResource, it gives users more functions that they might call incorrectly (`close()`, `acquireRef()`)

This function already has its own special lifetime rules, that one of the complete() functions must be called

how about we make this a normal class

but store the nativeHandle as some private member, which still IS a CrtResource (so we get leak tracking)

src/main/java/software/amazon/awssdk/crt/io/TlsKeyOperation.java
```diff
@@ -0,0 +1,226 @@
...
+     */
+    public void complete(byte[] output) {
+        if (getNativeHandle() == 0 || this.clearCalled == true) {
+            Log.log(LogLevel.Error, LogSubject.CommonGeneral,
+                "No native handle set in TlsKeyOperation! Cannot complete operation");
+            return;
+        }
+
+        tlsKeyOperationComplete(getNativeHandle(), output);
+        clear();
```
graebm, line 134: need SOME kind of synchronization somewhere
either in java, or C

Java has a cool `synchronize` keyword that makes this kind of thing simple

src/native/custom_key_op_handler.c
```diff
@@ -0,0 +1,170 @@
...
+static void s_aws_custom_key_op_handler_perform_operation(
+    struct aws_custom_key_op_handler *key_op_handler,
+    struct aws_tls_key_operation *operation) {
+
+    struct aws_jni_custom_key_op_handler *op_handler = (struct aws_jni_custom_key_op_handler *)key_op_handler->impl;
+    AWS_FATAL_ASSERT(op_handler != NULL);
+
+    /* Get the Java ENV */
+    JNIEnv *env = aws_jni_acquire_thread_env(op_handler->jvm);
+    AWS_FATAL_ASSERT(env != NULL);
```
graebm, line 16: merge from main, and check out how other code is calling `aws_jni_acquire_thread_env()` now

the rules have changed

src/native/custom_key_op_handler.c
```diff
@@ -0,0 +1,170 @@
...
+
+    jbyteArray jni_input_data = NULL;
+    jobject jni_operation = NULL;
+    bool success = false;
+    bool exception_occured = false;
+
+    if (operation == NULL) {
+        goto clean_up;
```
graebm, line 24: just assert this is true.
it should never happen, let's not add real error handling

src/native/custom_key_op_handler.c
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    /**
+     * Invoke TlsKeyOperationHandler.performOperation() through the invokePerformOperation
+     * function. This function will also catch any exceptions and clear the operation
+     * with an exception should it occur.
+     */
+    (*env)->CallVoidMethod(
+        env, op_handler->jni_custom_key_op, tls_key_operation_handler_properties.perform_operation_id, jni_operation);
```
graebm, line 56: my idea:
DON'T call the virtual function directly.
INSTEAD call some internal helper function, that calls the virtual function in a try/catch block, and then does op.completeExceptionally() entirely out in Java.

and our Java TlsKeyOperation class is protected against double-complete, so we're all good

and this function is probably impossible to fail, so we can just FATAL_ASSERT or something

src/native/custom_key_op_handler.c
```diff
@@ -0,0 +1,170 @@
...
+struct aws_jni_custom_key_op_handler *aws_custom_key_op_handler_java_new(
+    JNIEnv *env,
+    struct aws_allocator *allocator,
+    jobject jni_custom_key_op) {
+
+    struct aws_jni_custom_key_op_handler *java_custom_key_op_handler =
+        aws_mem_calloc(allocator, 1, sizeof(struct aws_jni_custom_key_op_handler));
+    if (java_custom_key_op_handler == NULL) {
+        return NULL;
+    }
```
graebm, line 126: aws allocation can't fail

src/native/custom_key_op_handler.c
```diff
@@ -0,0 +1,170 @@
...
+
+    aws_ref_count_init(
+        &java_custom_key_op_handler->key_handler.ref_count,
+        &java_custom_key_op_handler->key_handler,
+        (aws_simple_completion_callback *)s_aws_custom_key_op_handler_destroy);
+    java_custom_key_op_handler->key_handler.vtable = &s_aws_custom_key_op_handler_vtable;
+    java_custom_key_op_handler->key_handler.impl = (void *)java_custom_key_op_handler;
+    /* Make a global reference so the Java interface is kept alive */
+    java_custom_key_op_handler->jni_custom_key_op = (*env)->NewGlobalRef(env, jni_custom_key_op);
```
graebm, line 141: FATAL_ASSERT that NewGlobalRef succeeded

src/native/custom_key_op_handler.h
```diff
@@ -0,0 +1,61 @@
...
+#include "java_class_ids.h"
+#include "tls_context_pkcs11_options.h"
+
+/**
+ * A struct that contains all the data for a Java custom_key_op_handler that is exposed
+ * in Java so the customer can perform their own custom private key operation handler functions.
+ */
+struct aws_jni_custom_key_op_handler {
+    JavaVM *jvm;
+
```
graebm, line 21: you can just forward declare this, and define it in the .c file

src/native/custom_key_op_handler.h
```diff
@@ -0,0 +1,61 @@
...
+    /** The allocator to use */
+    struct aws_allocator *allocator;
+
+    /**
+     * Certificate's file path on disk (UTF-8).
+     * The certificate must be PEM formatted and UTF-8 encoded.
+     * Zero out if passing in certificate by some other means (such as file contents).
+     * (Can also be zero out if it is unused, like in PKCS11 implementation)
+     */
+    struct aws_byte_cursor cert_file_path;
```
graebm, line 43: you don't need this anymore

src/native/custom_key_op_handler.h
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * Certificate's file contents (UTF-8).
+     * The certificate must be PEM formatted and UTF-8 encoded.
+     * Zero out if passing in certificate by some other means (such as file path).
+     * (Can also be zero out if it is unused, like in PKCS11 implementation)
+     */
+    struct aws_byte_cursor cert_file_contents;
```
graebm, line 51: you don't need this anymore

src/native/custom_key_op_handler.h
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+#include <jni.h>
+
+#include <aws/common/string.h>
+#include <aws/io/tls_channel_handler.h>
+
+#include "crt.h"
+#include "java_class_ids.h"
+#include "tls_context_pkcs11_options.h"
```
graebm, line 13: you can probably move some of these #includes into the .c file instead

src/native/custom_key_op_handler.h
```diff
@@ -0,0 +1,61 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#include <jni.h>
```
graebm, line 6: missing IFNDEF for this header

src/native/tls_context_options.c
```diff
@@ -62,9 +67,14 @@ static void s_jni_tls_ctx_options_destroy(struct jni_tls_ctx_options *tls) {
 
     aws_tls_ctx_pkcs11_options_from_java_destroy(tls->pkcs11_options);
 
+    struct aws_allocator *allocator = aws_jni_get_allocator();
+    if (tls->custom_key_op_handler != NULL) {
+        /* Release the reference to the Java class */
+        aws_custom_key_op_handler_java_release(allocator, tls->custom_key_op_handler);
```
graebm, line 73: trivial: don't need to null-check

src/native/custom_key_op_handler.c
```diff
@@ -0,0 +1,170 @@
...
+
+    AWS_LOGF_DEBUG(
+        AWS_LS_COMMON_IO,
+        "java_custom_key_op_handler=%p: Initalizing Custom Key Operations",
+        (void *)java_custom_key_op_handler);
+    return java_custom_key_op_handler;
+}
+
+void aws_custom_key_op_handler_java_release(
+    struct aws_allocator *allocator,
```
graebm, line 152: you do not need to pass an `allocator` to this function

just call aws_jni_get_allocator()

src/native/tls_context_options.c
```diff
@@ -145,6 +156,76 @@ jlong JNICALL Java_software_amazon_awssdk_crt_io_TlsContextOptions_tlsContextOpt
...
+
+        jobject jni_custom_key_op_handle = (*env)->GetObjectField(
+            env, jni_custom_key_op, tls_context_custom_key_operation_options_properties.operation_handler_field_id);
+        if (!jni_custom_key_op_handle) {
+            aws_jni_throw_runtime_exception(
+                env, "could not get custom operation handler from jni_custom_key_op_handle!");
+            goto on_error;
+        }
+
+        tls->custom_key_op_handler = aws_custom_key_op_handler_java_new(env, allocator, jni_custom_key_op_handle);
```
graebm, line 169: again, don't need to pass allocators to functions you declare in aws-crt-java

src/native/tls_context_options.c
```diff
@@ -145,6 +156,76 @@ jlong JNICALL Java_software_amazon_awssdk_crt_io_TlsContextOptions_tlsContextOpt
...
+
+        if (jni_custom_key_op_cert_path && jni_custom_key_op_cert_contents) {
+            /* Cannot have both a certificate path and data - use one or the other */
+            aws_jni_throw_runtime_exception(
+                env, "Custom key operation handler: cannot have both certificate file path and certificate contents!");
+            goto on_error;
+        } else if (jni_custom_key_op_cert_contents) {
+            /* If we have the certificate contents, then use it directly */
+            tls->certificate = aws_jni_new_string_from_jstring(env, jni_custom_key_op_cert_contents);
+            struct aws_byte_cursor certificate_byte_cursor = aws_byte_cursor_from_string(tls->certificate);
```
graebm, line 187: not checking result of aws_jni_new_string_from_jstring() here, but we do check it in the next if-block

figure out which approach is the right one

src/native/tls_context_pkcs11_options.c
```diff
@@ -144,24 +144,45 @@ struct aws_tls_ctx_pkcs11_options *aws_tls_ctx_pkcs11_options_from_java_new(JNIE
...
+    if (!binding->cert_file_contents) {
+        if (!s_read_optional_string(
+                env,
+                options_jni,
+                tls_context_pkcs11_options_properties.certificateFileContents,
+                &binding->cert_file_contents,
+                &binding->options.cert_file_contents)) {
+            goto error;
+        }
     }
```
graebm, line 186: these changes can be reverted when you sync to latest aws-c-io

src/main/java/software/amazon/awssdk/crt/io/TlsKeyOperation.java
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    /**
+     * The function to call when you have modified the input data using the private key and are ready to
+     * return it for use in the mutual TLS Handshake.
+     *
+     * @param output The modified input data that has been modified by the custom key operation
+     */
+    public synchronized void complete(byte[] output) {
+        if (nativeResource.internalGetNativeHandle() == 0) {
```
graebm, line 121: trivial: there's a handy `isNull()` function you can call

src/main/java/software/amazon/awssdk/crt/io/TlsKeyOperation.java
```diff
@@ -0,0 +1,189 @@
...
+        };
+
+        /* Exposing acquireNativeHandle */
+        public void internalAcquireNativeHandle(long handle) {
+            acquireNativeHandle(handle);
+        }
+
+        /* Exposing getNativeHandle */
+        public long internalGetNativeHandle() {
+            return getNativeHandle();
```
graebm, line 168: don't need this. getNativeHandle() is already public

src/main/java/software/amazon/awssdk/crt/io/TlsKeyOperation.java
```diff
@@ -0,0 +1,189 @@
...
+        protected void releaseNativeHandle() {}
+
+        protected boolean canReleaseReferencesImmediately() {
+            return true;
+        };
+
+        /* Exposing acquireNativeHandle */
+        public void internalAcquireNativeHandle(long handle) {
+            acquireNativeHandle(handle);
+        }
```
graebm, line 164: trivial: Just have your constructor take the `long handle` and then the constructor can call the protected functions

src/main/java/software/amazon/awssdk/crt/io/TlsKeyOperation.java
```diff
@@ -0,0 +1,189 @@
...
+
+        tlsKeyOperationCompleteExceptionally(nativeResource.internalGetNativeHandle(), ex);
+        nativeResource.close();
+    }
+
+    /**
+     * The TlsKeyOperation has special lifetime rules, where you have to call one of the complete functions, and
+     * by using this private, internal-only CRT resource, we can still get the benefits of using a CRT resource
+     * for detecting memory leaks, while not exposing functionality that would conflict with the lifetime rules.
+     */
```
graebm, line 153: good explanation 👍👍👍👍

src/native/custom_key_op_handler.c
```diff
@@ -4,23 +4,51 @@
...
+     * A reference to the Java class that this struct is linked to.
+     * The interface, strings, etc, can be gotten from this class.
+     */
+    jobject jni_custom_key_op;
+
+    /**
+     * The C class containing the key operations. We extend/define it's VTable to allow
+     * us to have it call into the customer's Java code.
+     */
+    struct aws_custom_key_op_handler key_handler;
```
graebm, line 25: trivial: typically when making "subclasses" in C we'd put this as the first member and name it "base"

src/native/custom_key_op_handler.c
```diff
@@ -4,23 +4,51 @@
...
-    if (operation == NULL) {
+    /* Get the Java ENV */
+    JNIEnv *env = aws_jni_acquire_thread_env(op_handler->jvm);
+    if (env == NULL) {
+        /* JVM is likely shutting down. Do not crash but log error. */
+        AWS_LOGF_ERROR(
+            AWS_LS_COMMON_IO,
+            "java_custom_key_op_handler=%p perform operation: Could not get Java ENV!",
+            (void *)op_handler);
+        success = false;
```
graebm, line 51: success already is false

src/native/custom_key_op_handler.c
```diff
@@ -29,6 +57,7 @@ static void s_aws_custom_key_op_handler_perform_operation(
     jni_input_data = aws_jni_byte_array_from_cursor(env, &input_data);
     if (jni_input_data == NULL) {
         aws_jni_check_and_clear_exception(env);
+        success = false;
```
graebm, line 60: success already is false

src/native/custom_key_op_handler.c
```diff
@@ -44,6 +73,7 @@ static void s_aws_custom_key_op_handler_perform_operation(
         (jint)aws_tls_key_operation_get_digest_algorithm(operation));
     if (jni_operation == NULL) {
         aws_jni_check_and_clear_exception(env);
+        success = false;
```
graebm, line 76: success already is false

src/main/java/software/amazon/awssdk/crt/io/TlsKeyOperation.java
```diff
@@ -0,0 +1,189 @@
...
+            return getNativeHandle();
+        }
+    }
+
+    static private void invokePerformOperation(TlsKeyOperationHandler handler, TlsKeyOperation operation) {
+        try {
+            handler.performOperation(operation);
+        } catch (Exception ex) {
+            Log.log(LogLevel.Error, LogSubject.CommonGeneral,
+                "Exception occured! Completing exceptionally...");
```
graebm, line 177: See if there's some standard way to log a stack trace

src/native/custom_key_op_handler.c
```diff
@@ -52,23 +82,15 @@ static void s_aws_custom_key_op_handler_perform_operation(
...
-        success = true;
-    }
+    (*env)->CallStaticVoidMethod(
+        env,
+        tls_key_operation_properties.cls,
+        tls_key_operation_properties.invoke_operation_id,
+        op_handler->jni_custom_key_op,
+        jni_operation);
+    /* This should never fail, but just to be extra sure... */
+    AWS_FATAL_ASSERT(!aws_jni_check_and_clear_exception(env));
```
graebm, line 92: trivial: 'should never fail because the function we're calling wraps the user callback in a try/catch"

src/native/custom_key_op_handler.c
```diff
@@ -139,24 +157,28 @@ struct aws_jni_custom_key_op_handler *aws_custom_key_op_handler_java_new(
...
+void aws_custom_key_op_handler_java_release(struct aws_custom_key_op_handler *custom_key_op_handler) {
 
+    if (custom_key_op_handler == NULL) {
+        return;
+    }
+    struct aws_jni_custom_key_op_handler *java_custom_key_op_handler =
+        (struct aws_jni_custom_key_op_handler *)custom_key_op_handler->impl;
     if (!java_custom_key_op_handler) {
         return;
     }
```
graebm, line 180: trivial: remove this if-check, just march on and crash

the way it is we're silently ignoring a coding error of ours

graebm (approved):
fix & ship

TwistedTwigleg:
Thanks for all the reviews and help! Merging into `main`...

PR #512, opened 2022-07-15
support new configures for stream manager

TingDaoK:
Supports new configures for stream manager:
  - priorKnowledge
  - closeConnectionOnServerError
  - connectionPingPeriodMs
  - connectionPingTimeoutMs
  - TlsConnectionOptions

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #514, opened 2022-07-22
update submodules

sbSteveK:
Update Submodules


TingDaoK:
I'll update it from https://github.com/awslabs/aws-crt-java/pull/512. I broke the API from aws-c-http. Also, we need to leave s2n and aws-lc out for now, as we need to fix the visibility issue frist

PR #518, opened 2022-08-01
Fix event stream server crash during shutting down

TingDaoK:
*Issue #, if available:*

- If the `on_new_connection` failed, the `connection_callback_data` will not be properly set for the connection. And the channel will start to shutdown.
- During the channel shutdown callback, `s_on_connection_shutdown_fn` will use `connection_callback_data` that was not properly setup. It leads to a crash

*Description of changes:*

- Check the `connection_callback_data` was set or not, if not, nothing needs to be cleaned up from callback, just early out

### DIFF
--- a/src/native/event_stream_rpc_server.c
+++ b/src/native/event_stream_rpc_server.c
@@ -396,6 +396,7 @@ static int s_on_new_connection_fn(
         goto error;
     }
 
+    /* Nothing can fail after here */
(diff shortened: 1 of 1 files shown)


bretambrose (commented):
Wouldn't it be simpler to just check callback data for null?

bretambrose (approved):

DmitriyMusatkin (approved):

PR #521, opened 2022-08-15
Adds support for Anonymous Credentials

waahm7:
*Issue #, if available:* https://github.com/aws/aws-sdk-java-v2/issues/3174

*Description of changes:*

- Adds support for Anonymous Credentials. 
- Makes default Credentials constructor private to avoid accidently creating anonymous credentials.

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/auth/credentials/Credentials.java
+++ b/src/main/java/software/amazon/awssdk/crt/auth/credentials/Credentials.java
@@ -33,6 +33,7 @@ public class Credentials {
 
     /**
      * Anonymous Credentials constructor. Use Anonymous Credentials when you want to skip signing.
+     * @return Anonymous Credentials
      */
     public static Credentials createAnonymousCredentials(){
         return new Credentials();
--- a/src/main/java/software/amazon/awssdk/crt/auth/credentials/Credentials.java
+++ b/src/main/java/software/amazon/awssdk/crt/auth/credentials/Credentials.java
@@ -14,7 +14,10 @@ public class Credentials {
     private byte[] secretAccessKey;
     private byte[] sessionToken;
 
-    private Credentials() {}
+    /**
+     * Anonymous Credentials Constructor. The recommended approach is to use createAnonymousCredentials() instead.
+     */
+    public Credentials() {}
 
     /**
      * @param accessKeyId - access key id to use
--- a/src/main/java/software/amazon/awssdk/crt/auth/credentials/Credentials.java
+++ b/src/main/java/software/amazon/awssdk/crt/auth/credentials/Credentials.java
@@ -15,7 +15,8 @@ public class Credentials {
     private byte[] sessionToken;
 
     /**
-     * Anonymous Credentials Constructor. The recommended approach is to use createAnonymousCredentials() instead.
+     * Anonymous Credentials constructor. Use Anonymous Credentials when you want to skip signing.
+     * The recommended approach is to use createAnonymousCredentials().
      */
     public Credentials() {}
 
--- a/src/main/java/software/amazon/awssdk/crt/auth/credentials/Credentials.java
+++ b/src/main/java/software/amazon/awssdk/crt/auth/credentials/Credentials.java
@@ -14,10 +14,7 @@ public class Credentials {
     private byte[] secretAccessKey;
     private byte[] sessionToken;
 
-    /**
-     * Default constructor
-     */
-    public Credentials() {}
+    private Credentials() {}
 
     /**
      * @param accessKeyId - access key id to use
@@ -25,9 +22,20 @@ public class Credentials {
      * @param sessionToken - (optional) session token to use
      */
     public Credentials(byte[] accessKeyId, byte[] secretAccessKey, byte[] sessionToken) {
+        if (accessKeyId == null || secretAccessKey == null) {
+            throw new IllegalArgumentException("Credentials - accessKeyId and secretAccessKey must be non null");
+        }
+
         this.accessKeyId = accessKeyId;
         this.secretAccessKey = secretAccessKey;
         this.sessionToken = sessionToken;
+    }
+
+    /**
+     * Anonymous Credentials constructor. Use Anonymous Credentials when you want to skip signing.
+     */
+    public static Credentials createAnonymousCredentials(){
+        return new Credentials();
     }
 
     /**
--- a/src/native/aws_signing.c
+++ b/src/native/aws_signing.c
@@ -154,6 +154,7 @@ static void s_aws_complete_signing_result(
     struct aws_string *signature = NULL;
     aws_signing_result_get_property(result, g_aws_signature_property_name, &signature);
 
+    /* Signature will be NULL for anonymous credentials. */
     if (signature != NULL) {
         struct aws_byte_cursor signature_cursor = aws_byte_cursor_from_string(signature);
         java_signature = aws_jni_byte_array_from_cursor(env, &signature_cursor);
--- a/src/native/aws_signing.c
+++ b/src/native/aws_signing.c
@@ -154,7 +154,7 @@ static void s_aws_complete_signing_result(
     struct aws_string *signature = NULL;
     aws_signing_result_get_property(result, g_aws_signature_property_name, &signature);
 
-    /* Signature will be NULL for anonymous credentials. */
+    /* Anonymous requests don't have a signature because they are not signed. */
     if (signature != NULL) {
         struct aws_byte_cursor signature_cursor = aws_byte_cursor_from_string(signature);
         java_signature = aws_jni_byte_array_from_cursor(env, &signature_cursor);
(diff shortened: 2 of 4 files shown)


src/main/java/software/amazon/awssdk/crt/auth/credentials/Credentials.java
```diff
@@ -14,22 +14,31 @@ public class Credentials {
     private byte[] secretAccessKey;
     private byte[] sessionToken;
 
-    /**
-     * Default constructor
-     */
-    public Credentials() {}
+    private Credentials() {}
```
DmitriyMusatkin, line 17: are you sure SDK is not using this currently and making this private will not break build for them?

waahm7: I have tried searching it in their code and didn't find any usage. It doesn't break any of their test either.

TingDaoK (commented):
Do we want the anonymous credentials provider to be in the bindings? If not, why do we add the anonymous credentials provider in C?

src/main/java/software/amazon/awssdk/crt/auth/credentials/Credentials.java
waahm7: Both solutions seem fine to me. As it wouldn't have worked before anyhow, it will just replace a runtime error with compile time if someone was using it. @graebm What do you think?

waahm7:
> Do we want the anonymous credentials provider to be in the bindings? If not, why do we add the anonymous credentials provider in C?

I am not sure, we can create an anonymous credentials provider in Java binding which can map to their AnonymousCredentialProvider. The reason we added the anonymous credentials provider in C was to make it possible to pass anonymous credentials in a provider as the static provider implementation was not allowing us to do it easily.

src/main/java/software/amazon/awssdk/crt/auth/credentials/Credentials.java
graebm: Changing it to private is an API-breaking change. Don't do that. Like Dmitriy said, we're a public library and don't know what ALL customers' code looks like.

Keep it public. But in the documentation, encourage people to use `createAnonymousCredentials()` instead since it's more clear

src/native/aws_signing.c
```diff
@@ -154,8 +154,10 @@ static void s_aws_complete_signing_result(
     struct aws_string *signature = NULL;
     aws_signing_result_get_property(result, g_aws_signature_property_name, &signature);
 
-    struct aws_byte_cursor signature_cursor = aws_byte_cursor_from_string(signature);
-    java_signature = aws_jni_byte_array_from_cursor(env, &signature_cursor);
+    if (signature != NULL) {
```
graebm, line 157: add comment that signature will be null if anonymous credentials were used

src/test/java/software/amazon/awssdk/crt/test/SigningTest.java
```diff
@@ -271,6 +274,53 @@ public Credentials getCredentials() {
...
+                config.setTime(System.currentTimeMillis());
+                config.setUseDoubleUriEncode(true);
+                config.setShouldNormalizeUriPath(true);
+                config.setSignedBodyValue(AwsSigningConfig.AwsSignedBodyValue.EMPTY_SHA256);
+                config.setCredentials(Credentials.createAnonymousCredentials());
+
+                CompletableFuture<HttpRequest> result = AwsSigner.signRequest(request, config);
+                HttpRequest signedRequest = result.get();
+                assertNotNull(signedRequest);
+                assertFalse(hasHeader(signedRequest, "Authorization"));
```
graebm, line 294: trivial: assert that the original request and the signed request have the same number of headers

src/test/java/software/amazon/awssdk/crt/test/SigningTest.java
```diff
@@ -271,6 +274,53 @@ public Credentials getCredentials() {
...
+            }
+    }
+
+    @Test(expected = CrtRuntimeException.class)
+    public void testSigningNullFail() throws Exception {
```
graebm, line 299: add a comment explaining what this is testing ... I can't figure it out at first glance

src/test/java/software/amazon/awssdk/crt/test/SigningTest.java
```diff
@@ -271,6 +274,53 @@ public Credentials getCredentials() {
...
+            config.setService("service");
+            config.setTime(System.currentTimeMillis());
+            config.setUseDoubleUriEncode(true);
+            config.setShouldNormalizeUriPath(true);
+            config.setSignedBodyValue(AwsSigningConfig.AwsSignedBodyValue.EMPTY_SHA256);
+
+            CompletableFuture<HttpRequest> result = AwsSigner.signRequest(request, config);
+            HttpRequest signedRequest = result.get();
+            assertNotNull(signedRequest);
+            assertFalse(hasHeader(signedRequest, "Authorization"));
```
graebm, line 315: Are you trying to assert that an exception IS thrown?
Because this test will pass when no exception is raised

src/test/java/software/amazon/awssdk/crt/test/SigningTest.java
```diff
@@ -271,6 +274,53 @@ public Credentials getCredentials() {
...
+        } catch (ExecutionException e){
+            assertEquals(e.getCause().getClass(), CrtRuntimeException.class);
+            CrtRuntimeException crtRuntimeException = (CrtRuntimeException) e.getCause();
+            assertEquals(crtRuntimeException.errorCode, 6149);
```
graebm, line 319: assert on the `.errorName` instead of the `.errorCode`
we realize that CrtRuntimeException isn't very java-esque (it uses error codes instead of subclasses). But checking the integer value is particularly meaningless to someone just reading this code

src/test/java/software/amazon/awssdk/crt/test/SigningTest.java
waahm7: It won't pass because of the expected exception in test annotation. I have added some comments to make it more clear. Please take a look and let me know if it is still confusing. Thanks.

src/test/java/software/amazon/awssdk/crt/test/SigningTest.java
waahm7: I have added a comment and tried to improve the naming.

graebm (approved):

src/main/java/software/amazon/awssdk/crt/auth/credentials/Credentials.java
```diff
@@ -15,7 +15,8 @@ public class Credentials {
     private byte[] sessionToken;
 
     /**
-     * Default constructor
```
DmitriyMusatkin, line 18: nit: this comment sounds a bit awkward. consider rewriting.
Something similar to following? And also consider adding comment to regular constructor.
Constructs Anonymous Credentials. 
Signer will skip signing request if Anonymous Credentials are passed in.
This method is deprecated and left for backwards compatibility. Prefer createAnonymousCredentials instead.

src/native/aws_signing.c
```diff
@@ -154,8 +154,11 @@ static void s_aws_complete_signing_result(
     struct aws_string *signature = NULL;
     aws_signing_result_get_property(result, g_aws_signature_property_name, &signature);
 
-    struct aws_byte_cursor signature_cursor = aws_byte_cursor_from_string(signature);
-    java_signature = aws_jni_byte_array_from_cursor(env, &signature_cursor);
+    /* Signature will be NULL for anonymous credentials. */
```
DmitriyMusatkin, line 157: Nit: for comments prefer saying why something is happening instead of static what is happening

src/test/java/software/amazon/awssdk/crt/test/SigningTest.java
```diff
@@ -272,6 +275,61 @@ public Credentials getCredentials() {
...
+            config.setRegion("us-east-1");
+            config.setService("service");
+            config.setTime(System.currentTimeMillis());
+            config.setUseDoubleUriEncode(true);
+            config.setShouldNormalizeUriPath(true);
+            config.setSignedBodyValue(AwsSigningConfig.AwsSignedBodyValue.EMPTY_SHA256);
+
+            CompletableFuture<HttpRequest> result = AwsSigner.signRequest(request, config);
+            HttpRequest signedRequest = result.get(); //This should throw an exception
+        } catch (ExecutionException e){
```
DmitriyMusatkin, line 324: minor: im not a big fan of rethrowing exception in the test to make sure its matches the type. 
what about using whenComplete and checking if it returns exception or result?

src/native/credentials.c
```diff
@@ -32,6 +32,10 @@ struct aws_credentials *aws_credentials_new_from_java_credentials(JNIEnv *env, j
     jbyteArray session_token =
         (*env)->GetObjectField(env, java_credentials, credentials_properties.session_token_field_id);
 
+    if (access_key_id == NULL && secret_access_key == NULL) {
```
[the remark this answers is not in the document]
waahm7: Is there any other way to create credentials apart from using Credentials.java? If not, I think we just remove the or null check because partial nulls are not possible.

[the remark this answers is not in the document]
DmitriyMusatkin: external projects might be using this directly. 
at the very least i think we should error out for defensive reasons to prevent bugs

DmitriyMusatkin (approved):
address comments and ship

PR #522, opened 2022-08-22
Use latest versions of S2N-TLS and AWS-LC

graebm:
* Update all submodules, including S2N-TLS and AWS-LC which we hadn't updated for a while due to issues which are now resolved (see https://github.com/awslabs/aws-c-common/pull/925).
* Fix `pkcs11_test_setup.py`. These tests hadn't run in months 😬


TwistedTwigleg (approved):
👍 

I'm not sure why LGTM is finding Javascript issues though. I don't think we have Javascript in the Java CRT...

PR #526, opened 2022-08-24
Fix credential provider delegate for anonymous credentials

waahm7:

### DIFF
--- a/src/native/credentials_provider.c
+++ b/src/native/credentials_provider.c
@@ -557,9 +557,22 @@ static int s_credentials_provider_delegate_get_credentials(
     jbyteArray java_session_token =
         (*env)->GetObjectField(env, java_credentials, credentials_properties.session_token_field_id);
 
+    /**
+     * Construct Anonymous Credentials.
+     * Anonymous Credentials are used to skip signing.
+     */
+    if (java_access_key_id == NULL && java_secret_access_key == NULL) {
+        struct aws_credentials *native_credentials = aws_credentials_new_anonymous(allocator);
+        callback(native_credentials, AWS_ERROR_SUCCESS, callback_user_data);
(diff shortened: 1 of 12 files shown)


src/native/credentials_provider.c
```diff
@@ -557,9 +557,22 @@ static int s_credentials_provider_delegate_get_credentials(
     jbyteArray java_session_token =
         (*env)->GetObjectField(env, java_credentials, credentials_properties.session_token_field_id);
 
+    /**
+     * Construct Anonymous Credentials.
+     * Anonymous Credentials are used to skip signing.
+     */
+    if (java_access_key_id == NULL && java_secret_access_key == NULL) {
```
TingDaoK, line 564: what if someone created a credentials that have access key and secret access key as null but token is non-null?

waahm7: It is not possible because credentials are immutable and constructors do not allow that.

TingDaoK: `return Credentials(null, null, "something")`?

waahm7: It will fail at https://github.com/awslabs/aws-crt-java/blob/main/src/main/java/software/amazon/awssdk/crt/auth/credentials/Credentials.java#L29

waahm7: In C, aws_credentials_new will return null as well and not an object with has null,null,"something".

TingDaoK: oh, I see. My local repo is out of date.

TingDaoK (approved):

src/native/credentials_provider.c
TingDaoK: well, here, it will not go through `aws_credentials_new` at all.

PR #528, opened 2022-09-08
Expose Additional Configuration for S3ClientConfig

waahm7:
*Description of changes:*
- Exposes ProxyOptions, EnvironmentVariableProxySetting, ConnectTimeOutMs, S3TcpKeepAliveOptions, and monitoring options for the S3Client. 


src/main/java/software/amazon/awssdk/crt/http/HttpProxyEnvironmentVariableSetting.java
```diff
@@ -0,0 +1,103 @@
...
+        private int environmentVariableType;
+
+        HttpProxyEnvironmentVariableType(int val) {
+            environmentVariableType = val;
+        }
+
+        public int getValue() {
+            return environmentVariableType;
+        }
+    }
```
waahm7, line 50: This enum name matches the name in the C code. However, I think it is confusing and am trying to think of a better name.

src/main/java/software/amazon/awssdk/crt/http/HttpProxyEnvironmentVariableSetting.java
```diff
@@ -0,0 +1,103 @@
...
+import software.amazon.awssdk.crt.io.TlsContext;
+import software.amazon.awssdk.crt.http.HttpProxyOptions.HttpProxyConnectionType;
+
+/**
+ * This class provides access to Http proxy environment variable configuration
+ * setting
+ */
+public class HttpProxyEnvironmentVariableSetting {
+
+    private TlsContext tlsContext;
```
TingDaoK, line 17: Why `TlsContext` instead of `TlsConnectionOptions` as what we have in C?

src/main/java/software/amazon/awssdk/crt/s3/S3TcpKeepAliveOptions.java
```diff
@@ -0,0 +1,40 @@
+package software.amazon.awssdk.crt.s3;
+
+/**
+ * This class provides access to setting Tcp Keep Alive Options.
+ * If interval or timeout are zero, then default values are used.
+ */
+public class S3TcpKeepAliveOptions {
```
TingDaoK, line 7: Does this really need to be a separate class?

src/native/http_connection_manager.h
```diff
@@ -6,6 +6,7 @@
 #ifndef AWS_JNI_CRT_HTTP_CONNECTION_MANAGER_H
 #define AWS_JNI_CRT_HTTP_CONNECTION_MANAGER_H
 
+#include <http_proxy_options.h>
```
TingDaoK, line 9: trivial: move this include to `http_connection_manager.c`, it doesn't need to be in the header file.

src/main/java/software/amazon/awssdk/crt/s3/S3TcpKeepAliveOptions.java
TingDaoK: There is a `SocketOptions` that has the tcp keep alive settings. This `S3TcpKeepAliveOptions` seems like a unnecessary complexity to me, it's not doing anything special to S3. 

I don't believe we need this in C [here](https://github.com/awslabs/aws-c-s3/pull/204/files#diff-e45799b2b3d8b2c5e2bba0698cb763ed106c53e07f3c5c049f4b2eee58c26382R155) as well. I'd use socket options instead.

waahm7: SocketOptions has additional properties that we don't want to allow override.

src/main/java/software/amazon/awssdk/crt/http/HttpProxyEnvironmentVariableSetting.java
waahm7: Thanks, fixed.

src/main/java/software/amazon/awssdk/crt/s3/S3Client.java
```diff
@@ -24,8 +28,45 @@ public class S3Client extends CrtResource {
...
+            environmentVariableType = environmentVariableSetting.getEnvironmentVariableType().getValue();
+        }
+
+        HttpMonitoringOptions monitoringOptions = options.getMonitoringOptions();
+        long monitoringThroughputThresholdInBytesPerSecond = 0;
+        int monitoringFailureIntervalInSeconds = 0;
+        if (monitoringOptions != null) {
+            monitoringThroughputThresholdInBytesPerSecond = monitoringOptions.getMinThroughputBytesPerSecond();
+            monitoringFailureIntervalInSeconds = monitoringOptions.getAllowableThroughputFailureIntervalSeconds();
+        }
```
TingDaoK, line 67: Trivial: Why not create something like `aws_http_proxy_options_from_java` instead?

TingDaoK (approved):
fix&ship

src/main/java/software/amazon/awssdk/crt/s3/S3ClientOptions.java
```diff
@@ -35,6 +38,42 @@ public class S3ClientOptions {
...
+     * Optional.
+     * Proxy configuration for http connection.
+     */
+    private HttpProxyOptions proxyOptions;
+
+    /**
+     * Optional.
+     * Configuration for fetching proxy configuration from environment.
+     * By Default read proxy configuration from environment is enabled.
+     * Only works when proxy_options is not set. If both are set, configuration from
```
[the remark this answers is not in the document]
waahm7: Thanks, good catch.

src/main/java/software/amazon/awssdk/crt/s3/S3Client.java
waahm7: I found examples of both ways in our code and this felt simpler instead of trying parse each value in the C code. It is also consistent with ConnectionManager and StreamManager.

PR #530, opened 2022-09-20
Http stream binding fix

TingDaoK:
*Issue #, if available:*
- We destroy the HTTP stream jni binding from complete callback of the stream. However the Java Stream object is still alive from Java, and when it gets GCed from Java, it will access the jni binding again and lead to a crash.

*Description of changes:*
- Refcount the binidng as it needs to be kept alive for both the java object and the native stream callbacks.

~TODO: That's not the only structure that has the issue. We need to go through the whole code base and check for any place that has the similar issue.~

### DIFF
--- a/src/native/http2_stream_manager.c
+++ b/src/native/http2_stream_manager.c
@@ -300,7 +300,7 @@ static void s_on_stream_acquired(struct aws_http_stream *stream, int error_code,
         (*env)->CallVoidMethod(
             env, callback_data->java_async_callback, async_callback_properties.on_failure, crt_exception);
         (*env)->DeleteLocalRef(env, crt_exception);
-        aws_http_stream_binding_destroy(env, callback_data->stream_binding);
+        aws_http_stream_binding_release(env, callback_data->stream_binding);
     } else {
         callback_data->stream_binding->native_stream = stream;
         jobject j_http_stream =
@@ -312,10 +312,10 @@ static void s_on_stream_acquired(struct aws_http_stream *stream, int error_code,
                 env, callback_data->java_async_callback, async_callback_properties.on_failure, crt_exception);
             (*env)->DeleteLocalRef(env, crt_exception);
             (*env)->ExceptionClear(env);
-            aws_http_stream_binding_destroy(env, callback_data->stream_binding);
+            aws_http_stream_binding_release(env, callback_data->stream_binding);
         } else {
-            /* Stream is activated once we acquired from the Stream Manager */
-            aws_atomic_store_int(&callback_data->stream_binding->activated, 1);
+            /* Stream is activated once we acquired from the Stream Manager. Hold the refcount on the binding */
+            aws_http_stream_binding_acquire(callback_data->stream_binding);
             callback_data->stream_binding->java_http_stream_base = (*env)->NewGlobalRef(env, j_http_stream);
             (*env)->CallVoidMethod(
                 env,
@@ -358,7 +358,7 @@ JNIEXPORT void JNICALL Java_software_amazon_awssdk_crt_http_Http2StreamManager_h
         return;
     }
 
-    struct http_stream_binding *stream_binding = aws_http_stream_binding_alloc(env, jni_http_response_callback_handler);
+    struct http_stream_binding *stream_binding = aws_http_stream_binding_new(env, jni_http_response_callback_handler);
     if (!stream_binding) {
         /* Exception already thrown */
         return;
@@ -368,7 +368,7 @@ JNIEXPORT void JNICALL Java_software_amazon_awssdk_crt_http_Http2StreamManager_h
         aws_http_request_new_from_java_http_request(env, marshalled_request, jni_http_request_body_stream);
     if (stream_binding->native_request == NULL) {
         /* Exception already thrown */
-        aws_http_stream_binding_destroy(env, stream_binding);
+        aws_http_stream_binding_release(env, stream_binding);
         return;
     }
 
--- a/src/native/http2_stream_manager.c
+++ b/src/native/http2_stream_manager.c
@@ -314,7 +314,7 @@ static void s_on_stream_acquired(struct aws_http_stream *stream, int error_code,
             (*env)->ExceptionClear(env);
             aws_http_stream_binding_release(env, callback_data->stream_binding);
         } else {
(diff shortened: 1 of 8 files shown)


src/native/http_request_response.c
```diff
@@ -62,50 +62,65 @@ void aws_java_http_stream_from_native_delete(JNIEnv *env, jobject jHttpStream) {
...
 
-    callback->java_http_response_stream_handler = (*env)->NewGlobalRef(env, java_callback_handler);
-    AWS_FATAL_ASSERT(callback->java_http_response_stream_handler);
-    AWS_FATAL_ASSERT(!aws_byte_buf_init(&callback->headers_buf, allocator, 1024));
+    binding->java_http_response_stream_handler = (*env)->NewGlobalRef(env, java_callback_handler);
+    AWS_FATAL_ASSERT(binding->java_http_response_stream_handler);
+    AWS_FATAL_ASSERT(!aws_byte_buf_init(&binding->headers_buf, allocator, 1024));
 
-    aws_atomic_init_int(&callback->activated, 0);
+    aws_atomic_init_int(&binding->ref, 1);
```
graebm, line 121: just curious, why not use aws_ref_count?

src/native/http_request_response.c
```diff
@@ -534,25 +550,27 @@ JNIEXPORT void JNICALL Java_software_amazon_awssdk_crt_http_HttpStreamBase_httpS
...
-        aws_atomic_store_int(&cb_data->activated, 0);
-        (*env)->DeleteGlobalRef(env, cb_data->java_http_stream_base);
+        (*env)->DeleteGlobalRef(env, binding->java_http_stream_base);
         aws_jni_throw_runtime_exception(
             env, "HttpStream activate failed with error %s\n", aws_error_str(aws_last_error()));
+    } else {
+        /* Succeed activating the stream, holding the binding to make sure it lives across all the callbacks until
+         * stream completes */
+        aws_http_stream_binding_acquire(binding);
     }
```
graebm, line 562: we should only +1 the refcount if the stream is changing from not-activated -> activated

Nothing is stopping java users from calling `stream.activate()` multiple times.

except I don't think there's currently any way to detect whether or not the stream is activated based on the public C API. `aws_http_stream_activate()` will return AWS_OP_SUCCESS if it's called multiple times. We thought that would nice for people :|

src/native/http_request_response.c
```diff
@@ -116,17 +131,17 @@ int aws_java_http_stream_on_incoming_headers_fn(
...
+    struct http_stream_binding *binding = (struct http_stream_binding *)user_data;
     int resp_status = -1;
     int err_code = aws_http_stream_get_incoming_response_status(stream, &resp_status);
     if (err_code != AWS_OP_SUCCESS) {
         AWS_LOGF_ERROR(AWS_LS_HTTP_STREAM, "id=%p: Invalid Incoming Response Status", (void *)stream);
         return AWS_OP_ERR;
     }
 
-    callback->response_status = resp_status;
+    binding->response_status = resp_status;
```
graebm, line 142: thank you for renaming all of these
I'm sure there's some historical reason these were all named `callback`, but that should have changed long ago

src/native/http_request_response.c
TingDaoK: The release needs to take `JNIEnv *env`, and I don't think we can save that in the structure?

src/native/http_request_response.c
TingDaoK: Good Catch!

Options:
- Add an activated check function in C, I believe the python binding has the exact same issue, it can use that function as well. [here](https://github.com/awslabs/aws-crt-python/blob/main/source/http_stream.c#L315)
- From the language level to skip calling activate in native code multiple times, which may benefit the performance?? 

Prefer the first one as we will need to handle the case for all the language bindings and it's easily to miss, I think

graebm: Option 3: shutdown_complete callback in C, and the bindings can use that to know when to cleanup

src/native/http_request_response.c
```diff
@@ -332,60 +356,63 @@ static jobject s_make_request_general(
...
     } else if (!jHttpStreamBase) {
-        /* Failed to create java HttpStream, but did create native aws_http_stream.
-          Close connection and mark native_stream for release.
-          callback_data will clean itself up when stream completes. */
-        aws_http_connection_close(native_conn);
-        aws_http_stream_release(callback_data->native_stream);
+        /* Failed to create java HttpStream, but did create native aws_http_stream. As we never activate the native
+         * stream, we can just release the native stream, and release the binding. */
+        aws_http_stream_release(stream_binding->native_stream);
+        aws_http_stream_binding_release(env, stream_binding);
```
graebm: don't think this is right anymore.
before we run these 2 lines:
- stream_binding has refcount 1
- native_stream has refcount 1

then we call `aws_http_stream_release(stream_binding->native_stream);`
- stream_binding refcount goes to 0
    -  destroy callback fires...
    - native_stream refcount goes to 0

then we call `aws_http_stream_binding_release(env, stream_binding);`
kablooie, the stream_binding was already cleaned up

graebm: the error handling in this function ... is just so confusing

man, we've gotten better at error handling over the years

this is bad old days stuff

graebm: I pushed a cleanup of this function. I was one of the early authors. I feel some responsibility for this mess

src/native/http2_stream_manager.c
```diff
@@ -312,10 +312,10 @@ static void s_on_stream_acquired(struct aws_http_stream *stream, int error_code,
                 env, callback_data->java_async_callback, async_callback_properties.on_failure, crt_exception);
             (*env)->DeleteLocalRef(env, crt_exception);
             (*env)->ExceptionClear(env);
-            aws_http_stream_binding_destroy(env, callback_data->stream_binding);
+            aws_http_stream_binding_release(env, callback_data->stream_binding);
         } else {
-            /* Stream is activated once we acquired from the Stream Manager */
-            aws_atomic_store_int(&callback_data->stream_binding->activated, 1);
+            /* Acquire for the native stream */
+            aws_http_stream_binding_acquire(callback_data->stream_binding);
```
graebm, line 318: don't we need to do binding_acquire() as long as `native_stream` exists? because the destroy_callback will decref the binding?

graebm (approved):

PR #531, opened 2022-09-21
Cognito credentials provider

bretambrose:
* Cognito credentials provider binding

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/auth/credentials/CognitoCredentialsProvider.java
+++ b/src/main/java/software/amazon/awssdk/crt/auth/credentials/CognitoCredentialsProvider.java
@@ -220,11 +220,27 @@ public class CognitoCredentialsProvider extends CredentialsProvider {
         addReferenceTo(tlsContext);
     }
 
+    private void writeLengthPrefixedBytesSafe(ByteBuffer buffer, byte[] bytes) {
+        if (bytes != null) {
+            buffer.putInt(bytes.length);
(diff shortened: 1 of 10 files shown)


src/main/java/software/amazon/awssdk/crt/auth/credentials/CognitoCredentialsProvider.java
```diff
@@ -0,0 +1,265 @@
...
+        acquireNativeHandle(nativeHandle);
+        addReferenceTo(clientBootstrap);
+        addReferenceTo(tlsContext);
+    }
+
+    private byte[] marshalLoginsForJni(ArrayList<CognitoLoginTokenPair> logins) {
+        int size = 0;
+
+        for (CognitoLoginTokenPair login : logins) {
+            size += login.identityProviderName.length + login.identityProviderToken.length + (BUFFER_INT_SIZE * 2);
```
TingDaoK, line 227: Can `identityProviderName` be null? If it's null, the exception will be raised here.

TingDaoK (approved):

src/main/java/software/amazon/awssdk/crt/auth/credentials/CognitoCredentialsProvider.java
bretambrose: I don't think name can be null but it might be possible that token might be in an unauthenticated setting.  So I'll add protections around both anyways.

PR #543, opened 2022-10-21
Fix inaccurate docs for onReponseBody

graebm:
**Issue:**
The docs for `onResponseBody` told users that they were required to copy the bytes immediately, which is not true anymore.

**Description of Changes:**
Correct docs, and add a bunch more about flow-control aka read-backpressure aka window-management

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/http/HttpClientConnectionManagerOptions.java
+++ b/src/main/java/software/amazon/awssdk/crt/http/HttpClientConnectionManagerOptions.java
@@ -106,9 +106,12 @@ public class HttpClientConnectionManagerOptions {
     public TlsConnectionOptions getTlsConnectionOptions() { return tlsConnectionOptions; }
 
     /**
-     * Sets the IO channel window size to use for connections in the connection pool
-     * @param windowSize The initial window size to use for each connection
+     * Sets the starting size of each HTTP stream's flow-control window.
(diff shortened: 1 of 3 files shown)


waahm7 (approved):

src/main/java/software/amazon/awssdk/crt/http/HttpClientConnectionManagerOptions.java
```diff
@@ -198,23 +201,25 @@ public HttpClientConnectionManagerOptions withProxyOptions(HttpProxyOptions prox
...
-     * If this is enabled, you must call HttpStream.UpdateWindow() for every
-     * byte read from the OnIncomingBody callback.
      * @return true if manual window management is used, false otherwise
+     * @see #withManualWindowManagement
      */
     public boolean isManualWindowManagement() { return manualWindowManagement; }
 
     /**
-     * If set to true, then the TCP read back pressure mechanism will be enabled. You should
+     * If set to true, then you must manage the TCP read backpressure mechanism. You should
```
TingDaoK, line 210: `the TCP read backpressure` is that still true? we actually have the connection manager support both h2 and h1.

graebm: good point. I'll edit

TingDaoK (approved):

PR #544, opened 2022-10-24
Musl testing

bretambrose:
* Add an additional native artifact directory component for C runtime
* Add Musl detection in build (CMakeLists.txt) and at runtime
* Add Musl-based CI; <del>remove old cross-compiled arm CI in favor of the Musl-based arm jobs. </del> Keep both as cross-compiled are used with Glibc.
* Add environment/system property override for architecture selection to support ambiguous (arm6 vs arm7) cases
* Tweak some tests to work in a Musl-based environment


.builder/actions/aws_crt_java_test.py
```diff
@@ -22,20 +23,25 @@ def run(self, env):
...
         with self._write_secret_to_temp_file(env, "unit-test/rootca") as root_ca_file, self._write_secret_to_temp_file(env, "unit-test/certificate") as cert_file, \
         self._write_secret_to_temp_file(env, "unit-test/privatekey") as key_file, self._write_secret_to_temp_file(env, "ecc-test/certificate") as ecc_cert_file, \
         self._write_secret_to_temp_file(env, "ecc-test/privatekey") as ecc_key_file:
 
-            test_command = "mvn -P continuous-integration -B test -DredirectTestOutputToFile=true -DreuseForks=false " \
+            test_command = "mvn -P continuous-integration {} -B test -DredirectTestOutputToFile=true -DreuseForks=false " \
                 "-DrerunFailingTestsCount=5 -Daws.crt.memory.tracing=2 -Daws.crt.debugnative=true -Daws.crt.ci=true " \
-                "-Dendpoint={} -Dcertificate={} -Dprivatekey={} -Drootca={} -Decc_certificate={} -Decc_privatekey={}".format(endpoint, \
+                "-Dendpoint={} -Dcertificate={} -Dprivatekey={} -Drootca={} -Decc_certificate={} -Decc_privatekey={}".format(additional_profiles, endpoint, \
                 cert_file.name, key_file.name, root_ca_file.name, ecc_cert_file.name, ecc_key_file.name)
```
graebm, line 38: trivial: use fstrings instead of .format() (they didn't exist until python 3.6, but the world is python 3.7+ these days)

~"{} -B test".format(additional_profiles)~ -> ` f"{additional_profiles} -B test"`

.builder/actions/aws_crt_java_test.py
```diff
@@ -22,20 +23,25 @@ def run(self, env):
...
 
             all_test_result = os.system(test_command)
 
         env.shell.setenv('AWS_CRT_SHUTDOWN_TESTING', '1')
-        shutdown_test_result = os.system("mvn -P continuous-integration -B test -DredirectTestOutputToFile=true -DreuseForks=false \
-            -Daws.crt.memory.tracing=2 -Daws.crt.debugnative=true -Dtest=ShutdownTest")
+        shutdown_test_result = os.system("mvn -P continuous-integration {} -B test -DredirectTestOutputToFile=true -DreuseForks=false \
+            -Daws.crt.memory.tracing=2 -Daws.crt.debugnative=true -Dtest=ShutdownTest".format(additional_profiles))
```
graebm, line 44: trivial: fstrings pls

.github/workflows/ci.yml
```diff
@@ -80,18 +80,46 @@ jobs:
...
+    steps:
+      - name: Install qemu/docker
+        run: docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
+      - name: Checkout Sources
+        uses: actions/checkout@v2
+        with:
+          submodules: true
+      - name: Build ${{ env.PACKAGE_NAME }}
+        run: |
+          export AWS_CRT_ARCH=armv7
```
graebm, line 120: just curious, how does this environment variable get all the way into docker ?

CMakeLists.txt
```diff
@@ -51,6 +51,26 @@ if (ANDROID)
     set(AWS_LIBRARY_OUTPUT_DIR "${CMAKE_BINARY_DIR}/../../../lib")
 endif()
 
+if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
```
graebm, line 54: trivial: add comment at start of this if/else block like:
`# Detect C Runtime`

src/test/java/software/amazon/awssdk/crt/test/TlsContextOptionsTest.java
```diff
@@ -224,7 +224,7 @@ public void testMtlsPkcs11() {
         Assume.assumeTrue(System.getProperty("NETWORK_TESTS_DISABLED") == null);
         Pkcs11LibTest.assumeEnvironmentSetUpForPkcs11Tests();
 
-        try (Pkcs11Lib pkcs11Lib = new Pkcs11Lib(Pkcs11LibTest.TEST_PKCS11_LIB);
+        try (Pkcs11Lib pkcs11Lib = new Pkcs11Lib(Pkcs11LibTest.TEST_PKCS11_LIB, Pkcs11Lib.InitializeFinalizeBehavior.STRICT);
```
graebm, line 227: Add a comment here about how softhsm (give version if you can) was crashing on platform (give exact config if you can) on shutdown if it wasn't finalized, so we can only test the STRICT config

and if Musl only allows a library to be unloaded once, mention that too, because it's really going to bite someone in the butt if they try to add more tests

graebm (approved):
many trivial comments and shipit

AMAZING WORK!!!

.github/workflows/ci.yml
bretambrose: It didn't, but after the latest builder PR (https://github.com/awslabs/aws-crt-builder/pull/203) it should.  Oops.

.github/workflows/ci.yml
```diff
@@ -7,7 +7,7 @@ on:
       - 'docs'
 
 env:
-  BUILDER_VERSION: v0.9.32
+  BUILDER_VERSION: v0.9.31
```
waahm7, line 10: Trivial: builder version should be 0.9.32.

bretambrose: That was intentional.  The 0.9.32 release did not complete container builds successfully.  Is there something that requires 0.9.32?

waahm7: It contains the following fix https://github.com/awslabs/aws-crt-builder/pull/214. @graebm recently updated the CI in https://github.com/awslabs/aws-crt-java/pull/577.

graebm (approved):
fix & shipit

pom.xml
```diff
@@ -141,6 +142,13 @@
                 <cmake.buildtype>Debug</cmake.buildtype>
             </properties>
         </profile>
+        <profile>
+            <id>alpine</id>
+            <properties>
+                <crt.surefire.version>3.0.0-M4</crt.surefire.version>
```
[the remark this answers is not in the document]
waahm7: We don't need it and I have removed this and the alpine profile.

.builder/actions/build-classifier.py
```diff
@@ -0,0 +1,21 @@
+
+import Builder
+import sys
+import os
+import tempfile
+from builder.core.host import current_host
+
+class BuildClassifier(Builder.Action):
```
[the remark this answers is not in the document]
waahm7: Thanks, I have added a comment.

.builder/actions/build-classifier.py
```diff
@@ -0,0 +1,21 @@
...
+
+        actions = []
+
+        crt_classifier = os.getenv("CRT_CLASSIFIER")
+        if crt_classifier is not None and os.system(f"mvn -B install -DskipTests -Dshared-lib.skip=true -Dcrt.classifier={crt_classifier}"):
+
+            # Failed
+            actions.append("exit 1")
+
+        return Builder.Script(actions, name='build-classifier')
```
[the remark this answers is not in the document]
waahm7: Thanks, I have updated to the first suggestion.

README.md
```diff
@@ -87,6 +87,7 @@ The `aws-crt` JAR in Maven Central is a large "uber" jar that contains compiled
 - linux-aarch_64
 - linux-x86_32
 - linux-x86_64
+- linux-x86_64-musl
```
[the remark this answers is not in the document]
waahm7: Thanks, added.

waahm7 (commented):
I made the following changes
- Removes Alpine profile and different Surefire version, that is not needed.
- Fix musl build script to correctly mount the source and use that instead of pulling from Github.
- Keep previous cross compile jobs as they use glibc.
- Add CI and CD scripts for armv7 and armv8.
- Use Qemu from ECR. Docker has a pull limit for users, and I encountered issues during testing due to this limit. I will update the builder to keep this image up-to-date

README.md
```diff
@@ -82,11 +82,14 @@ The `aws-crt` JAR in Maven Central is a large "uber" jar that contains compiled
...
-- linux-armv6
-- linux-armv7
+- linux-armv6 (no auto-detect)
+- linux-armv7 (no auto-detect)
 - linux-aarch_64
 - linux-x86_32
 - linux-x86_64
+- linux-x86_64-musl (no auto-detect)
+- linux-armv7-musl (no auto-detect)
+- linux-aarch_64-musl (no auto-detect)
```
graebm, line 92: I see that the os-maven-plugin doesn't detect musl, but is there a classifier that other Java projects have settled on? I googled and stumbled across `linux_musl` being used as the platform [here](https://jreleaser.org/guide/latest/tools/jdks-maven.html), so it would be like `linux_musl-x86_64` instead of ~linux-x86_64-musl~

can you do some more research on this?

.github/workflows/ci.yml
```diff
@@ -79,6 +79,48 @@ jobs:
...
           chmod a+x builder
           ./builder build -p ${{ env.PACKAGE_NAME }} --target=linux-${{ matrix.arch }} --spec=downstream
+          
+  linux-musl:
+    runs-on: ubuntu-22.04 # latest
+    strategy:
+      matrix:
+        image:
+          - alpine-3.16-x64
+          - alpine-3.16-x86
```
graebm, line 89: debatable: should we bother testing this? we don't package it up
```suggestion
```

builder.json
```diff
@@ -78,6 +78,27 @@
...
+            },
+            "hosts": {
+                "alpine": {
+                    "!packages": [
+                        "maven",
+                        "openjdk8"
+                    ],
+                    "!build_steps": [
+                        "mvn -P continuous-integration -B compile",
+                        "build-classifier"
```
graebm, line 90: oh builder.json ... you confusing scamp

are we using `targets/hosts/alpine` to override the overrides from `target/linux/armvX`, to put stuff back to its default values so it actually runs tests ??? 🫠

if the old linux-armv6/7/64 CI jobs ACTUALLY ran their tests, all these overrides wouldn't be necessary.

but we don't have aws-crt-builder docker images for armv6/7, so it's not a simple fix

just ranting...

builder.json
```diff
@@ -78,6 +78,27 @@
                         "-DBUILD_TESTING=OFF"
                     ]
                 }
+            },
+            "hosts": {
+                "alpine": {
```
graebm, line 83: suggestion: add a comment explaining why this is necessary
```suggestion
                "_comment": "???????what is going on here?????",
                "alpine": {
```

README.md
graebm: if we do end up going with linux_musl, I wonder if it makes more sense to consider "linux_musl" its own platform everywhere else in our code, so we can continue with `platform/arch` paradigm instead of changing to `platfom/arch/cruntime` paradigm

e.g. organize the files like:
```
linux/x86_64/libaws-crt-jni.dylib
linux_musl/x86_64/libaws-crt-jni.dylib
osx/x86_64/libaws-crt-jni.dylib
```

instead of 
```
linux/x86_64/glibc/libaws-crt-jni.dylib
linux_musl/x86_64/musl/libaws-crt-jni.dylib
osx/x86_64/cruntime/libaws-crt-jni.dylib
```

codebuild/cd/alpine-arm64-build.yml
```diff
@@ -0,0 +1,22 @@
+version: 0.2
```
graebm, line 1: trivial: names these files "musl" or "linux_musl", not ~alpine~. The fact that we're building for "musl" won't change. The detail of using "alpine" might change

README.md
waahm7: I couldn't find much information on which classifiers is common. I found [tokyocabinet-linux_musl-x64/0.0.14](https://mvnrepository.com/artifact/com.fizzed/tokyocabinet-linux_musl-x64/0.0.14) which uses the `linux_musl` classifier, and [rocksdbjni](https://mvnrepository.com/artifact/org.rocksdb/rocksdbjni) which uses `linux-x86_64-musl`. Looking at the documentation for the [os-maven-plugin](https://github.com/trustin/os-maven-plugin) and [osdetector-gradle-plugin](https://github.com/google/osdetector-gradle-plugin), I think it makes more sense to use `linux-x86_64-musl`. 

In the section titled "Customized deployments for specific releases of Linux" in their documentation, different variants of Linux are classified by appending identifiers to the end, like `linux-<arch>-debian`. For Alpine Linux, the value of `${os.detected.release.like}` is `alpine`. So the users of the `osdetector-gradle-plugin` can do something like the following to get auto-detect to work for alpine.
```
def getLibCVariant() {
    if (osdetector.release != null && osdetector.release.isLike('alpine')) {
        return '-musl'
    } else {
        return ''
    }
}

dependencies {
    implementation("software.amazon.awssdk.crt:aws-crt:0.22.5:${osdetector.classifier}${getLibCVariant()}")
}
```
I am not sure if it is this simple for os-maven-plugin but users can still achieve something similar by using profiles and taking a dependency on the following if `os.detected.release.like` is `alpine`
```
 <dependencies>
	<dependency>
	    <groupId>com.example</groupId>
	    <artifactId>example-native-lib</artifactId>
	    <version>${version}</version>
	    <classifier>${os.detected.classifier}-musl</classifier>
	</dependency>
</dependencies>
```

.github/workflows/ci.yml
waahm7: Maybe it is worth keeping it in CI so that we know that we are not breaking anything for 32-bit Musl. We don't specifically package OpenWrt or Raspberry either.

graebm: those were introduced because we knew we've had customers using them, and fixed bugs, and added those CI passes as part of the bugfix. I'm not aware of anyone using 32bit musl. I'd rather not add complexity with no known beneficiaries. There's an infinite number of platforms to support if that's the case

waahm7: Thanks, makes sense. We can add it when we know there are customers using it.

builder.json
waahm7: yeah, I have added a comment.

builder.json
waahm7: Thanks, I have added a comment.

codebuild/cd/alpine-arm64-build.yml
waahm7: Thanks, I have updated the names to musl-*.

graebm (approved):

PR #552, opened 2022-11-15
Java keystore and Base64 support

TwistedTwigleg:
*Issue #, if available:*

Not a CRT issue, but there is a Java V2 SDK issue: https://github.com/aws/aws-iot-device-sdk-java-v2/issues/180

*Description of changes:*

Adds a new option for making a `TlsContextOptions` that supports taking a Java keystore as input and then pulling the certificate and private key out of it. Associated sample (including CI) is in the accompanying SDK PR: https://github.com/aws/aws-iot-device-sdk-java-v2/pull/337

Edit: Also exposes Base64 encoding and decoding to the StringUtils class. This is necessary to support Android API level 24 and still encode/decode Base64 data from the Java keystore.

________

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/io/TlsContextOptions.java
+++ b/src/main/java/software/amazon/awssdk/crt/io/TlsContextOptions.java
@@ -358,11 +358,13 @@ public final class TlsContextOptions extends CrtResource {
 
     /**
      * Helper which creates mutual TLS (mTLS) options using a certificate and private key
-     * stored in a Java keystore. Will fail if there is no certificate and key at the given certificate alias.
+     * stored in a Java keystore.
+     * Will throw an exception if there is no certificate and key at the given certificate alias, or there is some other
+     * error accessing or using the passed-in Java keystore.
      *
-     * Note: function assumes the passed keystore has already been loaded from a file by calling "keystore.load()" or similar"
+     * Note: function assumes the passed keystore has already been loaded from a file by calling "keystore.load()" or similar.
      *
-     * @param keyStore The Java keystore to use. Assumed to be loaded with certificates and keys
+     * @param keyStore The Java keystore to use. Assumed to be loaded with the desired certificate and key
      * @param certificateAlias The alias of the certificate and key to use.
      * @param certificatePassword The password of the certificate and key to use.
      * @throws CrtRuntimeException if an error occurs, like the keystore cannot be opened or the certificate is not found.
--- a/src/main/java/software/amazon/awssdk/crt/io/TlsContextOptions.java
+++ b/src/main/java/software/amazon/awssdk/crt/io/TlsContextOptions.java
@@ -356,6 +356,53 @@ public final class TlsContextOptions extends CrtResource {
         return options;
     }
 
+    /**
+     * Helper which creates mutual TLS (mTLS) options using a certificate and private key
+     * stored in a Java keystore. Will fail if there is no certificate and key at the given certificate alias.
+     *
+     * Note: function assumes the passed keystore has already been loaded from a file by calling "keystore.load()" or similar"
+     *
+     * @param keyStore The Java keystore to use. Assumed to be loaded with certificates and keys
+     * @param certificateAlias The alias of the certificate and key to use.
+     * @param certificatePassword The password of the certificate and key to use.
+     * @throws CrtRuntimeException if an error occurs, like the keystore cannot be opened or the certificate is not found.
+     * @return A set of options for setting up an mTLS connection
+     */
+    public static TlsContextOptions createWithMtlsJavaKeystore(
+        java.security.KeyStore keyStore, String certificateAlias, String certificatePassword) {
+
+        TlsContextOptions options = new TlsContextOptions();
+        String certificate;
+        try {
+            java.security.cert.Certificate certificateData = keyStore.getCertificate(certificateAlias);
+            if (certificateData == null) {
+                throw new CrtRuntimeException("Certificate at given certificate alias is empty");
+            }
+            certificate = "-----BEGIN CERTIFICATE-----\n" +
+                          java.util.Base64.getEncoder().encodeToString(certificateData.getEncoded()) +
+                          "-----END CERTIFICATE-----\n";
+        } catch (java.security.KeyStoreException | java.security.cert.CertificateEncodingException ex) {
+            throw new CrtRuntimeException("Could not get certificate from Java keystore");
+        }
+        String privateKey;
+        try {
+            java.security.Key keyData = keyStore.getKey(certificateAlias, certificatePassword.toCharArray());
+            if (keyData == null) {
+                throw new CrtRuntimeException("Private key at given certificate alias is empty");
+            }
+            privateKey = "-----BEGIN RSA PRIVATE KEY-----\n" +
+                         java.util.Base64.getEncoder().encodeToString(keyData.getEncoded()) +
+                         "-----END RSA PRIVATE KEY-----\n";
+        } catch (java.security.KeyStoreException | java.security.NoSuchAlgorithmException ex) {
+            throw new CrtRuntimeException("Could not get key from Java keystore");
+        } catch (java.security.UnrecoverableKeyException ex) {
+            throw new CrtRuntimeException("Could not get key from Java keystore due to key being unrecoverable");
+        }
+        options.initMtls(certificate, privateKey);
+        options.verifyPeer = true;
+        return options;
+    }
+
     /*******************************************************************************
      * .with() methods
      ******************************************************************************/
--- a/src/main/java/software/amazon/awssdk/crt/io/TlsContextOptions.java
+++ b/src/main/java/software/amazon/awssdk/crt/io/TlsContextOptions.java
@@ -383,7 +383,7 @@ public final class TlsContextOptions extends CrtResource {
             String certificateString = new String(StringUtils.base64Encode(certificateData.getEncoded()));
             certificate = "-----BEGIN CERTIFICATE-----\n" + certificateString + "-----END CERTIFICATE-----\n";
         } catch (java.security.KeyStoreException | java.security.cert.CertificateEncodingException ex) {
-            throw new RuntimeException(ex);
+            throw new RuntimeException("Failed to get certificate from Java keystore", ex);
         }
         String privateKey;
         try {
@@ -394,7 +394,7 @@ public final class TlsContextOptions extends CrtResource {
             String keyString = new String(StringUtils.base64Encode(keyData.getEncoded()));
             privateKey = "-----BEGIN RSA PRIVATE KEY-----\n" + keyString + "-----END RSA PRIVATE KEY-----\n";
         } catch (java.security.KeyStoreException | java.security.NoSuchAlgorithmException | java.security.UnrecoverableKeyException ex) {
-            throw new RuntimeException(ex);
+            throw new RuntimeException("Failed to get private key from Java keystore", ex);
         }
         options.initMtls(certificate, privateKey);
         options.verifyPeer = true;
--- a/src/main/java/software/amazon/awssdk/crt/io/TlsContextOptions.java
+++ b/src/main/java/software/amazon/awssdk/crt/io/TlsContextOptions.java
@@ -380,7 +380,7 @@ public final class TlsContextOptions extends CrtResource {
             if (certificateData == null) {
                 throw new CrtRuntimeException("Certificate at given certificate alias is empty");
             }
-            String certificateString = StringUtils.simpleBase64ToString(certificateData.getEncoded());
+            String certificateString = StringUtils.simpleBase64Encode(certificateData.getEncoded());
             certificate = "-----BEGIN CERTIFICATE-----\n" + certificateString + "-----END CERTIFICATE-----\n";
         } catch (java.security.KeyStoreException | java.security.cert.CertificateEncodingException ex) {
             throw new CrtRuntimeException("Could not get certificate from Java keystore");
@@ -391,7 +391,7 @@ public final class TlsContextOptions extends CrtResource {
             if (keyData == null) {
                 throw new CrtRuntimeException("Private key at given certificate alias is empty");
             }
-            String keyString = StringUtils.simpleBase64ToString(keyData.getEncoded());
+            String keyString = StringUtils.simpleBase64Encode(keyData.getEncoded());
             privateKey = "-----BEGIN RSA PRIVATE KEY-----\n" + keyString + "-----END RSA PRIVATE KEY-----\n";
         } catch (java.security.KeyStoreException | java.security.NoSuchAlgorithmException ex) {
             throw new CrtRuntimeException("Could not get key from Java keystore");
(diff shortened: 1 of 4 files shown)


src/main/java/software/amazon/awssdk/crt/utils/StringUtils.java
```diff
@@ -23,4 +23,32 @@ public static String join(CharSequence delimiter, Iterable<? extends CharSequenc
         }
         return sb.toString();
     }
+
+    /**
+     * Decode Base64 byte[] array into a String in pure Java for Android and Java 7+ support.
+     * Based on: https://stackoverflow.com/a/4265472
+    */
+    private final static char[] BASE64_ALPHABET = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/".toCharArray();
```
TwistedTwigleg, line 31: Not ideal, but **currently** we support Android API level 24. This, unfortunately, means we cannot use the built-in `Base64` library, as it is supported from API level 26+. We also cannot use `javax.xml.bind.DatatypeConverter` because it was removed in Java 11+.

To work around this, I found a pure Java based solution that works for the purposes of the Java keystore, at least on Mac.

However, a better solution *might* be to just bump the minimum version of Android we support to API level 26, which is Android 8, and then we can use the built-in Java `Base64` library. [Android 8](https://en.wikipedia.org/wiki/Android_Oreo) is no longer supported and got its last update a year ago, but [only 83% use Android 8+](https://9to5google.com/2022/05/20/android-2022-distribution-numbers-chart/), so we would lose roughly 5% of the possibly supported Android base, so maybe not.

Regardless, the pure Java solution should work for now.

src/main/java/software/amazon/awssdk/crt/utils/StringUtils.java
```diff
@@ -23,4 +23,48 @@ public static String join(CharSequence delimiter, Iterable<? extends CharSequenc
         }
         return sb.toString();
     }
+
+    /**
+     * Encode a byte[] array into a Base64 String in pure Java
+     */
+    private final static char[] BASE64_VALID_CHARACTERS = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/".toCharArray();
+    public static String simpleBase64Encode(byte[] data) {
```
bretambrose, line 31: Can we add some tests for this ala the tests for the common impl: https://github.com/awslabs/aws-c-common/blob/main/tests/encoding_test.c#L395

TwistedTwigleg: The tests in the common impl have been added! I skipped all of the decoder tests, as this function just encodes and there is no decoder for now since we only need encoding for the Java keystore.

src/main/java/software/amazon/awssdk/crt/io/TlsContextOptions.java
```diff
@@ -356,6 +356,53 @@ public static TlsContextOptions createWithMtlsWindowsCertStorePath(String certif
...
+     */
+    public static TlsContextOptions createWithMtlsJavaKeystore(
+        java.security.KeyStore keyStore, String certificateAlias, String certificatePassword) {
+
+        TlsContextOptions options = new TlsContextOptions();
+        String certificate;
+        try {
+            java.security.cert.Certificate certificateData = keyStore.getCertificate(certificateAlias);
+            if (certificateData == null) {
+                throw new CrtRuntimeException("Certificate at given certificate alias is empty");
```
graebm, line 381: "empty" is a weird way to put it. The [docs for getCertificate()](https://docs.oracle.com/javase/7/docs/api/java/security/KeyStore.html#getCertificate(java.lang.String)) say it returns null if:
> ... alias does not exist or does not contain a certificate.

src/main/java/software/amazon/awssdk/crt/io/TlsContextOptions.java
```diff
@@ -356,6 +356,53 @@ public static TlsContextOptions createWithMtlsWindowsCertStorePath(String certif
...
+            }
+            String certificateString = StringUtils.simpleBase64Encode(certificateData.getEncoded());
+            certificate = "-----BEGIN CERTIFICATE-----\n" + certificateString + "-----END CERTIFICATE-----\n";
+        } catch (java.security.KeyStoreException | java.security.cert.CertificateEncodingException ex) {
+            throw new CrtRuntimeException("Could not get certificate from Java keystore");
```
graebm, line 386: Seems like the `KeyStoreException` or `CertificateEncodingException` could have useful information in them. Can we just rethrow these wrapped in a RuntimeException?

(same feedback applies everywhere you're catching checked exceptions, and throwing something else)

src/main/java/software/amazon/awssdk/crt/io/TlsContextOptions.java
```diff
@@ -356,6 +356,53 @@ public static TlsContextOptions createWithMtlsWindowsCertStorePath(String certif
...
+        }
+        String privateKey;
+        try {
+            java.security.Key keyData = keyStore.getKey(certificateAlias, certificatePassword.toCharArray());
+            if (keyData == null) {
+                throw new CrtRuntimeException("Private key at given certificate alias is empty");
```
graebm, line 392: same, "empty" is weird way of putting it. docs say:
> ... given alias does not exist or does not identify a key-related entry

src/main/java/software/amazon/awssdk/crt/utils/StringUtils.java
```diff
@@ -23,4 +23,58 @@ public static String join(CharSequence delimiter, Iterable<? extends CharSequenc
...
```
graebm, line 31: ‼️why can't we just use Java's built-in [Base64.Encoder](https://docs.oracle.com/javase/8/docs/api/java/util/Base64.Encoder.html)?

graebm (changes requested):

src/main/java/software/amazon/awssdk/crt/utils/StringUtils.java
TwistedTwigleg: See: https://github.com/awslabs/aws-crt-java/pull/552#discussion_r1023361516

Summary: We cannot use `Base64.Encoder` because we currently have the minimum Android API level set to 24, which does not support all of Java 8 (it is really Java 7 with some additions). API Level 26+ supports Java 8 and `Base64`, but that would be dropping roughly 5% of Android users based on 2022 statistics.

There are other built-in solutions outside of `Base64`, but:
* `javax.xml.bind.DatatypeConverter` works for Base64 support for API level 24, but was dropped in Java version 11 onward, where it was dropped completely in favor of `Base64`.
* `android.util.Base64` works for Android all the way back to API level 8, but is not included in standard Java. I could also not find a way to say "If in Android, use android.utils.Base64, otherwise use Java 8's Base64".
* Appache Commons Codec has a Base64 implementation, but you have to pull in their entire conversion library. Seems heavy for just Base64 encoding for just the Java keystore. Also unsure on license for library.

So making a Base64 encoder in pure Java without using any modules fixes this issue, allowing Android with API level 24 to work and all the versions of Java we support. It is not ideal, ideally we'd just drop Android API Level 24 for Android and then use `Base64`, but dropping roughly 5% of the Android devices just to support the Java keystore seems like a bad tradeoff.

graebm:
```
/*
 * Base 64 encodes the contents of to_encode and stores the result in output.
 */
AWS_COMMON_API
int aws_base64_encode(const struct aws_byte_cursor *AWS_RESTRICT to_encode, struct aws_byte_buf *AWS_RESTRICT output);
```
[https://github.com/awslabs/aws-c-common/blob/03d1b0e83094d1168b70b007f998b6a52ae507cf/include/aws/common/encoding.h#L69

src/test/java/software/amazon/awssdk/crt/test/StringUtilsTest.java
```diff
@@ -10,7 +10,7 @@
 import software.amazon.awssdk.crt.utils.StringUtils;
 
 
-public class StringUtilsTest {
+public class StringUtilsTest extends CrtTestFixture {
```
TwistedTwigleg, line 13: Just a note, mostly to myself: If you use anything that uses native C code, you HAVE to have `extends CrtTestFixture` or otherwise the test will fail!

What is really unfortunate is that it just reports it failed trying to call the native function but it doesn't say why it failed, that it was because the CRT was not loaded, or anything - it just says it fails calling the native function.
Spent WAY to long trying to find out why it was not working, debugging everything from top to bottom, and scratching my head on why it only failed on tests.

Moral of the story: always make sure `CrtTestFixture` is present when testing anything that touches native code. If you get strange errors but only in tests, double check to make sure `CrtTestFixture` is present.

src/native/string_utils.c
```diff
@@ -0,0 +1,96 @@
...
+
+    struct aws_byte_cursor formatted_data_cursor = aws_byte_cursor_from_buf(&formatted_data);
+    jbyteArray return_data = aws_jni_byte_array_from_cursor(env, &formatted_data_cursor);
+    aws_byte_buf_clean_up_secure(&formatted_data);
+
+    return return_data;
+
+clean_up:
+    aws_jni_byte_cursor_from_jbyteArray_release(env, jni_data, data_cursor);
+    return NULL;
```
graebm, line 52: I am not a fan of functions with multiple error-handling paths, where each path has slightly different cleanup needs. It's just so hard to get right, and error-handling paths can be very hard to test.

That's why, in our C APIs, we always try to have the clean-up function tolerate being called on something that's zeroed out or null. So, you can just naively try to clean up even if it was never initialized.

I suggest structuring the function like this:
```
jbyteArray JNICALL Java_software_thisFunction(...) {

    /* declare everything that may need cleanup, and zero it out */
    struct aws_byte_buf formatted_data;
    AWS_ZERO_STRUCT(formatted_data)
    jbyteArray returnData = NULL;

    struct aws_byte_cursor data_cursor =  aws_jni_byte_cursor_from_jbyteArray_acquire(env, jni_data);
    if (data_cursor.ptr == NULL) {
        goto cleanup;
    }

    ... blah blah blah if anything goes wrong goto cleanup

    return_data = aws_jni_byte_array_from_cursor(env, &formatted_data_cursor);

    /* hey check it out even the success case falls through to cleanup */

clean_up:
    aws_jni_byte_cursor_from_jbyteArray_release(env, jni_data, data_cursor);
    aws_byte_buf_clean_up(&formatted_data);
    return returnData;
}
```

src/main/java/software/amazon/awssdk/crt/io/TlsContextOptions.java
```diff
@@ -356,6 +356,51 @@ public static TlsContextOptions createWithMtlsWindowsCertStorePath(String certif
...
+        String certificate;
+        try {
+            java.security.cert.Certificate certificateData = keyStore.getCertificate(certificateAlias);
+            if (certificateData == null) {
+                throw new CrtRuntimeException("Certificate at given certificate alias does not exist or does not contain a certificate");
+            }
+            String certificateString = new String(StringUtils.base64Encode(certificateData.getEncoded()));
+            certificate = "-----BEGIN CERTIFICATE-----\n" + certificateString + "-----END CERTIFICATE-----\n";
+        } catch (java.security.KeyStoreException | java.security.cert.CertificateEncodingException ex) {
+            throw new RuntimeException(ex);
```
graebm, line 386: You could add a message to this with some context, like `throw new RuntimeException("Failed to get certificate", ex);`

src/main/java/software/amazon/awssdk/crt/io/TlsContextOptions.java
```diff
@@ -356,6 +356,51 @@ public static TlsContextOptions createWithMtlsWindowsCertStorePath(String certif
...
+        String privateKey;
+        try {
+            java.security.Key keyData = keyStore.getKey(certificateAlias, certificatePassword.toCharArray());
+            if (keyData == null) {
+                throw new CrtRuntimeException("Private key at given certificate alias does not exist or does not identify a key-related entity");
+            }
+            String keyString = new String(StringUtils.base64Encode(keyData.getEncoded()));
+            privateKey = "-----BEGIN RSA PRIVATE KEY-----\n" + keyString + "-----END RSA PRIVATE KEY-----\n";
+        } catch (java.security.KeyStoreException | java.security.NoSuchAlgorithmException | java.security.UnrecoverableKeyException ex) {
+            throw new RuntimeException(ex);
```
graebm, line 397: trivial: same, you could add a message along with the cause

src/native/string_utils.c
```diff
@@ -0,0 +1,102 @@
...
+
+    struct aws_byte_cursor data_cursor;
+    AWS_ZERO_STRUCT(data_cursor);
+    struct aws_byte_buf formatted_data;
+    AWS_ZERO_STRUCT(formatted_data);
+    jbyteArray return_data = NULL;
+
+    // Explicitly return NULL if the data we get is null to avoid an exception. NULL should return NULL
+    if (jni_data == NULL) {
+        return return_data;
```
graebm, line 73: NULL should result in a NullPointerException

aws_jni_byte_cursor_from_jbyteArray_acquire() does this check for you, and raises the appropriate exception

src/native/string_utils.c
```diff
@@ -0,0 +1,102 @@
...
+    jbyteArray jni_data) {
+    (void)jni_class;
+
+    struct aws_byte_cursor data_cursor;
+    AWS_ZERO_STRUCT(data_cursor);
+    struct aws_byte_buf formatted_data;
+    AWS_ZERO_STRUCT(formatted_data);
+    jbyteArray return_data = NULL;
+
+    // Explicitly return NULL if the data we get is null to avoid an exception. NULL should return NULL
```
graebm, line 25: NULL should result in a NullPointerException

src/test/java/software/amazon/awssdk/crt/test/StringUtilsTest.java
```diff
@@ -20,4 +22,95 @@ public void testJoin() {
...
     }
+
+    @Test
+    public void testBase64EncodeEmpty() {
+        assertEquals("", new String(StringUtils.base64Encode("".getBytes())));
+    }
+
+    @Test
+    public void testBase64EncodeNull() {
+        assertEquals(null, StringUtils.base64Encode(null));
```
graebm, line 33: `assertThrows(NullPointerException.class(), ...`

src/test/java/software/amazon/awssdk/crt/test/StringUtilsTest.java
```diff
@@ -20,4 +22,95 @@ public void testJoin() {
...
+    }
+
+    @Test
+    public void testBase64DecodeEmpty() {
+        assertEquals("", new String(StringUtils.base64Decode("".getBytes())));
+    }
+
+    @Test
+    public void testBase64DecodeNull() {
+        assertEquals(null, StringUtils.base64Decode(null));
```
graebm, line 73: assertThrows

graebm (approved):
fix & shipit

TwistedTwigleg:

PR #557, opened 2022-11-30
Canary adjustments, MQTT5 tests against IoT Core

TwistedTwigleg:
*Description of changes:*

Adds additional error logging to the Canary (particularly around disconnects and exceptions) to help and try to figure out why sometimes the socket is closed when running the Canary for hours. Also fixes the ClientID not being set correctly in the Canary, adds some code to make sure operations cannot be set to run in parallel on the same client (though unlikely to happen), and adjusts the Canary to run against IoT Core.

This PR also adjusts all the MQTT5 tests to have the option to run against IoT Core when possible. This allows GitHub actions to run MQTT5 tests, and allows enabling a test that was temporarily shelved when running in Codebuild (testing IoT Core connection packet limits).
The MQTT5 Codebuild job was removed and the Codebuild MQTT5 tests now run in the integration tests Codebuild job.

Finally, these CI changes to run MQTT5 tests in GitHub actions showed a number of memory leaks and instability problems with the MQTT5 client. Some of these fixes were implemented in other PRs (#570 and #571), but this PR also includes a number of fixes as well. These fixes are:
* JNI exceptions are no longer thrown by intermediate functions, like getting/setting packet properties. Now exceptions are ONLY thrown at the highest level.
* Fixed issue where JNI frame could be popped without a push in a few locations
* Fixed memory leaks when an operation had an error before trying to call the operation in the native client
* Fixed code cleaning AWS arrays not having proper safety checks

There are also a few minor changes that are just clean up:
* Removed uploading Java error reports in both Codebuild and GitHub CI. Neither was working, so no reason to keep them.
* Added code to print segfault error reports in Codebuild, should they occur.
* Fixed MQTT5 sample using the wrong variable name for the return result in `onConnectionFailure`
__________


xiazhvera (approved):

TwistedTwigleg:

PR #566, opened 2022-12-09
Add tests to make sure interrupting operations works

TwistedTwigleg:
*Description of changes:*

Adds MQTT5 tests to make sure that interrupting operations works as expected and does not cause crashes. With the current behavior, when you interrupt an operation you get an exception, with the cause of that exception being a CRT exception of `AWS_ERROR_MQTT5_USER_REQUESTED_STOP`.

__________

### DIFF
--- a/src/test/java/software/amazon/awssdk/crt/test/Mqtt5ClientTest.java
+++ b/src/test/java/software/amazon/awssdk/crt/test/Mqtt5ClientTest.java
@@ -287,11 +287,6 @@ public class Mqtt5ClientTest extends CrtTestFixture {
                 try (Mqtt5Client client = new Mqtt5Client(builder.build())) {
                     assertNotNull(client);
                 }
-
-                elg.close();
-                hr.close();
-                bootstrap.close();
-                socketOptions.close();
             }
 
         } catch (Exception ex) {
@@ -426,7 +421,6 @@ public class Mqtt5ClientTest extends CrtTestFixture {
                 events.connectedFuture.get(60, TimeUnit.SECONDS);
                 DisconnectPacketBuilder disconnect = new DisconnectPacketBuilder();
                 client.stop(disconnect.build());
-                client.close();
             }
         } catch (Exception ex) {
             fail(ex.getMessage());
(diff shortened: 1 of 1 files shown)


src/test/java/software/amazon/awssdk/crt/test/Mqtt5ClientTest.java
```diff
@@ -2512,5 +2512,133 @@ public void Retain_UC1() {
...
+        Assume.assumeTrue(mqtt5DirectMqttPort != null);
+        int messageCount = 10;
+        String testUUID = UUID.randomUUID().toString();
+        String testTopic = "test/MQTT5_Binding_Java_" + testUUID;
+
+        try {
+            Mqtt5ClientOptionsBuilder builder = new Mqtt5ClientOptionsBuilder(mqtt5DirectMqttHost, mqtt5DirectMqttPort);
+            LifecycleEvents_Futured events = new LifecycleEvents_Futured();
+            builder.withLifecycleEvents(events);
+            Mqtt5Client client = new Mqtt5Client(builder.build());
```
bretambrose, line 2535: Havent looked at the other tests, but from an example standpoint, I would much prefer to see the client built in a try-with-resources block, dropping the explicit close at the end.  The way it's currently done, any exception in the try block leaks the client.

TwistedTwigleg: Done! Now all CrtResources use try-with-resources blocks. The only exceptions are the overflow tests, which would have been really messy to convert (and they shouldn't have usable CrtResources that could leak anyway).

src/test/java/software/amazon/awssdk/crt/test/Mqtt5ClientTest.java
```diff
@@ -218,77 +218,81 @@ public void New_UC2() {
...
+                try (Mqtt5Client client = new Mqtt5Client(builder.build())) {
+                    assertNotNull(client);
+                }
 
-            elg.close();
-            hr.close();
-            bootstrap.close();
-            socketOptions.close();
-            client.close();
+                elg.close();
```
bretambrose, line 291: These shouldn't be necessary.  If done correctly you should only ever have to manually close something that you let escape the try block via doing an addRef on it

TwistedTwigleg: Good catch - I missed that one! I did a check and now anywhere `close()` is called is either in commented out tests (which I need to get back to and/or fix) and in the overflow tests.

bretambrose (approved):

TwistedTwigleg:

PR #567, opened 2022-12-12
Change loading method about JNI native library.

Kensaku-MASUDA:
I think that we can not address arm cpu architecture by JAVA system properties including any kind of embedded JVM. But dymanic shared object loader know collect architecture.
For example, some embedded jvm ported by sun(oracle) has following properties

- ARM Cortex-A15 /Wind River Linux 6
    - os.name
        - linux
    - os.arch
        - arm
    - sun.arch.abi
        - gnueabi
    - sun.arch.data.model
        - 32
- ARM Cortex-A57 /Wind River Linux 9
    - os.name
        - linux
    - os.arch
        - arm
    - sun.arch.abi
        - gnueabihf
    - sun.arch.data.model
        - 32
- ARM Cortex-A57 /Wind River Linux 9
    - os.name
        - linux
    - os.arch
        - arm
    - sun.arch.abi
        - gnueabihf
    - sun.arch.data.model
        - 64

Changes:
- CRT#getArchIdentifier()
    - Change method return type from string to array  of string
- CRT#extractAndLoadLibrary()
    - trying to load dynamic link library that collect one.

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/CRT.java
+++ b/src/main/java/software/amazon/awssdk/crt/CRT.java
@@ -103,24 +103,20 @@ public final class CRT {
     /**
      * @return a string describing the detected architecture the CRT is executing on
      */
-    public static String getArchIdentifier() throws UnknownPlatformException {
+    public static String[] getArchIdentifier() throws UnknownPlatformException {
         CrtPlatform platform = getPlatformImpl();
         String arch = normalize(platform != null ? platform.getArchIdentifier() : System.getProperty("os.arch"));
 
         if (arch.matches("^(x8664|amd64|ia32e|em64t|x64|x86_64)$")) {
-            return "x86_64";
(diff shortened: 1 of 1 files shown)


bretambrose:
In theory, I like this idea, but a surface scan leaves an unresolved question: I'm unsure if this will handle the case of a host that is armv6 only.  Looking at openjdk's architecture checks (https://github.com/openjdk/jdk/blob/master/src/hotspot/os/linux/os_linux.cpp#L1447) I don't see anything that makes me think that mismatch will be caught.  So, given the priority, it may always load the v7 binary and then crash as soon as an unexpected instruction gets decoded.  

I don't have access to armv6 hardware to test this on.  Mostly we work with emulation (qemu, which emulates everything, maybe there's a way to make it only emulate v6 and we could verify) or aarch64 hosts (which would hide the issue).

PR #568, opened 2022-12-15
Rejoin always option and tests

bretambrose:
* Adds support for a new session behavior type that always attempts to rejoin a session, even if existing session state is unknown.


TwistedTwigleg (approved):
Looks good to me - Thanks for adding support!

PR #569, opened 2022-12-16
split the artifact to be platform specific

TingDaoK:
- Add classifier to the maven build
- Change code pipeline to get platform specific jar

Issue:
- Mac ARM build generates a lot of stuff and got passed down. https://us-east-1.console.aws.amazon.com/codesuite/codebuild/123124136734/projects/aws-crt-java-cd-deploy-snapshot/build/aws-crt-java-cd-deploy-snapshot%3A961e5727-0db8-4984-9493-b249a60959f7/?region=us-east-1
- Use S3 bucket for all the artifacts. 

TODO:
- The uber jar is around 11MB, the biggest platform-sepcific jar is around 3MB.
- We can still create dynamic link SSL jar that may reduce the linux jar from 3MB to 1MB.


builder.json
```diff
@@ -49,7 +48,9 @@
                 "armv6": {
                     "!packages": [],
                     "!build_steps": [
-                        "build"
+                        "build",
+                        "mv target/cmake-build/aws-crt-java/* target/cmake-build/",
```
TingDaoK, line 52: move the cross-compiled libs to the expected lib path to create jar. json file doesn't support comment.

Previous `"build"` only use cmake to build the shared lib.

codebuild/cd/cross-unix-build.sh
```diff
@@ -6,10 +6,10 @@ set -ex
 
 cd `dirname $0`/../..
```
TingDaoK, line 7: Actually, I think this script is a dead script that no one uses.

codebuild/cd/generic-unix-build.sh
```diff
@@ -12,24 +12,25 @@ if [ -z "$AWS_CRT_TARGET" ]; then
     AWS_CRT_TARGET=$AWS_CRT_HOST
 fi
 
-LIB_PATH=target/cmake-build/lib
 SKIP_INSTALL=
 
-# Cross compiles do not need local installs, and they have a different lib output path
 if [[ "$AWS_CRT_TARGET" != "$AWS_CRT_HOST" ]]; then
     SKIP_INSTALL=--skip-install
-    LIB_PATH=target/cmake-build/aws-crt-java/lib
```
TingDaoK, line 21: I did `"mv target/cmake-build/aws-crt-java/* target/cmake-build/"` in builder.json to replace this

codebuild/cd/manylinux-x64-build.yml
```diff
@@ -11,16 +11,21 @@ phases:
...
+      - JAVA_HOME=/opt/java-se-8u40-ri/ mvn -B install -DskipTests -Plinux-x64
+      - ls target
   post_build:
     commands:
       # get the shared libs from the native build
       - mkdir -p ../dist
       - cp -rv target/cmake-build/lib ../dist/
-
+      # get the platform specific jar with classifier
+      - cp target/*.jar ../dist/
```
TingDaoK, line 24: The image used doesn't have aws-cli installed. Still use codebuild to transfer the artifacts.

codebuild/cd/osx-arm64-build.sh
```diff
@@ -5,9 +5,16 @@ set -ex
...
+export GIT_TAG=$(git describe --tags)
 
-mvn -B compile -P mac-arm64
+# use a fix deploy version for platform specific jar
+mvn versions:set -DnewVersion=deploy
+mvn -B install -DskipTests -P mac-arm64
 
 # Copy artifacts to dist
 mkdir -p ../dist
+# Note: as we ran mvn install, the dist has unexpected files, using S3 bucket to transfer artifact instead
```
TingDaoK, line 16: Haven't really dig into why. But the mac-arm64 build is the only job that I saw unexpected files, and it has the same setting as the x64 build. this [log](https://us-east-1.console.aws.amazon.com/codesuite/codebuild/123124136734/projects/aws-crt-java-cd-deploy-snapshot/build/aws-crt-java-cd-deploy-snapshot%3A961e5727-0db8-4984-9493-b249a60959f7/?region=us-east-1) shows that.

pom.xml
```diff
@@ -115,6 +116,7 @@
             <properties>
                 <cmake.min_osx_version>-DCMAKE_OSX_DEPLOYMENT_TARGET=10.9</cmake.min_osx_version>
                 <cmake.osx_arch>-DCMAKE_OSX_ARCHITECTURES=arm64</cmake.osx_arch>
+                <configured.classifier>mac-arm64</configured.classifier>
```
TingDaoK, line 119: All those arm classifiers are not same as the one in [here](https://github.com/trustin/os-maven-plugin#property-osdetectedarch), which is a recommended tool to auto check the os.arch. 

I am not sure about this, on one side, I think we can let user know the arch name we chose to be more clear about the arch (especially we have those armv6/7/8 linux build). One the other side, I think we can adapt what the auto tool find, like [here](https://github.com/awslabs/aws-crt-java/blob/main/src/main/java/software/amazon/awssdk/crt/CRT.java#L106)

codebuild/cd/cross-unix-build.sh
graebm: let's be bold and delete it

pom.xml
DmitriyMusatkin: different arm versions of crt are not interchangeable, right? so we need need to extend the recommended tools list anyways.
im leaning more towards our own list, but clearly defining the name for it (ex. crt-platform.classifier) and specifying in the readme which ones we publish to maven

codebuild/cd/deploy-platform-specific-jar.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/sh
```
graebm, line 1: let's rename this
deploy-platform-specific-jars.sh
~jar~ -> jars

codebuild/cd/deploy-platform-specific-jar.sh
```diff
@@ -0,0 +1,40 @@
...
+  # Need to sign the file to deploy to staging repo
+  DEPLOY_FILE_GOAL=gpg:sign-and-deploy-file
+  # I got the URL from the deployment log and tests around.
+  DEPLOY_REPOSITORY_URL=https://aws.oss.sonatype.org:443/service/local/staging/deployByRepositoryId/${STAGING_REPO_ID}
+fi
+
+PLATFORM_ARRAY=("linux-armv6" "linux-armv7" "linux-armv8" "linux-x86_32" "linux-x86_64" "mac-arm64" "mac-x86_64" "windows-x86_32" "windows-x86_64")
+for str in ${PLATFORM_ARRAY[@]}; do
+  FILES="${FILES}target/aws-crt-deploy-$str.jar,"
+  CLASSIFIER="${CLASSIFIER}${str},"
```
graebm, line 21: trivial: ~CLASSIFIER~ -> CLASSIFIERS

codebuild/cd/deploy-snapshot.yml
```diff
@@ -22,12 +22,16 @@ phases:
     commands:
       - cd $CODEBUILD_SRC_DIR/aws-crt-java
       - export PKG_VERSION=$(git describe --tags | cut -f2 -dv)
+      - echo $PKG_VERSION
       - mkdir -p target/cmake-build/lib
       - aws s3 cp --recursive s3://aws-crt-java-pipeline/v${PKG_VERSION}/lib $CODEBUILD_SRC_DIR/aws-crt-java/target/cmake-build/lib
+      - aws s3 cp --recursive s3://aws-crt-java-pipeline/v${PKG_VERSION}/jar $CODEBUILD_SRC_DIR/aws-crt-java/target/
+      # mv all the classified jar to target/
```
graebm, line 29:
```suggestion
      # mv all the platform-specific jars to target/
```

codebuild/cd/deploy-snapshot.yml
```diff
@@ -22,12 +22,16 @@ phases:
...
+      - mv $CODEBUILD_SRC_DIR_linux_x64/dist/aws-crt-deploy-linux-x86_64.jar $CODEBUILD_SRC_DIR/aws-crt-java/target/
+      - mv $CODEBUILD_SRC_DIR_linux_x86/dist/aws-crt-deploy-linux-x86_32.jar $CODEBUILD_SRC_DIR/aws-crt-java/target/
+      # cp all the shared lib to cmake-build
```
graebm, line 32:
```suggestion
      # cp all the shared libs to cmake-build
```

codebuild/cd/deploy-snapshot.yml
```diff
@@ -22,12 +22,16 @@ phases:
     commands:
       - cd $CODEBUILD_SRC_DIR/aws-crt-java
       - export PKG_VERSION=$(git describe --tags | cut -f2 -dv)
+      - echo $PKG_VERSION
       - mkdir -p target/cmake-build/lib
       - aws s3 cp --recursive s3://aws-crt-java-pipeline/v${PKG_VERSION}/lib $CODEBUILD_SRC_DIR/aws-crt-java/target/cmake-build/lib
```
graebm, line 27: trivial: can we re-order these so that we do all the stuff to put the libs in the right place, then do all the stuff to put the jars in the right place?

right now, the steps are kind of mixed together

codebuild/cd/generic-unix-build.sh
```diff
@@ -12,24 +12,29 @@ if [ -z "$AWS_CRT_TARGET" ]; then
...
-./builder build -p aws-crt-java --target=$AWS_CRT_TARGET run_tests=false
 
 # Upload the lib to S3
 GIT_TAG=$(git describe --tags)
-aws s3 cp --recursive $LIB_PATH s3://aws-crt-java-pipeline/${GIT_TAG}/lib
+
+# use a fix deploy version for platform specific jar
+mvn versions:set -DnewVersion=deploy
+./builder build -p aws-crt-java --target=$AWS_CRT_TARGET run_tests=false
+# Builder corss compile the shared lib to `target/cmake-build/aws-crt-java/`, move it to th expected path for mvn to generate jar.
```
graebm, line 34:
```suggestion
# Builder cross-compiles the shared lib to `target/cmake-build/aws-crt-java/`, move it to the expected path for mvn to generate the jar.
```

pom.xml
```diff
@@ -128,6 +130,44 @@
             </activation>
             <properties>
                 <cmake.cflags>-m32</cmake.cflags>
+                <configured.classifier>linux-x86_32</configured.classifier>
+            </properties>
+        </profile>
+
+        <profile>
+            <id>windows-64</id>
```
graebm, line 138: let's have the profile ID match the classifier, like "windows-x86_64"

pom.xml
```diff
@@ -128,6 +130,44 @@
...
+        <profile>
+            <id>windows-32</id>
+            <properties>
+                <configured.classifier>windows-x86_32</configured.classifier>
+            </properties>
+        </profile>
+        <profile>
+            <id>linux-armv6</id>
+            <properties>
+                <configured.classifier>linux-armv6</configured.classifier>
```
graebm, line 152: you said we're matching the os-maven-plugin, but it just has "arm_32" and "aarch64"

we want our stuff to work with this plugin

if we need different armv6/armv7 builds, should the "arm_32" just include 2 different binaries?

pom.xml
graebm: the os-maven-plugin uses "osx" not "mac"

README.md
```diff
@@ -130,3 +130,62 @@ To debug native code with VSCode or CLion or any other IDE:
 3. Set the parameters to be the ones used by the ```mvn``` script, as per above
 4. Set the working directory to the `aws-crt-java` directory
 5. On windows, you will need to manually load the PDB via the Modules window in Visual Studio, as it is not embedded in the JAR. It will be in the ```target/cmake-build/lib/windows/<arch>``` folder.
+
+## Platform-Specific JARs
+
+Since v0.20.5, platform-specific JARs are available to reduce the size of the artificts. To consume the platform-specific JAR of aws-crt-java, you can use Maven classifier (https://maven.apache.org/pom.html) property to specify the platform. Without classifier, you will get the uber jar to support all distributions.
```
DmitriyMusatkin, line 136: nit: artificts -> artifacts

README.md
```diff
@@ -130,3 +130,62 @@ To debug native code with VSCode or CLion or any other IDE:
...
+- linux-x86_64
+- osx-aarch_64
+- osx-x86_64
+- windows-x86_32
+- windows-x86_64
+
+### Auto-detected
+There are build plugins such as os-maven-plugin (https://github.com/trustin/os-maven-plugin) that help project automatically determine the OS at build time.
+
+**NOTES**: For auto-detected `linux-arm_32` archtecture, you need to specific `linux-armv6` or `linux-armv7` to consume the excat platform-specific jar.
```
DmitriyMusatkin, line 169: minor: i would reword it slightly to say that we follow common platform names (as in the link) to make it easier to detect platform. But for arm_32 we have 2 types that are not compatible and user needs to manually make the decision.

DmitriyMusatkin: are we sure that aarch_64 build are compatible across all arm 64 types? arm 8/9/etc...?

DmitriyMusatkin (approved):

README.md
```diff
@@ -130,3 +130,62 @@ To debug native code with VSCode or CLion or any other IDE:
 3. Set the parameters to be the ones used by the ```mvn``` script, as per above
 4. Set the working directory to the `aws-crt-java` directory
 5. On windows, you will need to manually load the PDB via the Modules window in Visual Studio, as it is not embedded in the JAR. It will be in the ```target/cmake-build/lib/windows/<arch>``` folder.
+
+## Platform-Specific JARs
+
+Since v0.20.5, platform-specific JARs are available to reduce the size of the artificts. To consume the platform-specific JAR of aws-crt-java, you can use Maven classifier (https://maven.apache.org/pom.html) property to specify the platform. Without classifier, you will get the uber jar to support all distributions.
+
+``` xml
+        <!-- Linux x86_64 distribution  -->
```
graebm, line 139:
```suggestion
        <!-- Platform-specific Linux x86_64 JAR -->
```

I didn't like introducing a new word ~distribution~ when we're already using the term "platform"

Just trying to reduce the number of different words we use to describe "platform-specific" (distribution, platform, OS, architecture, classifier). Since Java devs don't usually need to think about that kind of stuff

README.md
```diff
@@ -130,3 +130,62 @@ To debug native code with VSCode or CLion or any other IDE:
...
+        <dependency>
+            <groupId>software.amazon.awssdk.crt</groupId>
+            <artifactId>aws-crt</artifactId>
+            <version>0.20.5</version>
+            <classifier>linux-x86_64</classifier>
+        </dependency>
+
```
graebm, line 146: Make these two different code blocks. Someone would never want to specify both the uber jar AND a platform specific jar, but since they're the same code block here it kind of looks like that's what you're supposed to do

README.md
```diff
@@ -130,3 +130,62 @@ To debug native code with VSCode or CLion or any other IDE:
...
+        <!-- Uber JAR with all distributions -->
+        <dependency>
+            <groupId>software.amazon.awssdk.crt</groupId>
+            <artifactId>aws-crt</artifactId>
+            <version>0.20.5</version>
+        </dependency>
+```
+
+### Available classifier
```
graebm, line 155:
```suggestion
### Available platform classifiers
```

README.md
```diff
@@ -130,3 +130,62 @@ To debug native code with VSCode or CLion or any other IDE:
...
+- linux-aarch_64
+- linux-x86_32
+- linux-x86_64
+- osx-aarch_64
+- osx-x86_64
+- windows-x86_32
+- windows-x86_64
+
+### Auto-detected
+There are build plugins such as os-maven-plugin (https://github.com/trustin/os-maven-plugin) that help project automatically determine the OS at build time.
```
graebm, line 167: are there really other build plugins that provide the exact classifiers we're using? if not, just mention the one like
```suggestion
### Auto-detect
The [os-maven-plugin](https://github.com/trustin/os-maven-plugin) can automatically detect your platform's classifier at build time.
```

graebm (commented):
nitpicking the user instructions

README.md
TingDaoK: TBH, I don't know, but I think if we don't support all the arm64 other than armv8, we should know, as it is not supported now as well. This change doesn't break/change the behavior for those type.

README.md
```diff
@@ -130,3 +130,62 @@ To debug native code with VSCode or CLion or any other IDE:
 3. Set the parameters to be the ones used by the ```mvn``` script, as per above
 4. Set the working directory to the `aws-crt-java` directory
 5. On windows, you will need to manually load the PDB via the Modules window in Visual Studio, as it is not embedded in the JAR. It will be in the ```target/cmake-build/lib/windows/<arch>``` folder.
+
+## Platform-Specific JARs
+
+Since v0.20.5, platform-specific JARs are available to reduce the size of the artificts. To consume the platform-specific JAR of aws-crt-java, you can use Maven classifier (https://maven.apache.org/pom.html) property to specify the platform. Without classifier, you will get the uber jar to support all distributions.
+
+``` xml
```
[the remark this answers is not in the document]
TingDaoK: I believe we do, as we basically use the same logic to as tensorflow. [here](https://github.com/tensorflow/java/blob/master/deploy.sh#L68).

graebm (approved):

PR #570, opened 2022-12-22
Use correct JNI call

bretambrose:
* Crash fix - use correct JNI function when creating an MQTT5 client


TwistedTwigleg (approved):
Thanks!

PR #571, opened 2022-12-27
General stability and memory leak improvements to MQTT5

TwistedTwigleg:
*Description of changes:*

Makes the following improvements to the MQTT5 code:

* Fixes memory leak in MQTT5 client by freeing host name cursor (previously was never freed)
* Fixes memory leaks in packets that were using JNI strings or JNI byte arrays with the holder struct in mqtt5_packets.c
* Removes double pointer code for getting data from jobjects and instead uses a boolean pointer to determine if the optional was set or not
  * Fixes workaround where enums had to be set to UINT8_MAX to determine if the enum was set or not.
  * Simplifies knowing what to pass or not, and removes some workarounds I had to implement in a few places when only a single pointer was needed not a double pointer.
  * (Debatable) Removing the double pointers makes the code less scary from a memory management perspective in my opinion. Less complexity = easier to find memory leaks and/or race conditions
* Fixes publish callback popping the JNI local frame even though it is popped right after (leading to double pops!)
* Adjusts operation callbacks for better formatting and error checking
* Fixes MQTT5 sample not compiling due to incorrect types
* Adjusts developer preview string in the MQTT5 client options builder for consistency
* Minor fixes and better logging for MQTT5 canary
* Removes C++ style comments (`// comment` -> `/* comment */`)
* Properly set and clean HTTP proxy as strings rather than incorrectly as byte arrays

__________


xiazhvera (approved):

TwistedTwigleg:

PR #574, opened 2023-01-04
Remove S3NativeClient

TingDaoK:
- No one actual uses S3NativeClient for now
- It's a loooot of code that contributes 1MB to our Jar. Let's get rid of it.


DmitriyMusatkin:
can we get rid of smithy-crt as well, which has codegen tools?

DmitriyMusatkin:
nit: settings.gradle makes references to the folders you deleted

DmitriyMusatkin (approved):

PR #581, opened 2023-01-19
deprecate s3 client level endpoint setting

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
s3 client level endpoint option was only used for tls context previously. However on CRT side it was overriden per request with the request specific host. So in practice it was not used. Set option to deprecated as it does not affect anything. 


waahm7 (approved):

TingDaoK (commented):
Pick up the change from aws-c-s3 as well?

PR #584, opened 2023-01-27
Use long for window size

waahm7:
*Description of changes:*
- Use long for window size to make it consistent with S3 window size

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/http/HttpClientConnectionManagerOptions.java
+++ b/src/main/java/software/amazon/awssdk/crt/http/HttpClientConnectionManagerOptions.java
@@ -15,14 +15,14 @@ import software.amazon.awssdk.crt.io.TlsContext;
  */
 public class HttpClientConnectionManagerOptions {
     public static final int DEFAULT_MAX_BUFFER_SIZE = 16 * 1024;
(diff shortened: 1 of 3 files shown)


src/main/java/software/amazon/awssdk/crt/http/HttpClientConnectionManagerOptions.java
```diff
@@ -15,14 +15,14 @@
  */
 public class HttpClientConnectionManagerOptions {
     public static final int DEFAULT_MAX_BUFFER_SIZE = 16 * 1024;
-    public static final int DEFAULT_MAX_WINDOW_SIZE = Integer.MAX_VALUE;
+    public static final long DEFAULT_MAX_WINDOW_SIZE = Integer.MAX_VALUE;
```
TingDaoK, line 18: trivial: should this be changed to long max?

TingDaoK (approved):

graebm (approved):

src/main/java/software/amazon/awssdk/crt/http/HttpClientConnectionManagerOptions.java
waahm7: I am not sure if that will work on 32-bit systems where the maximum value of size_t is 32 bits.

PR #586, opened 2023-02-03
MQTT311 extra connection callbacks

TwistedTwigleg:
*Description of changes:*

Adds additional connection callbacks to MQTT311. Specifically, adds `onConnectionSuccess`, `onConnectionFailure`, and `onConnectionClosed` These callbacks mirror those  in V1.

Also adjusts the MQTT311 connection statistics tests so they are less flakey/inconsistent.

Will close: https://github.com/aws/aws-iot-device-sdk-java-v2/issues/203

__________


src/main/java/software/amazon/awssdk/crt/mqtt/OnConnectionSuccessReturn.java
```diff
@@ -0,0 +1,27 @@
+package software.amazon.awssdk.crt.mqtt;
+
+/**
+ * The data returned when the connection success callback is invoked in a connection.
+ * @see software.amazon.awssdk.crt.mqtt.MqttClientConnectionEvents
+ */
+public class OnConnectionSuccessReturn {
```
TwistedTwigleg, line 7: Note: Java is the only binding that does not expose the ReturnCode from the server when a successful connection is made. However, this return code has not been exposed at all in MQTT311 for Java, while the others have exposed it prior in some form.

Because of this, I haven't exposed it here, but should we add support for it?

xiazhvera (approved):

src/main/java/software/amazon/awssdk/crt/mqtt/OnConnectionSuccessReturn.java
TwistedTwigleg: No, I don't think it's really required to be honest, I just noticed it was a difference. On a success, it should always just be the same result (SUCCESS) since I do not think its possible to get a non-success code.

Thanks for taking a look at the PR!

TwistedTwigleg:

PR #593, opened 2023-03-11
Add serializable to clients

davidlghellin:
Description of changes:

Add serializable to the clients, this would help to work in Spark for example, since it needs the objects to be serializable


bretambrose:
Just an off-the-cuff response without looking too closely (or knowing anything about how serialization is used), but these types are fundamentally not-serializable.  They contain raw pointers to unmanaged memory and they cannot be cloned.

PR #595, opened 2023-03-15
Ability to pass request file path to S3, for faster uploads

graebm:
See https://github.com/awslabs/aws-c-s3/pull/266 for details

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/s3/S3MetaRequestOptions.java
+++ b/src/main/java/software/amazon/awssdk/crt/s3/S3MetaRequestOptions.java
@@ -159,8 +159,9 @@ public class S3MetaRequestOptions {
     /**
      * Set the initial HTTP request.
      *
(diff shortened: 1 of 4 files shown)


src/main/java/software/amazon/awssdk/crt/s3/S3MetaRequestOptions.java
```diff
@@ -163,6 +174,24 @@ public HttpRequest getHttpRequest() {
...
     }
 
+    /**
+     * If set, this file is sent as the request's body.
+     *
+     * When uploading a file, you should set this for better performance than if you
+     * set a body stream on {@link withHttpRequest}).
+     *
+     * @param requestFilePath path to file to send as the request's body.
+     * @return this
```
waahm7, line 184: trivial: I think we should mention that if the user sets both, the body stream will be ignored.

waahm7 (approved):

PR #598, opened 2023-03-30
Thread name

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
Add thread name if available when attaching callbacks to jvm

### DIFF
--- a/src/native/crt.c
+++ b/src/native/crt.c
@@ -72,11 +72,32 @@ static JNIEnv *s_aws_jni_get_thread_env(JavaVM *jvm) {
 #endif
     if ((*jvm)->GetEnv(jvm, (void **)&env, JNI_VERSION_1_6) == JNI_EDETACHED) {
         AWS_LOGF_DEBUG(AWS_LS_COMMON_GENERAL, "s_aws_jni_get_thread_env returned detached, attaching");
+        
+        struct aws_string *thread_name = NULL;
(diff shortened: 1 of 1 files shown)


src/native/crt.c
```diff
@@ -72,11 +72,32 @@ static JNIEnv *s_aws_jni_get_thread_env(JavaVM *jvm) {
...
+
+        struct aws_string *thread_name = NULL;
+        if (aws_thread_current_name(aws_jni_get_allocator(), &thread_name)) {
+            /* Retrieving thread name can fail for multitude of reasons and is
+            not fatal. Ignore the error and continue. */
+            AWS_LOGF_DEBUG(AWS_LS_COMMON_GENERAL, "Failed to retrieve name for the thread.");
+        }
+
+        struct JavaVMAttachArgs attach_args = {
+            .version = JNI_VERSION_1_6,
```
TingDaoK, line 84: Why `.version = JNI_VERSION_1_6`?

DmitriyMusatkin: 3 lines above this change we are also getting env with `JNI_VERSION_1_6`. Dont know much context on why we hardcoded 1.6 since i dont think we ever supported anything older than java 8. But sure static var on min supported version might be good.

DmitriyMusatkin:
> 

i checked in the flight recorder that thread name shows up. dont think there is an easy alternative way to verify that it was set

TingDaoK (approved):

PR #602, opened 2023-04-13
Handle jvm oom

TingDaoK:
- `aws_jni_byte_array_from_cursor` can failed with `OutOfMemoryError` exception.
- We didn't handle the failure and passing in the NULL pointer to the callback, which covers the real OOM issue and result in a NullPointerException. Handle that.


DmitriyMusatkin (approved):

src/native/s3_client.c
```diff
@@ -325,6 +325,11 @@ static int s_on_s3_meta_request_body_callback(
     }
 
     jobject jni_payload = aws_jni_byte_array_from_cursor(env, body);
+    if (jni_payload == NULL) {
+        /* JVM is out of memory, but native code can still have memory available, handle it and don't crash. */
+        aws_jni_check_and_clear_exception(env);
+        return aws_raise_error(AWS_ERROR_JAVA_CRT_JVM_OUT_OF_MEMORY);
+    }
```
graebm, line 332: We can't just return, we need to ensure aws_jni_release_thread_env() is called

PR #606, opened 2023-04-18
increase test timeout time

TwistedTwigleg:
*Description of changes:*

Increases the timeout time for MQTT5 operations from `180` to `300`, hopefully fixing some flaky tests. It also changes the tests to all use a single define for the operation timeout time, so it can easily be adjusted in one place rather than over 100 different places in the test file.

________


TwistedTwigleg:
Thanks! Enabling auto-merge into main...

PR #611, opened 2023-05-01
Additional unit-tests and unit-test adjustment

TwistedTwigleg:
*Description of changes:*

Adds new unit tests to check for MQTT311 and MQTT5 connections for:
* PKCS11
* PKCS12
* Java Keystore
* Windows Certificate Store
* etc.

As well as adjusting existing tests to use environment variables instead of Java properties. It uses a new [CRT builder action here](https://github.com/awslabs/aws-crt-builder/pull/231) to setup the environment variables. Finally, it adjusts Codebuild to also use the CRT builder so everything has the same setup across the board.

_______


src/test/java/software/amazon/awssdk/crt/test/CustomKeyOpsTest.java
```diff
@@ -131,31 +129,109 @@ public void performOperation(TlsKeyOperation operation) {
...
+        public void onStopped(Mqtt5Client client, software.amazon.awssdk.crt.mqtt5.OnStoppedReturn onStoppedReturn) {}
     }
 
     @Test
-    public void testExceptionFailurePath() {
+    public void testHappyPathMQTT5() {
         skipIfNetworkUnavailable();
-        Assume.assumeTrue(TEST_PRIVATEKEY != null && TEST_PRIVATEKEY != "");
+        // Not ideal, but I don't see the point of making a new file just for a single MQTT5 Custom Key Ops test.
+        String AWS_TEST_MQTT5_IOT_CORE_HOST = System.getenv("AWS_TEST_MQTT5_IOT_CORE_HOST");
```
sbSteveK, line 181: Is there a reason why these strings can't be added to MqttClientConnectionFixture.java like the rest?

src/test/java/software/amazon/awssdk/crt/test/IotServiceTest.java
```diff
@@ -9,47 +9,69 @@
...
+        Assume.assumeTrue(AWS_TEST_MQTT311_IOT_CORE_RSA_CERT != null);
+        Assume.assumeTrue(AWS_TEST_MQTT311_IOT_CORE_RSA_KEY != null);
         Consumer<MqttMessage> messageHandler = (message) -> {};
+        int port = 8883;
+
+        try (TlsContextOptions contextOptions = TlsContextOptions.createWithMtlsFromPath(
+                AWS_TEST_MQTT311_IOT_CORE_RSA_CERT,
+                AWS_TEST_MQTT311_IOT_CORE_RSA_KEY);)
+            {
+                // if (TlsContextOptions.isAlpnSupported()) {
```
sbSteveK, line 42: Delete or comment why we're keeping this commented block?

src/test/java/software/amazon/awssdk/crt/test/Mqtt5ClientTest.java
```diff
@@ -310,13 +227,12 @@ public void New_UC1() {
...
     public void New_UC2() {
         skipIfNetworkUnavailable();
-        Assume.assumeTrue(mqtt5DirectMqttHost != null);
-        Assume.assumeTrue(mqtt5DirectMqttPort != null);
-        Assume.assumeTrue(mqtt5BasicAuthPassword != null);
-        Assume.assumeTrue(mqtt5BasicAuthUsername != null);
-        Assume.assumeTrue(mqtt5ProxyHost != null);
-        Assume.assumeTrue(mqtt5ProxyPort != null);
-
+        Assume.assumeTrue(AWS_TEST_MQTT5_DIRECT_MQTT_HOST != null);
```
sbSteveK, line 230: Trivial: I'm pretty sure you can use Assume.assumeNotNull() which takes I believe any number of arguments that would be great for these multiple checks that things aren't null.

src/test/java/software/amazon/awssdk/crt/test/MqttClientConnectionFixture.java
```diff
@@ -306,13 +320,16 @@ void checkOperationStatistics(
                 fail("Incomplete operations size:" + incomplete_ops_size + " did not equal expected value:" + expectedIncompleteOperationSize);
             }
 
-            if (unacked_ops_count != expectedUnackedOperationCount) {
-                fail("Unacked operations count:" + unacked_ops_count + " did not equal expected value:" + expectedUnackedOperationCount);
-            }
+            // While in theory we should be checking the unacked operations as well, it isn't really straight forward
```
sbSteveK, line 323: Do we want to change this function and remove the two arguments that aren't being actually checked (or anything done to them at all) for now and leave a TODO so that if you're looking at the tests themselves you're not giving it these two members expecting something is being done from that perspective?

src/test/java/software/amazon/awssdk/crt/test/ProxyTest.java
```diff
@@ -49,15 +49,13 @@
...
-# AWS_TEST_TLS_CERT_PATH - file path to certificate used to initialize the tls context of the x509 provider connection
-# AWS_TEST_TLS_KEY_PATH - file path to the key used to initialize the tls context of the x509 provider connection
-# AWS_TEST_TLS_ROOT_CERT_PATH - file path to the root CA used to initialize the tls context of the x509 provider connection
-# AWS_TEST_X509_ENDPOINT - AWS account-specific endpoint to source x509 credentials from
-# AWS_TEST_X509_THING_NAME - associated name of the x509 thing
-# AWS_TEST_X509_ROLE_ALIAS - associated role alias ...
-
-# AWS_TEST_IOT_SIGNING_REGION - AWS region to make a websocket connection to
-# AWS_TEST_IOT_MQTT_ENDPOINT - AWS account-specific endpoint to connect to IoT core by
+# AWS_TEST_MQTT311_ROOTCA = The RootCA for connecting
```
sbSteveK, line 52: This appears to be the only place where we provide any kind of description of what these variables represent. Should we remove them from here or move this information to somewhere more meaningful?

sbSteveK (approved):

sbSteveK (approved):

src/test/java/software/amazon/awssdk/crt/test/Mqtt5ClientTest.java
```diff
@@ -1362,69 +1362,6 @@ public void ConnNegativeID_UC7() {
         }
     }
 
-    /* Double Client ID disconnect and then reconnect test */
```
TwistedTwigleg, line 1468: Removing the double client ID reconnect test. It wasn't part of the testing plan anyway and it is **extremely** flakey and inconsistent. We know reconnect works and I don't think it is adding anything but it does cause CI to fail (not only here, but in other PRs too).

We can always add it back if there's a need but for now, I'm just going to remove it.

TwistedTwigleg:

PR #616, opened 2023-05-02
Fix AWS-LC on ManyLinux1 

waahm7:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ on:
       - 'docs'
 
 env:
-  BUILDER_VERSION: v0.9.32
+  BUILDER_VERSION: v0.9.40
   BUILDER_SOURCE: releases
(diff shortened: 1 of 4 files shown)


sbSteveK (approved):

TingDaoK:
should we also remove the hack we did before from pom.xml?

.github/workflows/ci.yml
```diff
@@ -26,8 +26,6 @@ jobs:
         image:
           - manylinux2014-x64
           - manylinux2014-x86
-          - manylinux1-x64
-          - manylinux1-x86
           - al2-x64
```
graebm: why are you removing manylinux1 from CI? I thought we were trying to ensure it keeps working?

waahm7: I added it but it was not working. I tried removing it temporarily, however, this fix is not working for now.

PR #626, opened 2023-05-22
Android Related Cleanup

sbSteveK:
* Run tests in Android CI
* Cleanup of unused files


android/build.gradle
```diff
@@ -4,7 +4,7 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:7.0.0'
+        classpath 'com.android.tools.build:gradle:7.1.2'
```
xiazhvera, line 7: Out of curiosity, is there any reason why we upgrade to 7.1.2? Does the gradle version here related to the Android version?

xiazhvera (approved):

android/build.gradle
sbSteveK: This bump doesn't impact the android version. Other dependencies appear to have used a higher version than we currently had it set to that was causing SDK XML version conflicts.

PR #633, opened 2023-06-02
Init logging

sbSteveK:
* Add ability to initialize logging without needing to set system properties


src/main/java/software/amazon/awssdk/crt/Log.java
```diff
@@ -134,8 +134,17 @@ public static void initLoggingFromSystemProperties() throws IllegalArgumentExcep
...
+
+    /**
+     * Initializes the logging system
+     * @param destinationString determines where logging should be routed to
+     * @param filenameString filename to use if destination for logs is a file
+     * @param levelString detail level of logs that should be shown
+     */
+    public static void initLogging(String destinationString, String filenameString, String levelString){
+        if(destinationString == null && levelString == null){
             return;
```
xiazhvera, line 148: We probably should throw exception or return bool when the argument is invalid.

xiazhvera (approved):
Comments on error handling, while I do realized we never have any return/exception on initialization failure, I feel it would be nice to add one though.

src/main/java/software/amazon/awssdk/crt/Log.java
```diff
@@ -134,8 +134,17 @@ public static void initLoggingFromSystemProperties() throws IllegalArgumentExcep
...
+        initLogging(destinationString, filenameString, levelString);
+    }
+
+    /**
+     * Initializes the logging system
+     * @param destinationString determines where logging should be routed to
+     * @param filenameString filename to use if destination for logs is a file
+     * @param levelString detail level of logs that should be shown
+     */
+    public static void initLogging(String destinationString, String filenameString, String levelString){
```
TingDaoK, line 146: I don't think this is a necessary change. We already have those different functions to do the same thing this function is trying to do. https://github.com/awslabs/aws-crt-java/blob/main/src/main/java/software/amazon/awssdk/crt/Log.java#L174-L197

PR #637, opened 2023-06-08
S3 configs now can take AwsSigningConfig

TingDaoK:
- Add support for s3 client and meta request to take `AwsSigningConfig`. Instead of deprecating the old method with `credentials provider`, just adding the new one and adapt the old one.
- Revamp the `AwsSigningConfig` binding
- Added `getDefaultS3SigningConfig` as helper.
- Added test for special char which Java-SDK has but we don't and catches the issue about different default value between C and Java.

A weird finding:
- After I revamp the test, the filepath test crashes with the assert [here](https://github.com/awslabs/aws-c-s3/blob/v0.3.4/source/s3_auto_ranged_put.c#L292) and fixed by [here](https://github.com/awslabs/aws-c-s3/commit/931d258db4573684cdb6b0c3e73d69689295f70a#diff-1b77484ee8493ab0ab8653a326f4bb0ac42a569e63261469b653ab544d27d99aL292). I don't think it starts to fail because I revamp the test. I am not sure how it bypass the assert before.
- So, I also updated all the submodules to pick up the fix


src/main/java/software/amazon/awssdk/crt/s3/S3Client.java
```diff
@@ -65,12 +67,19 @@ public S3Client(S3ClientOptions options) throws CrtRuntimeException {
             monitoringThroughputThresholdInBytesPerSecond = monitoringOptions.getMinThroughputBytesPerSecond();
             monitoringFailureIntervalInSeconds = monitoringOptions.getAllowableThroughputFailureIntervalSeconds();
         }
+        AwsSigningConfig signingConfig = options.getSigningConfig();
+        boolean created_signing_config = false;
+        if(signingConfig == null) {
```
waahm7, line 72: Trivial: The name `created_signing_config` seems a bit strange to me (you can keep it if you like it). Maybe we could update it to something like `did_create_signing_config` or `is_signing_config_present`, or we could even eliminate this variable entirely and directly use `options.getSigningConfig() == null` check later.

src/main/java/software/amazon/awssdk/crt/s3/S3ClientOptions.java
```diff
@@ -100,6 +108,21 @@ public CredentialsProvider getCredentialsProvider() {
         return credentialsProvider;
     }
 
+    /**
+     * The configuration related to signing used by S3 client. It will override the credentials provider if provided.
+     * `AwsSigningConfig.getDefaultS3SigningConfig(region, credentialsProvider);` can be used as helper to create the default configuration to be used for S3.
+     * @param signingConfig configuration related to signing via an AWS signing process.
+     * @return this
+     */
```
waahm7, line 116: debatable: I don’t like the idea of having multiple ways to do things and priority is defined by the order in which we added them. What happens if we need to add a third way in future? We should just error out if multiple things are set.

src/main/java/software/amazon/awssdk/crt/s3/S3Client.java
```diff
@@ -130,9 +142,13 @@ public S3MetaRequest makeMetaRequest(S3MetaRequestOptions options) {
             requestFilePath = options.getRequestFilePath().toString().getBytes(UTF8);
         }
 
-        long credentialsProviderNativeHandle = 0;
-        if (options.getCredentialsProvider() != null) {
-            credentialsProviderNativeHandle = options.getCredentialsProvider().getNativeHandle();
+        AwsSigningConfig signingConfig = options.getSigningConfig();
+        boolean created_signing_config = false;
+        if(signingConfig == null) {
+            if (options.getCredentialsProvider() != null) {
```
DmitriyMusatkin, line 148: what happens if cred provider is null? no default signing config?
why is this flow different from make client?

src/test/java/software/amazon/awssdk/crt/test/S3ClientTest.java
```diff
@@ -554,7 +553,16 @@ public void onFinished(S3FinishedResponseContext context) {
                 }
             };
 
-            final ByteBuffer payload = ByteBuffer.wrap(createTestPayload(1024 * 1024));
+            HttpHeader[] headers = {
+                    new HttpHeader("Host", ENDPOINT),
+                };
+            HttpRequest httpRequest;
+            String path = "/put_object_test_10MB.txt";
+            if(objectPath!=null) {
```
DmitriyMusatkin, line 561: nit: white space. or you can just do a ternary assignment

src/main/java/software/amazon/awssdk/crt/s3/S3Client.java
TingDaoK: meta request options allow empty signing config as the client level signing config will be used. I added documentation to the options method.

src/main/java/software/amazon/awssdk/crt/s3/S3ClientOptions.java
TingDaoK: I marked the credentials provider as deprecated to keep it consist with our C level, and as we have the helper function to set up the default signing config, is not that bad

src/main/java/software/amazon/awssdk/crt/s3/S3Client.java
TingDaoK: misunderstood your question, fixed.

PR #643, opened 2023-06-21
WIP | Musl support 

waahm7:
*Description of changes:*
- Removes Alpine profile and different Surefire version, that is not needed. 
- Fix musl build script to correctly mount the source and use that instead of pulling from Github.
- Keep previous cross compile jobs as they use glibc.
- Add CI and CD scripts for armv7 and armv8.

Todo:
- Update crt-builder to keep the Qemu image up-to-date.


codebuild/cd/musl-linux-build.sh
```diff
@@ -0,0 +1,29 @@
...
+CLASSIFIER=$1
+shift
+
+# Pry the builder version this CRT is using out of ci.yml
+BUILDER_VERSION=$(cat .github/workflows/ci.yml | grep 'BUILDER_VERSION:' | sed 's/\s*BUILDER_VERSION:\s*\(.*\)/\1/')
+echo "Using builder version ${BUILDER_VERSION}"
+
+aws ecr get-login-password | docker login 123124136734.dkr.ecr.us-east-1.amazonaws.com -u AWS --password-stdin
+export DOCKER_IMAGE=123124136734.dkr.ecr.us-east-1.amazonaws.com/${IMAGE_NAME}:${BUILDER_VERSION}
+export QEMU_IMAGE=123124136734.dkr.ecr.us-east-1.amazonaws.com/multiarch-qemu-user-static:latest
```
waahm7, line 16: Docker has a pull limit for users, and we have started running into that limit. So, we are using one from ECR instead.

PR #646, opened 2023-07-06
Blank line change to make a new release

waahm7:
*Description of changes:*
- Previous release got some conflicts with the Musl binaries I am running tests on because of an error I made with the tag. Need to make a new release.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -216,6 +216,7 @@
                 </plugins>
             </build>
         </profile>
+        
         <profile>
             <id>release</id>
             <distributionManagement>


pom.xml
```diff
@@ -216,6 +216,7 @@
                 </plugins>
             </build>
         </profile>
+
```
graebm, line 219:
```suggestion
```

If you delete this line, you'll still get a unique git commit hash, and the file won't have actually changed 😈
All you need is a unique commit hash to fool our distribution pipeline, it's fine if every file is identical

but next time: `git commit --allow-empty`

graebm (approved):
Commit my suggestion and shipit

pom.xml
waahm7: Nice, now I don't need to make blank line changes to trigger CIs in the future. I have cut a release by creating a new tag for the same commit, which went through.

PR #655, opened 2023-08-07
Close Native Static Resources on JVM Shutdown

waahm7:
*Description of changes:*
The good thing about the static resources is that they can be shared among multiple S3 clients, etc. However, currently, they always leak because we never close them once created. Close them on JVM shutdown. Please suggest if there is a better place to close these resources instead.


src/main/java/software/amazon/awssdk/crt/CRT.java
```diff
@@ -54,6 +58,9 @@ public final class CRT {
         {
             public void run()
             {
+                ClientBootstrap.closeStaticDefault();
```
DmitriyMusatkin, line 61: is it safe to do in this order? requests can still come after this method and will result in those resources being recreated

waahm7: Thanks, updated it to after JVM shutdown.

DmitriyMusatkin (approved):
LGTM. but java test seems to be failing

Issue #668, opened 2023-08-30
deadlock in class initialization of CRT and CrtResource


steven-aerts:
### Describe the bug

In our production code, we have a metrics thread which calls the `CRT` class for things like `CRT#getPlatformImpl` or `CRT#nativeMemory`, while other threads are using the crt library which heavily depends on the `CrtResource`.
During class initialization we are sporadically hitting a deadlock where a thread initializes `CRT` while another one is initializing `CrtResource`.  

Both threads end up waiting for each other, which is clearly visible in the detailed stack dumps:
```
"thread-3" #23 prio=5 os_prio=0 cpu=68.51ms elapsed=2834.21s tid=0x00007f19cc0bec10 nid=0xcbce7 in Object.wait()  [0x00007f19d2ffc000]
   java.lang.Thread.State: RUNNABLE
        at software.amazon.awssdk.crt.CRT.awsCrtInit(Native Method)
        - waiting on the Class initialization monitor for software.amazon.awssdk.crt.CrtResource
        at software.amazon.awssdk.crt.CRT.<clinit>(CRT.java:46)
       ...
"thread-1" #21 prio=5 os_prio=0 cpu=143.28ms elapsed=2834.35s tid=0x00007f19c80d5d10 nid=0xcbce5 in Object.wait()  [0x00007f1a2022a000]
   java.lang.Thread.State: RUNNABLE
        at software.amazon.awssdk.crt.CrtResource.<clinit>(CrtResource.java:104)
        - waiting on the Class initialization monitor for software.amazon.awssdk.crt.CRT
        at software.amazon.awssdk.http.crt.AwsCrtAsyncHttpClient.<init>(AwsCrtAsyncHttpClient.java:96)
        at software.amazon.awssdk.http.crt.AwsCrtAsyncHttpClient.<init>(AwsCrtAsyncHttpClient.java:72)
        at software.amazon.awssdk.http.crt.AwsCrtAsyncHttpClient$DefaultBuilder.build(AwsCrtAsyncHttpClient.java:393)
        at software.amazon.awssdk.http.crt.AwsCrtAsyncHttpClient.create(AwsCrtAsyncHttpClient.java:141)
        ...
```

### Expected Behavior

That those classes can be loaded independently from each other.

### Current Behavior

Class initialisation deadlock.

### Reproduction Steps

You can reproduce this problem reliably with the following code snippet, you hit the problem typically already in the first few iterations:

```java
[...]

PR #670, opened 2023-08-31
Lazy cache jni ids on all Java -> native functions

bretambrose:
https://github.com/awslabs/aws-crt-java/issues/668

Proposed solution:

Lazily cache all JNI Ids on entry from Java -> native.  Use thread-once semantics to make sure it gets done once and only once and subsequent calls wait for the first to complete.  We could do the same thing with a rw lock, but this is easier and less error-prone.

Remove JNI caching from awsCrtInit.

Moving forward, new Java -> native functions should also cache ids at their start.  I don't feel like this is a major gotcha since new Java -> native functions tend to be made by copy paste due to the name mangling and jni type signatures involved.

Includes a test that (if run by itself only) would deadlock with the old approach and does not deadlock with the new approach.


JonathanHenson (approved):

src/native/java_class_ids.h
```diff
@@ -896,6 +896,6 @@ struct java_boxed_array_list_properties {
 };
 extern struct java_boxed_array_list_properties boxed_array_list_properties;
 
-void cache_java_class_ids(JNIEnv *env);
+void aws_cache_jni_ids(JNIEnv *env);
```
graebm, line 899:
```suggestion
/**
 * All functions bound to JNI MUST call this before doing anything else.
 * This caches all java class IDs the first time it is called, any further calls are no-op, it is thread-safe.
 * The reason we do this lazily, rather than simply calling it once from awsCrtInit(),
 * is to avoid deadlock when multiple threads init the CRT simultaneously.
 * See: https://github.com/awslabs/aws-crt-java/pull/670
 */
void aws_cache_jni_ids(JNIEnv *env);
```

graebm (approved):

src/native/java_class_ids.h
bretambrose: Perfect; thorough and conscientious.

DmitriyMusatkin (approved):

src/native/java_class_ids.h
graebm: It never occurred to me before today to have code link its own PR 🔃🤪🔃

JonathanHenson: Probably a good new best practice.

steven-aerts (approved):

Issue #561, opened 2022-12-06
CRT doesn’t handle graceful shutdown properly


nicolas-laduguie:
Describe the bug
We are using AWS Java CRT in the context of AWS sdk V2 with transfer manager for files uploads.
When our Spring boot application is shutdown gracefully, if a multipart upload is started through TransferManager (developer preview) during the graceful shutdown, the transfer is initiated but gets stuck.

Expected Behavior

A multipart upload through TransferManager should finish successfully even though it's started during a graceful shutdown.

Current Behavior

The multipart upload starts as we can see logs like :

```
2022-11-30 14:34:35.205  INFO 116 --- [rding-upload-14] s.a.a.t.s.p.LoggingTransferListener      : [] Transfer initiated...
2022-11-30 14:34:35.208  INFO 116 --- [rding-upload-14] s.a.a.t.s.p.LoggingTransferListener      : [] |                    | 0.0%
[...]

PR #672, opened 2023-09-01
WIP proposal for manual control of CRT shutdown

bretambrose:
https://github.com/awslabs/aws-crt-java/issues/561

Tentative proposal for API review/feedback.


src/main/java/software/amazon/awssdk/crt/CRT.java
```diff
@@ -443,6 +440,55 @@ private static void jvmInit() {
         }
     }
 
+    private static int shutdownRefCount = 1;
+
+    /**
+     * Public API that allows a user to indicate interest in controlling the CRT's time of shutdown.  The
```
graebm, line 446: So the issue is: CRT registers a shutdown hook. But CRT doesn't know what OTHER systems (e.g. Spring) might try to use CRT during their shutdown hooks.

To use this new system, someone using Spring would need to register a shutdown hook after Spring has registered its Spring's shutdown hook, and from there they can call CRT.releaseShutdownRef().

How would the user know when Spring has registered its shutdown hook? Does Spring have an explicit init() function that causes it to get registered?

bretambrose: While it's part of the motivating use case, I don't think Spring or whatever shutdown hook API it may or may not have is relevant to the overall general issue.

graebm: What if we used [JNI_Unload](https://docs.oracle.com/javase/7/docs/technotes/guides/jni/spec/invocation.html) hook, instead of `Runtime.addShutdownHook()`, to register that the JVM is going away? Maybe that would happen after the normal shutdown hooks when the normal shutdown hooks are still using CRT?

graebm: also, in my googling I learned that the normal shutdown hooks all run concurrently on separate threads. It's not a sequential thing like C's atexit: https://docs.oracle.com/javase/8/docs/technotes/guides/lang/hook-design.html

bretambrose: Yeah I think we realized that last time we looked at this.  I'll look into JNI_OnUnload but that's a much more drastic change from current state (even if code wide it doesn't end up being much)

bretambrose: Primary concern is that just like finalizers, I doubt there's any guarantee that the CRT is going to get GCd inbetween the time the user's shutdown hook stops using the CRT and the JVM shuts down for good.

graebm: yeah I wish I could find an official way to hook in later.
I found an unofficial way, the way that `File.deleteOnExit()` uses to clean up after the shutdown hooks. It uses internal APIs. Can we use reflection to see if these APIs exist, and use them if so?

https://github.com/corretto/corretto-8/blob/790d5c6cdb98dcbd196bf3b1140964aee5cdb9f4/src/jdk/src/share/classes/java/io/DeleteOnExitHook.java#L39-L47

graebm (approved):

src/main/java/software/amazon/awssdk/crt/CRT.java
bretambrose: Adding for posterity: OnUnload isn't guaranteed to be called on shutdown and so we'll go with this for now; possibly reevaluate later if there are further issues.

DmitriyMusatkin: my 2 cents on it is that we should not be hooking into jvm by default at all and instead have an explicit way to release all resources. From a library perspective its just not a good thing to do to try to figure out how jvm will behave. If people want they can create their own hook for releasing crt (we can even provide them a helper for doing it).

How much benefit are we actually getting from the shutdown hook? is it actually releasing any resources that would be dangling? or we are just releasing some mem that would go away with the shutdown anyway?

bretambrose: > How much benefit are we actually getting from the shutdown hook? is it actually releasing any resources that would be dangling? or we are just releasing some mem that would go away with the shutdown anyway?

We are not seg-faulting when native threads attempt to get a jni env and instead get garbage.

DmitriyMusatkin: that only happens if we ran crt shutdown hook and then something else tried to use crt? if we never release crt resources during shutdown, all the calls to crt should work fine?

bretambrose: Anything that is still going on in an event loop thread that attempts to call back into Java during JVM shutdown is a crash risk.  All it takes is launching an async request of some sort and forgetting about it.

PR #674, opened 2023-09-04
Mqtt5 Adapter Binding

xiazhvera:
*Issue #, if available:*

*Description of changes:*


bretambrose (changes requested):

src/main/java/software/amazon/awssdk/crt/mqtt5/Mqtt5Client.java
```diff
@@ -202,6 +227,48 @@ private synchronized void setIsConnected(boolean connected) {
...
+     * Mqtt5 to Mqtt3 Adapter
+     ******************************************************************************/
+
+    /**
+     * Returns a MqttConnection Object based on Mqtt5Client configuration. A client id is required for NewConnection.
+     * @param callbacks connection event callbacks to use
+     *
+     * @return a MqttClientConnection
+     */
+    public MqttClientConnection NewConnection(MqttClientConnectionEvents callbacks) throws MqttException
```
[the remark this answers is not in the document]
xiazhvera: Added a constructor in MqttConnection. However, as MqttConnection and Mqtt5Client is not in the same package, it is not possible to use a package-private config, instead I add a `getClientOptions` for Mqtt5Client so that the MqttConnection could access the config.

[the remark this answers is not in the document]
xiazhvera: Also, after thinking about it. This is a better workflow as it reduced the chance that the Mqtt5Client user to access the Mqtt3 interface. I will also update it for CPP and Python.

src/main/java/software/amazon/awssdk/crt/mqtt5/Mqtt5Client.java
```diff
@@ -46,6 +46,16 @@ public class Mqtt5Client extends CrtResource {
...
 
+    /**
+     * A private config used to save config for mqtt3 connection creation
+     */
+    private Mqtt5ClientOptions clientOptions;
+
+    /**
+     * A private tlscontext
+     */
+    private TlsContext tlsContext;
```
bretambrose, line 57: Isn't this from clientOptions anyways and can be removed?

src/native/mqtt_connection.c
```diff
@@ -82,25 +83,20 @@ static void s_mqtt_jni_connection_release(struct mqtt_jni_connection *connection
     size_t old_value = aws_atomic_fetch_sub(&connection->ref_count, 1);
 
     AWS_LOGF_DEBUG(AWS_LS_MQTT_CLIENT, "mqtt_jni_connection release, ref count now = %d", (int)old_value - 1);
+}
 
-    if (old_value == 1) {
+/* The destroy function is called on Java MqttClientConnection resource release. */
+static void s_mqtt_jni_connection_destory(struct mqtt_jni_connection *connection) {
```
bretambrose, line 89: destroy

src/native/mqtt_connection.c
```diff
@@ -434,17 +498,10 @@ static void s_on_shutdown_disconnect_complete(struct aws_mqtt_client_connection
...
-    jobject mqtt_connection = (*env)->NewLocalRef(env, jni_connection->java_mqtt_connection);
-    if (mqtt_connection != NULL) {
-        (*env)->CallVoidMethod(env, mqtt_connection, crt_resource_properties.release_references);
-
-        (*env)->DeleteLocalRef(env, mqtt_connection);
-
-        aws_jni_check_and_clear_exception(env);
-    }
+    /* Release the underlying mqtt connection */
+    aws_mqtt_client_connection_release(jni_connection->client_connection);
```
bretambrose, line 502: With these changes, we shouldn't even need to acquire the env anymore.  This function should just be a aws_mqtt_client_connection_release call.

src/native/mqtt_connection.c
```diff
@@ -82,25 +83,20 @@ static void s_mqtt_jni_connection_release(struct mqtt_jni_connection *connection
...
-        if (aws_mqtt_client_connection_disconnect(
-                connection->client_connection, s_on_shutdown_disconnect_complete, connection) != AWS_OP_SUCCESS) {
-
-            /*
-             * This can happen under normal code paths if the client happens to be disconnected at cleanup/shutdown
-             * time. Log it (in case it was unexpected) and then invoke the shutdown callback manually.
-             */
-            s_on_shutdown_disconnect_complete(connection->client_connection, connection);
-        }
+        aws_mqtt_client_connection_release(connection->client_connection);
```
bretambrose, line 99: This should be a call to s_on_shutdown_disconnect_complete.  While that function should only be doing this release, if you call an async api with a callback and it gives an error (thus not invoking the callback), the pattern should be to call that callback yourself rather than repeating the logic inside of it.

src/main/java/software/amazon/awssdk/crt/mqtt5/Mqtt5Client.java
```diff
@@ -94,6 +106,11 @@ public Mqtt5Client(Mqtt5ClientOptions options) throws CrtRuntimeException {
         isConnected = false;
     }
 
+    /**
+     * @return the tls context used by all connections associated with this client.
+     */
+    public TlsContext getTlsContext() { return tlsContext; }
```
bretambrose, line 112: Seems unnecessary given that I can get the 5 options.

src/main/java/software/amazon/awssdk/crt/mqtt/MqttClientConnection.java
```diff
@@ -49,14 +52,45 @@ void deliver(String topic, byte[] payload, boolean dup, int qos, boolean retain)
...
+                                : 0));
+            }
+            options.setCleanSession(
+                    mqtt5options.getSessionBehavior().compareTo(Mqtt5ClientOptions.ClientSessionBehavior.CLEAN) <= 0);
+            options.setPingTimeoutMs(
+                    Math.toIntExact(mqtt5options.getPingTimeoutMs() != null ? mqtt5options.getPingTimeoutMs() : 0));
+            options.setProtocolOperationTimeoutMs(Math.toIntExact(
+                    mqtt5options.getAckTimeoutSeconds() != null ? mqtt5options.getAckTimeoutSeconds() : 0) * 1000);
+            return options;
+        } catch (ArithmeticException e) {
```
bretambrose, line 79: Still not seeing the value in rethrowing an exception that wraps the same exception type.

bretambrose (approved):

PR #675, opened 2023-09-05
Use properties for Unit Tests

sbSteveK:
Change unit tests to setup using system properties instead of pulling directly from environment variables.

Android does not support environment variables in a way that can be used to run unit tests.


graebm:
trivial: In the PR description, explain WHY you're making these changes

bretambrose (approved):

PR #703, opened 2023-10-24
Cleanup CrtS3RuntimeException

waahm7:
*Description of changes:*
- Remove XML parsing from CrtS3RuntimeException; it's error-prone, especially with S3 where there's a lot of customization and no fixed API contract. Expose errorPayload directly as is. It's a breaking change, but not many people should be using it for now.


TingDaoK (commented):
Do we still need this separate exception class? 
Can we use `S3FinishedResponseContext` directly instead?

PR #706, opened 2023-10-27
Android library loading fix

sbSteveK:

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/CRT.java
+++ b/src/main/java/software/amazon/awssdk/crt/CRT.java
@@ -300,8 +300,16 @@ public final class CRT {
             }
 
             // open a stream to read the shared lib contents from this JAR
+            String libResourcePath = "/" + os + "/" + getArchIdentifier() + "/" +  getCRuntime(os) + "/" + libraryName;
+            // Check whether there is a platform specific resource path to use
             CrtPlatform platform = getPlatformImpl();
-            String libResourcePath = platform != null ? platform.getResourcePath(getCRuntime(os), libraryName) : "/" + os + "/" + getArchIdentifier() + "/" +  getCRuntime(os) + "/" + libraryName;
+            if (platform != null){
+                String platformLibResourcePath = platform.getResourcePath(getCRuntime(os), libraryName);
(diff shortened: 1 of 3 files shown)


src/main/java/software/amazon/awssdk/crt/CRT.java
```diff
@@ -300,7 +300,8 @@ private static void extractAndLoadLibrary(String path) {
             }
 
             // open a stream to read the shared lib contents from this JAR
-            String libResourcePath = "/" + os + "/" + getArchIdentifier() + "/" +  getCRuntime(os) + "/" + libraryName;
+            CrtPlatform platform = getPlatformImpl();
+            String libResourcePath = platform != null ? platform.getResourcePath(getCRuntime(os), libraryName) : "/" + os + "/" + getArchIdentifier() + "/" +  getCRuntime(os) + "/" + libraryName;
```
bretambrose, line 304: When would platform be null?

sbSteveK: if `findPlatformImpl()` doesn't find CrtPlatformImpl in an OS specified location it will be null. As of right now, Android should be the only platform that implements it outside of testing environments.

bretambrose: Well, I'm not a fan of the fact that the default path is implemented twice.

sbSteveK: Changed it. It's a little messier but there's only a singular default path.

sbSteveK: just a note: I looked into some way of combining and using different ways to use existing functions but the problem was that android does so many very android specific different things in terms of `getOSIdentifier()` for this path, `getArchIdentifier()`, and omitting `getCRuntime()` entirely that it felt best to add this one filepath function in one place rather than peppering android specific things across `CRT.java` and/or introducing multiple platform specific functions and branches to OS/Arc/Runtime identifiers. Open to suggestions on it tho if you have any ideas.

bretambrose (approved):

PR #721, opened 2023-11-14
Mqtt5 Canary Update

xiazhvera:
*Issue #, if available:*

*Description of changes:*
1. Fix sleep time for TPS
2. Use random payload size and data for Published payload
3. Do not exit earlier when operation failed

### DIFF
--- a/utils/Canary/src/main/java/canary/mqtt5/Mqtt5Canary.java
+++ b/utils/Canary/src/main/java/canary/mqtt5/Mqtt5Canary.java
@@ -76,27 +76,29 @@ public class Mqtt5Canary {
 
     static int operationFutureWaitTime = 30;
 
+    static int MAX_PAYLOAD_SIZE = 65535; // Use UINT64_MAX for the payload size
+
     static void printUsage() {
         System.out.println(
-            "Usage:\n" +
-            "  --help            This message\n"+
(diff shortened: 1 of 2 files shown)


utils/Canary/src/main/java/canary/mqtt5/Mqtt5Canary.java
```diff
@@ -364,8 +370,6 @@ public void accept(Mqtt5WebsocketHandshakeTransformArgs t) {
 
     public static void setupOperations() {
         // For now have everything evenly distributed
-        clientsOperationsList.add(CANARY_OPERATIONS.OPERATION_NULL);
-        clientsOperationsList.add(CANARY_OPERATIONS.OPERATION_START);
```
sbSteveK, line 368: Just curious as to why we're removing these.

sbSteveK (approved):

utils/Canary/src/main/java/canary/mqtt5/Mqtt5Canary.java
xiazhvera: NULL:   The operation would do nothing, therefore I just remove it. 
START: I think it might be not necessary for a specific START operation, as we will always start the connection after the client stopped (There is a STOP operation in the list). If the client is stopped, then before the client does any other operation, it would start the client first.

PR #723, opened 2023-11-21
Fix mqtt5to3adapter suback failure handling

sfod:
*Description of changes:*

MQTT3 binding code decides if subscription succeeded or not only by error_code in SUBACK packet while ignoring QoS field. For MQTT3 connection it's probably OK since failed subscription causes disconnect anyway. But for MQTT5to3Adapter no disconnect happening, so in some scenarios when error_code is 0 and QoS is failure (e.g. Unauthorized QoS) user code receives success on an actually failed subscription.

The proper way to fix the issue is to return QoS to users, so they can handle errors. However, `connection.subscribe()` returns `CompletableFuture<Integer>` (Integer is for MQTT packet ID), so there is no way to achieve this without making breaking changes. Instead, on a failed QoS, an exception is set to the returned `CompletableFuture`.

### DIFF
--- a/src/native/mqtt_connection.c
+++ b/src/native/mqtt_connection.c
@@ -802,6 +802,10 @@ static void s_deliver_suback_failure(
     }
 }
 
+static bool s_is_qos_successful(enum aws_mqtt_qos qos) {
+    return qos < 128;
+}
+
 static void s_on_ack(
     struct aws_mqtt_client_connection *connection,
     uint16_t packet_id,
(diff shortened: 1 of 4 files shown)


src/native/mqtt_connection.c
```diff
@@ -759,6 +759,53 @@ static void s_on_op_complete(
...
+
+    AWS_FATAL_ASSERT(callback);
+    AWS_FATAL_ASSERT(callback->connection);
+    AWS_FATAL_ASSERT(env);
+
+    if (callback->async_callback) {
+        jobject jni_reason;
+        if (error_code != 0) {
+            jni_reason = s_new_mqtt_exception(env, error_code);
+        } else {
```
bretambrose, line 788: Let's add an error code instead of printing a custom string.

bretambrose (approved):

Issue #716, opened 2023-11-12
JVM Crash: Fatal Error - Issue with Native Method s3ClientMakeMetaRequest in S3Client


kcsurapaneni:
### Describe the bug

Encountered a fatal error when creating an `S3AsyncClient`, configuring it, closing it immediately, and then attempting to perform an asynchronous `putObject` operation. The crash is triggered, resulting in a JVM crash.

### Expected Behavior

Should gracefully handle the scenario where a `putObject` operation is attempted on a closed `S3AsyncClient` without causing a fatal error in the JVM.

### Current Behavior

The JVM crashes with an error in the native method `software.amazon.awssdk.crt.s3.S3Client.s3ClientMakeMetaRequest` when attempting to perform a `putObject` operation on a `S3AsyncClient` that has been closed.

### Reproduction Steps

1. Creates an asynchronous `S3AsyncClient` with specific configuration settings.
2. Creates a `PutObjectRequest` specifying the destination bucket and key for the S3 object.
3. Explicitly close the `S3AsyncClient` immediately after creation.
4. Attempt to asynchronously upload an object to S3 using the closed `S3AsyncClient`.

Observe the JVM crash and review the logs here in the sample project (https://github.com/kcsurapaneni/java-aws-sdk) available on GitHub for detailed information.

### Possible Solution

Ensure that the library gracefully handles scenarios where operations are attempted on a closed client.

### Additional Information/Context

Relevant code snippet

```
S3AsyncClient s3 = S3AsyncClient.crtBuilder()
                .region(Region.US_WEST_2)
                .targetThroughputInGbps(100.0)
                .minimumPartSizeInBytes(8 * MB)
                .build();

PutObjectRequest putOb = PutObjectRequest.builder()
                .bucket("temp-bucket")
                .key("uploads")
                .build();

s3.close();

CompletableFuture<PutObjectResponse> putObjectResponseCompletableFuture = s3.putObject(putOb, AsyncRequestBody.fromBytes("byteArr".getBytes()));

```

### aws-crt-java version used

0.28.6

### Java version used

1.8(Amazon Coretto), 17(Oracle, Zulu, Amazon Coretto)

### Operating System and version

macOS 14.1

PR #725, opened 2023-11-22
Gracefully handle calling make_meta_request on closed S3 client

waahm7:
*Issue #, if available:*
https://github.com/awslabs/aws-crt-java/issues/716

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/s3/S3Client.java
+++ b/src/main/java/software/amazon/awssdk/crt/s3/S3Client.java
@@ -125,19 +125,19 @@ public class S3Client extends CrtResource {
         if(isNull()) {
             Log.log(Log.LogLevel.Error, Log.LogSubject.S3Client,
                     "S3Client.makeMetaRequest has invalid client. The client can not be used after it is closed.");
-            return null;
+            throw new IllegalStateException("S3Client.makeMetaRequest has invalid client. The client can not be used after it is closed.");
         }
 
         if (options.getHttpRequest() == null) {
             Log.log(Log.LogLevel.Error, Log.LogSubject.S3Client,
                     "S3Client.makeMetaRequest has invalid options; Http Request cannot be null.");
-            return null;
+            throw new IllegalArgumentException("S3Client.makeMetaRequest has invalid options; Http Request cannot be null.");
         }
 
         if (options.getResponseHandler() == null) {
             Log.log(Log.LogLevel.Error, Log.LogSubject.S3Client,
                     "S3Client.makeMetaRequest has invalid options; Response Handler cannot be null.");
-            return null;
+            throw new IllegalArgumentException("S3Client.makeMetaRequest has invalid options; Response Handler cannot be null.");
         }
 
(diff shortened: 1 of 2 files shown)


src/main/java/software/amazon/awssdk/crt/s3/S3Client.java
```diff
@@ -122,6 +122,12 @@ private void onShutdownComplete() {
 
     public S3MetaRequest makeMetaRequest(S3MetaRequestOptions options) {
 
+        if(isNull()) {
+            Log.log(Log.LogLevel.Error, Log.LogSubject.S3Client,
+                    "S3Client.makeMetaRequest has invalid client. The client can not be used after it is closed.");
+            return null;
```
graebm, line 128: sins of the past: why is this function returning null instead of throwing exceptions?

is this C 💁‍♂️🦋 ?

src/main/java/software/amazon/awssdk/crt/s3/S3Client.java
```diff
@@ -122,6 +122,12 @@ private void onShutdownComplete() {
 
     public S3MetaRequest makeMetaRequest(S3MetaRequestOptions options) {
 
+        if(isNull()) {
```
graebm, line 125: The S3Client is intended to be a "thread safe" class, but I don't think there's any synchronization going on between `makeMetaRequest()` and `close()`

src/main/java/software/amazon/awssdk/crt/s3/S3Client.java
waahm7: Thanks, fixed.

src/main/java/software/amazon/awssdk/crt/s3/S3Client.java
waahm7: Looking at the other places, this is a common problem that we don't have any synchronization for close function. 

- https://github.com/awslabs/aws-crt-java/blob/a8bec8665c547eaadcc00388a6e2aab1faa3eb1e/src/main/java/software/amazon/awssdk/crt/http/HttpClientConnection.java#L44
- https://github.com/awslabs/aws-crt-java/blob/a8bec8665c547eaadcc00388a6e2aab1faa3eb1e/src/main/java/software/amazon/awssdk/crt/http/Http2ClientConnection.java#L187-L187
- https://github.com/awslabs/aws-crt-java/blob/a8bec8665c547eaadcc00388a6e2aab1faa3eb1e/src/main/java/software/amazon/awssdk/crt/eventstream/ClientConnection.java#L109
- https://github.com/awslabs/aws-crt-java/blob/a8bec8665c547eaadcc00388a6e2aab1faa3eb1e/src/main/java/software/amazon/awssdk/crt/http/HttpStreamBase.java#L97
- https://github.com/awslabs/aws-crt-java/blob/e50af06926946768aaa1e847b38298990280e3db/src/main/java/software/amazon/awssdk/crt/io/TlsKeyOperation.java#L139

We can solve this problem when we encounter it.

graebm (approved):

PR #726, opened 2023-11-24
Add test for shared subscriptions

sfod:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/test/java/software/amazon/awssdk/crt/test/Mqtt5ClientTestFixture.java
+++ b/src/test/java/software/amazon/awssdk/crt/test/Mqtt5ClientTestFixture.java
@@ -11,6 +11,7 @@ import software.amazon.awssdk.crt.mqtt5.packets.*;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.HashMap;
 
 import java.util.concurrent.CompletableFuture;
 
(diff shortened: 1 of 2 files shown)


src/test/java/software/amazon/awssdk/crt/test/Mqtt5ClientTestFixture.java
```diff
@@ -144,25 +147,38 @@ public void onMessageReceived(Mqtt5Client client, PublishReturn result) {
...
         public void onMessageReceived(Mqtt5Client client, PublishReturn result) {
             currentPublishCount += 1;
             if (currentPublishCount == desiredPublishCount) {
                 publishReceivedFuture.complete(null);
+                // An optional wait to ensure that no packets beyond expectations are arrived.
+                TIMEOUT_SCHEDULER.schedule(
+                        () -> afterCompletionFuture.complete(null), afterCompletionWaitSec, TimeUnit.SECONDS);
             } else if (currentPublishCount > desiredPublishCount) {
-                publishReceivedFuture.completeExceptionally(new Throwable("Too many publish packets received"));
+                afterCompletionFuture.completeExceptionally(new Throwable("Too many publish packets received"));
```
sfod, line 172: This block did not work as intended prior to this PR, as `publishReceivedFuture` is already completed at this point (unless `desiredPublishCount == 0`, but it's not the case here).
I decided to fix it, though maybe it's better to remove this check for good, because this class is used in two tests: "Happy path. No drop in connection, no retry, no reconnect" and "Shared subscriptions", where this functionality is extra.

xiazhvera (approved):

PR #733, opened 2023-12-13
To support VSOCK, stop limiting port to 16 bits.

graebm:
Issue: https://github.com/awslabs/aws-c-io/issues/576


src/native/event_stream_rpc_client.c
```diff
@@ -221,7 +221,7 @@ jint JNICALL Java_software_amazon_awssdk_crt_eventstream_ClientConnection_client
         .socket_options = socket_options,
         .tls_options = conn_options_ptr,
         .user_data = callback_data,
-        .port = (uint16_t)port,
+        .port = (uint32_t)port,
```
waahm7, line 224: I don't think this is safe since jint can be negative.

src/native/http2_stream_manager.c
```diff
@@ -155,8 +155,8 @@ JNIEXPORT jlong JNICALL Java_software_amazon_awssdk_crt_http_Http2StreamManager_
 
     struct aws_byte_cursor endpoint = aws_jni_byte_cursor_from_jbyteArray_acquire(env, jni_endpoint);
 
-    if (jni_port <= 0 || 65535 < jni_port) {
-        aws_jni_throw_illegal_argument_exception(env, "Port must be between 1 and 65535");
+    if (jni_port == 0) {
```
waahm7, line 158: Same, we were checking for negative values before and now we are not.

src/native/http_connection_manager.c
```diff
@@ -141,11 +141,6 @@ JNIEXPORT jlong JNICALL Java_software_amazon_awssdk_crt_http_HttpClientConnectio
     struct aws_allocator *allocator = aws_jni_get_allocator();
     struct aws_byte_cursor endpoint = aws_jni_byte_cursor_from_jbyteArray_acquire(env, jni_endpoint);
 
-    if (jni_port <= 0 || 65535 < jni_port) {
```
waahm7, line 144: No longer safe for negative values. Same comment for all the java to native conversion.

src/native/event_stream_rpc_client.c
graebm: It's safe in that -1 becomes 0xFFFFFFFF
The [VSOCK docs](https://man.archlinux.org/man/vsock.7.en#Address_format) even use language like:
> There are several special addresses: VMADDR_CID_ANY (-1U) means any address for binding

which reads to me like "32 bits signed unsigned whatever"

waahm7: It's not safe if it's done in a native layer unknown to the user.

graebm: When I google "vsock java", the first hit that comes up is vsockj, which uses `int` for port and `-1` for the ANY port
https://github.com/Cloud-Architects/vsockj/blob/487ebc0c61281bb109a2801ed19a85d3ab653395/vsockj-core/src/main/java/solutions/cloudarchitects/vsockj/VSockAddress.java#L26

graebm: The 2nd hit on google also uses int
https://github.com/kohlschutter/junixsocket/blob/cdb4c1a150222f88e62e025aada742ec3ccabcab/junixsocket-common/src/main/java/org/newsclub/net/unix/AFVSOCKSocketAddress.java#L94

graebm: I'm guessing it also plays well with Java's [URI](https://docs.oracle.com/javase/8/docs/api/java/net/URI.html#getPort--) class which uses int

graebm: Documented:
> For 32bit values exceeding Integer.MAX_VALUE use two's complement (i.e. -1 == 0xFFFFFFFF).

waahm7 (approved):

PR #738, opened 2023-12-21
Add support for metrics in stream response handler

ianbotsf:
*Issue #, if available:* https://github.com/awslabs/smithy-kotlin/issues/893

*Description of changes:* This change adds support for getting metrics as part of the Java-level `HttpStreamResponseHandler`.


src/native/http_request_response.c
```diff
@@ -311,6 +311,51 @@ void aws_java_http_stream_on_stream_destroy_fn(void *user_data) {
...
+        binding->java_http_response_stream_handler,
+        http_stream_response_handler_properties.onMetrics,
+        binding->java_http_stream_base,
+        jni_metrics);
+
+    if (aws_jni_check_and_clear_exception(env)) {
+        /* Close the Connection if the Java Callback throws an Exception */
+        aws_http_connection_close(aws_http_stream_get_connection(stream));
+    }
+
```
TingDaoK, line 354: We need to manually delete the local reference created when it's called from a C thread, `(*env)->DeleteLocalRef(env, jni_metrics);`, otherwise, it will result in a leak. We had the similar leak recently, referring to https://github.com/awslabs/aws-crt-java/pull/734

src/native/http_request_response.c
```diff
@@ -311,6 +311,51 @@ void aws_java_http_stream_on_stream_destroy_fn(void *user_data) {
...
+        env,
+        binding->java_http_response_stream_handler,
+        http_stream_response_handler_properties.onMetrics,
+        binding->java_http_stream_base,
+        jni_metrics);
+
+    if (aws_jni_check_and_clear_exception(env)) {
+        /* Close the Connection if the Java Callback throws an Exception */
+        aws_http_connection_close(aws_http_stream_get_connection(stream));
+    }
```
TingDaoK, line 353: Debatable: I know we have this pattern of error handling from `on_complete` callback, but I don't really like it. Especially as we introduced http/2, seems like a overkill to kill the whole connection, but, TBH, I don't really have better idea about how to handle this.

I would suggest to at least add a warning-level log here.

src/native/http_request_response.c
```diff
@@ -311,6 +311,51 @@ void aws_java_http_stream_on_stream_destroy_fn(void *user_data) {
...
+        http_stream_metrics_properties.http_stream_metrics_class,
+        http_stream_metrics_properties.constructor_id,
+        (jlong)metrics->send_start_timestamp_ns,
+        (jlong)metrics->send_end_timestamp_ns,
+        (jlong)metrics->sending_duration_ns,
+        (jlong)metrics->receive_start_timestamp_ns,
+        (jlong)metrics->receive_end_timestamp_ns,
+        (jlong)metrics->receiving_duration_ns,
+
+        /* Stream IDs are 31-bit unsigned integers, which fits into Java's regular (signed) 32-bit int */
```
TingDaoK, line 339: is this true? Don't we have 32bit unsigned? Why not just use long?

src/native/http_request_response.c
ianbotsf: Looking at other callbacks, it seems the common practice is to log at error level (vs warning) and to raise an error as well so I've done that here. Let me know if you'd strongly prefer this be logged at warning level or that the error not be raised.

src/native/http_request_response.c
ianbotsf: Well it is true that HTTP/2 stream IDs are 31-bit integers according to the spec ([RFC 7540 § 5.1.1](https://httpwg.org/specs/rfc7540.html#rfc.section.5.1.1)). It's also true that they're [treated as unsigned 32-bit by CRT](https://github.com/awslabs/aws-c-http/blob/6a1c157c20640a607102738909e89561a41e91e9/include/aws/http/request_response.h#L228).

I _could_ cast this to `long` but the value can never be outside the range of [0, 2³¹) so it fits inside a Java `int` without risk of loss/corruption. Given that it can fit into an `int` and `int`s are more idiomatic in Java than `long`s, that seems better.

Let me know if you feel strongly and I can switch this.

src/main/java/software/amazon/awssdk/crt/http/HttpStreamBaseResponseHandler.java
```diff
@@ -87,4 +87,12 @@ default int onResponseBody(HttpStreamBase stream, byte[] bodyBytesIn) {
...
     void onResponseComplete(HttpStreamBase stream, int errorCode);
 
+    /**
+     * Called right before stream is complete, whether successful or unsuccessful.
+     * @param stream The HTTP stream to which the metrics apply
+     * @param metrics The [HttpStreamMetrics] containing metrics for the given stream
+     */
+    default void onMetrics(HttpStreamBase stream, HttpStreamMetrics metrics) {
+        /* Optional callback, nothing to do by default */
+    }
```
TingDaoK, line 97: nitpick: can we move the metrics callback before the complete callback? So, that it's the same order as it gets invoked

TingDaoK (commented):
Looks good overall, can we also add a small test for it to just make sure we can successfully get the metrics?

TingDaoK (approved):
:shipit:

PR #739, opened 2023-12-22
Fix overridden timeout values for mqtt5 operations

sfod:
Currently, subscribe, unsubscribe, and publish operations are enqueued with overridden timeouts set to garbage values.
This PR fixes it by explicitly setting the overridden timeout to 0.

### DIFF
--- a/src/native/mqtt5_client.c
+++ b/src/native/mqtt5_client.c
@@ -1203,7 +1203,7 @@ JNIEXPORT void JNICALL Java_software_amazon_awssdk_crt_mqtt5_Mqtt5Client_mqtt5Cl
     return_data->java_client = java_client;
     return_data->jni_publish_future = (*env)->NewGlobalRef(env, jni_publish_future);
 
-    struct aws_mqtt5_publish_completion_options completion_options;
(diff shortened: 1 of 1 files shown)


waahm7 (approved):

src/native/mqtt5_client.c
```diff
@@ -1203,9 +1203,10 @@ JNIEXPORT void JNICALL Java_software_amazon_awssdk_crt_mqtt5_Mqtt5Client_mqtt5Cl
     return_data->java_client = java_client;
     return_data->jni_publish_future = (*env)->NewGlobalRef(env, jni_publish_future);
 
-    struct aws_mqtt5_publish_completion_options completion_options;
```
TingDaoK, line 1206: `AWS_ZERO_STRUCT` can help, or your fix works as well.

TingDaoK (approved):

src/native/mqtt5_client.c
sfod: Good point. But since aws-crt-python and aws-crt-nodejs both use designated initializers for `completion_options` as well, I'll stick to it, just for consistency.

PR #747, opened 2024-01-12
Implement stop method for mqtt5 client with no args

sfod:
Add a `stop` method which won't send DISCONNECT packet.


sbSteveK (approved):

sbSteveK:
We should probably simplify the pubsub in the v2 mqtt5 sample once this gets released and the new crt is pulled into the v2 sdk.

PR #755, opened 2024-02-05
Refact resource

TingDaoK:
- Put rwlock around the part where we access the native handler to prevent the race condition that user freed the native handle, but it's still passed down into the C code and cause the C part to crash

Why rwlock:
- Less critical part comparing to normal lock
- I like we always grab the lock and release it, which is more clear to me comparing to make the function sync, which I still don't know where the lock was released.
- The performance for sync method and rwlock and without any lock is similar using S3.


src/main/java/software/amazon/awssdk/crt/CrtResource.java
```diff
@@ -36,11 +37,11 @@ public abstract class CrtResource implements AutoCloseable {
     /**
      * Debug/diagnostic data about a CrtResource object
      */
-    public class ResourceInstance {
+    public static class ResourceInstance {
```
graebm, line 40: just curious, why? I don't understand why an inner class would be static vs non-static

src/main/java/software/amazon/awssdk/crt/CrtResource.java
```diff
@@ -86,17 +87,18 @@ public String toString() {
...
+    private static final boolean debugNativeObjects = System.getProperty(NATIVE_DEBUG_PROPERTY_NAME) != null;
     private static int resourceCount = 0;
     private static final Lock lock = new ReentrantLock();
     private static final Condition emptyResources  = lock.newCondition();
     private static final AtomicLong nextId = new AtomicLong(0);
+    private static final Instant creationTime = Instant.now();
 
     private final ArrayList<CrtResource> referencedResources = new ArrayList<>();
+    private final ReentrantReadWriteLock rwl = new ReentrantReadWriteLock();
     private long nativeHandle;
```
graebm, line 99: ~rwl~ -> `nativeHandleLock`? maybe name it after the data it's protecting?

src/main/java/software/amazon/awssdk/crt/CrtResource.java
```diff
@@ -175,11 +177,21 @@ protected void swapReferenceTo(CrtResource oldReference, CrtResource newReferenc
...
+        this.rwl.readLock().unlock();
+    }
+
+
     /**
      * Takes ownership of a native object where the native pointer is tracked as a long.
      * @param handle pointer to the native object being acquired
      */
     protected void acquireNativeHandle(long handle) {
+        this.rwl.writeLock().lock();
```
graebm, line 194: Use some mechanism to ensure a random exception doesn't bounce us out of this block before unlock

try with resources
try/finally
something

graebm: The advantage of `synchronized` is: it's dead simple. The only way to mess up is to forget to use it.

If we used `synchronized` we could add CI to catch where we forget to use it. We could like use Java reflection to scan all subclasses of CrtResource, scan all methods, check if [method.isSynchronized()](https://docs.oracle.com/javase%2F9%2Fdocs%2Fapi%2F%2F/com/sun/jdi/Method.html#isSynchronized--), and error if so (maybe there's be an allow-list for functions that didn't need to be synchronized?

TingDaoK:
won't fix it until -v-next.

PR #757, opened 2024-02-06
Fix Log Subject for MQTT5

sbSteveK:
Change Log Subject to MQTT5 in MQTT5 related logs.


sfod (commented):
I think, the log subject should be also updated in `src/native/mqtt5_packets.c` and `src/native/mqtt5_utils.c`, because they use `AWS_LS_MQTT_CLIENT` as well.

PR #758, opened 2024-02-07
Fix duplicated onConnectionSuccess and onConnectionFailure calls

sfod:
Do not call `onConnectionSuccess` and `onConnectionFailure` callbacks for the `on_connection_complete` event as they are called for connection success and connection failure events respectively.

Also, register `on_connection_success` and `on_connection_failure` events for mqtt5to3 adapter.


xiazhvera (approved):
Nice catch on the callback on `newFrom5Client`.

PR #764, opened 2024-02-22
S3: cancel the meta request before release the native handler

TingDaoK:
*Issue #, if available:*

*Description of changes:*

- Cancel the meta request before Java side close on it. If Java side already dropped the meta request, it doesn't make sense to keep the meta request flowing in the native part.
- help to prevent the possible leak when the meta request was closed, but not cancelled, the data will keep flowing and stalled because of backpressure
- Add it to the java part to make it a little bit easier to track (The close method is not easy to track down to the `releaseNativeHandle` though)

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/s3/S3MetaRequest.java
+++ b/src/main/java/software/amazon/awssdk/crt/s3/S3MetaRequest.java
@@ -42,6 +42,7 @@ public class S3MetaRequest extends CrtResource {
              * Cancel the meta request before drop the refcount.
              * The meta request is not referenced by Java any longer, everything from native to Java will be ignored.
              * Cancelling the meta request instead of letting it keep flowing.
+             * Note: If the meta request has not finished yet, it will be finished with `AWS_ERROR_S3_CANCELED`.
              **/
             s3MetaRequestCancel(getNativeHandle());
             s3MetaRequestDestroy(getNativeHandle());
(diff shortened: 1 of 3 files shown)


src/main/java/software/amazon/awssdk/crt/s3/S3MetaRequest.java
```diff
@@ -38,6 +38,12 @@ protected boolean canReleaseReferencesImmediately() {
     @Override
     protected void releaseNativeHandle() {
         if (!isNull()) {
+            /**
+             * Cancel the meta request before drop the refcount.
+             * The meta request is not referenced by Java any longer, everything from native to Java will be ignored.
+             * Cancelling the meta request instead of letting it keep flowing.
+             **/
+            s3MetaRequestCancel(getNativeHandle());
```
zoewangg, line 46: What happens if the request is already cancelled or paused?

TingDaoK: it won't do anything. The [cancel](https://github.com/awslabs/aws-c-s3/blob/main/source/s3_meta_request.c#L351-L358) is idempotent, and the [pause](https://github.com/awslabs/aws-c-s3/blob/main/source/s3_auto_ranged_put.c#L1668-L1677) also fall into the same path with cancel.

DmitriyMusatkin (approved):

PR #772, opened 2024-03-07
FIPS with classifier

TingDaoK:
*Issue #, if available:*

*Description of changes:*


.builder/actions/aws_crt_java_build.py
```diff
@@ -0,0 +1,25 @@
...
+        if os.getenv("CRT_FIPS") is not None:
+            env.shell.exec("mvn", "-P", "continuous-integration", "-B", "compile",
+                           "-Dcmake.crt_fips=ON", check=True)
+        else:
+            env.shell.exec("mvn", "-P", "continuous-integration",
+                           "-B", "compile", check=True)
+
+        # allow custom python to be used
+        parser = argparse.ArgumentParser()
+        parser.add_argument('--classifier')
```
TingDaoK, line 21: I use the argparser instead of env_variable because, in CI, `linux-musl-armv7` needs to pass the env var into the docker, which is not easy to do.

But for "CRT_FIPS", I also want our test to get the info and verify the run time check for FIPS is working as expected.

codebuild/cd/generic-unix-build.sh
```diff
@@ -29,7 +22,8 @@ fi
 BUILDER_VERSION=$(cat .github/workflows/ci.yml | grep 'BUILDER_VERSION:' | sed 's/\s*BUILDER_VERSION:\s*\(.*\)/\1/')
 echo "Using builder version ${BUILDER_VERSION}"
 
-aws s3 cp s3://aws-crt-builder/releases/${BUILDER_VERSION}/builder.pyz ./builder
+# aws s3 cp s3://aws-crt-builder/releases/${BUILDER_VERSION}/builder.pyz ./builder
+aws s3 cp s3://aws-crt-builder/channels/fips/builder.pyz ./builder
```
graebm, line 26: don't forget to fix this before merging

codebuild/cd/generic-unix-build.sh
```diff
@@ -43,3 +37,8 @@ JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64 mvn -B package -DskipTests -Dshared
 
 aws s3 cp --recursive --include "*.so" target/cmake-build/lib s3://aws-crt-java-pipeline/${GIT_TAG}/lib
 aws s3 cp target/ s3://aws-crt-java-pipeline/${GIT_TAG}/jar/ --recursive --exclude "*" --include "aws-crt*.jar"
+
+if [[ $AWS_CRT_TARGET != linux-armv8 ]]; then
+    # the FIPS subset contains FIPS build for linux-armv8 and linux-x64, and NON-FIPS build for rest of the linux platforms.
```
graebm, line 42: your comments says "linux-armv8 and linux-x64" but the if-statement is just for "linux-armv8"
Is this a bug, or am I confused?

src/native/crt.c
```diff
@@ -654,6 +657,30 @@ void JNICALL Java_software_amazon_awssdk_crt_CRT_dumpNativeMemory(JNIEnv *env, j
...
+ *
+ * Safeguard with macro's, for example because Libressl dosn't define
+ * FIPS_mode() by default.
+ *
+ * Note: FIPS_mode() does not change the FIPS state of libcrypto. This only returns the current state. Applications
+ * using s2n must call FIPS_mode_set(1) prior to s2n_init.
+ * */
+JNIEXPORT
+jboolean JNICALL Java_software_amazon_awssdk_crt_CRT_libcryptoIsFIPS(JNIEnv *env, jclass jni_crt_class) {
+    (void)env;
```
graebm, line 673: > must call FIPS_mode_set(1) prior to s2n_init

Is the CRT calling `FIPS_mode_set(1)` anywhere? Is there anything else we need to do to ensure S2N is doing everything the FIPS way?

Should we add an is_fips function to aws-c-io? aws-c-cal?

Should we simply call this function "isFips" instead of "libcryptoIsFips"?

TingDaoK: > Is the CRT calling FIPS_mode_set(1) anywhere? Is there anything else we need to do to ensure S2N is doing everything the FIPS way?

For java, we only support aws-lc, and aws-lc only rely on built-time to set the FIPS mode, [here](https://github.com/aws/aws-lc/blob/6d9a7675d1f7c2f159b76e4ade309220d54d96f4/include/openssl/crypto.h#L183-L185). So, we don't need to call `FIPS_mode_set(1)`. I copy/paste the part from s2n. I will remove this. 

s2n set up fips mode when libcrypto has FIPS mode enabled. [here](https://github.com/aws/s2n-tls/blob/main/crypto/s2n_fips.c#L51-L52). But, to really make sure we do the FIPS way, it's still the user's responsibility to use the right cypher suits, which is designed by s2n and aws-lc.

Discussed offline, for the possibility that we may support openssl in the future, changed the impl

> Should we add an is_fips function to aws-c-io? aws-c-cal?

probably not, they have access to `openssl/crypto.h`

codebuild/cd/generic-unix-build.sh
```diff
@@ -43,3 +36,10 @@ JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64 mvn -B package -DskipTests -Dshared
 
 aws s3 cp --recursive --include "*.so" target/cmake-build/lib s3://aws-crt-java-pipeline/${GIT_TAG}/lib
 aws s3 cp target/ s3://aws-crt-java-pipeline/${GIT_TAG}/jar/ --recursive --exclude "*" --include "aws-crt*.jar"
+
+if [[ $AWS_CRT_TARGET != linux-armv8 ]]; then
+    # This script used to compile for linux-armv6, linux-armv7 and linux-armv8 with no-fips
+    # Only when compile for linux-armv8, the shared lib should only be uploaded to /lib and not for /fips_lib
+    # Because: the FIPS subset contains FIPS build for linux-armv8 and linux-x64, and NON-FIPS build for all other platforms.
+    aws s3 cp --recursive --include "*.so" target/cmake-build/lib s3://aws-crt-java-pipeline/${GIT_TAG}/fips_lib
```
graebm, line 44: isn't linux-x86_64 also FIPS-compliant and use `generic-unix-build.sh`? I'm weirded out by `generic-unix-build.sh` being like "obviously I'm only used for arm builds"

related: this release process is so complicated, I'm worried we'll accidentally end up with a jar where the things we expected to be FIPS-compliant were accidentally overwritten by non-compliant libs.

Can we do a safety check to the final jar before it goes to the production Maven server? Ensure that all platforms we expected to be FIPS-compliant actually are?

And/Or can we simplify the release process so there's 1 clear place that sorts the fips-compliant from the non-fips-compliant. Like, instead of a bunch of different scripts copying libs into both `s3://.../lib/.` and `s3://.../fips_lib/.`. What if there was just 1 script that handled sorting? So the N scripts that upload their lib, do it to 1 place: lib/ or fips_lib/, as appropriate. Then another script does like:
* download everything from `s3://.../lib/`
* download everything from `s3://.../fips_lib/`
* locally, copy files from `lib/` to `fips_compat_lib/`
* locally, copy files from `fips_lib/` to `fips_compat_lib/`
* use `lib/` for regular uber-jar
* use `fips_compat_lib/` for `fips-compat` uber-jar classifier
* **BONUS POINTS**: use `fips_lib/` for `fips` uber-jar classifer that doesn't contain any non-compliant platforms

graebm (commented):
didn't get all the way through, but here are comments as I head out for the day

codebuild/cd/fips-aarch64-build.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+
+set -ex
+
+cd $(dirname $0)/../..
+
+git submodule update --init
+# use hte FIPS approved branch.
+cd crt/aws-lc/ && git checkout fips-2022-11-02 && cd ../../
```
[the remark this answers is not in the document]
TingDaoK: I don't think we should have 2 submodules, I can have the crt-java just use the fips branch for non-fips build as well. We can just check to make sure we are using the fips branch probably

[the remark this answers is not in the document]
TingDaoK: Also, the FIPS branch for aws-lc is relatively safer, since they are not actively develop on the FIPS branch. They only backport the needed change to the FIPS branch from main.

.github/workflows/ci.yml
```diff
@@ -101,6 +101,48 @@ jobs:
           aws s3 cp s3://aws-crt-test-stuff/ci/${{ env.BUILDER_VERSION }}/linux-container-ci.sh ./linux-container-ci.sh && chmod a+x ./linux-container-ci.sh
           ./linux-container-ci.sh ${{ env.BUILDER_VERSION }} aws-crt-${{ matrix.image }} build -p ${{ env.PACKAGE_NAME }}
 
+
+  linux-fips-x64:
+    runs-on: ubuntu-22.04 # latest
+    strategy:
+      matrix:
+        awslc-branch: [main, fips-2022-11-02]
```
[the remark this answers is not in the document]
graebm: we don't need to be checking out branches in CI anymore, right?

codebuild/cd/fips-aarch64-build.sh
graebm, line 1: put "linux" somewhere in this filename

codebuild/cd/fips-aarch64-build.sh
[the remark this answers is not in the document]
graebm: If the fips branch is fine for non-fips, then yeah let's just always use it.

codebuild/cd/fips-jar-build.yml
```diff
@@ -0,0 +1,31 @@
...
+      - sudo add-apt-repository ppa:openjdk-r/ppa
+      - sudo apt-get update -y
+      - sudo apt-get install openjdk-11-jdk-headless maven wget unzip -y -f
+  build:
+    commands:
+      - cd $CODEBUILD_SRC_DIR/aws-crt-java
+      # upload artifacts to S3
+      - export GIT_TAG=$(git describe --tags)
+      - mkdir -p target/cmake-build/lib
+      # prepare FIPS uber jar, download the regular libs
```
graebm, line 17: trivial:
```suggestion
      # prepare fips-compat uber jar, download the regular libs
```

CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ cmake_minimum_required(VERSION 3.1)
 
 project(aws-crt-jni C)
 option(BUILD_DEPS "Builds aws common runtime dependencies as part of build" ON)
+option(CRT_FIPS "Whether to build with FIPS compliance" OFF)
```
graebm, line 5: debatable: I see you failing if FIPS=ON and this is 32bit linux, but there are still a lot of other cases where we do NOT fail (e.g. Windows, Apple, musl, armv6/7)

I'm just spooked about us *thinking* we're making FIPS builds but one day a year from now, one of the many intermediate steps for it to actually be FIPS doesn't work right during the release process

codebuild/cd/fips-jar-build.yml
```diff
@@ -0,0 +1,31 @@
...
+      - aws s3 cp --recursive s3://aws-crt-java-pipeline/${GIT_TAG}/lib $CODEBUILD_SRC_DIR/aws-crt-java/target/cmake-build/lib
+      # Override with the FIPS libs
+      - aws s3 cp --recursive s3://aws-crt-java-pipeline/${GIT_TAG}/fips_lib $CODEBUILD_SRC_DIR/aws-crt-java/target/cmake-build/lib
```
graebm, line 20: again, spooked about accidentally uploading fips jar without the expected fips libs. Can we assert that `s3://.../fips_lib` has the expected files in it?

codebuild/cd/manylinux-x86-build.yml
```diff
@@ -7,22 +7,19 @@ phases:
   pre_build:
     commands:
       - export CC=gcc
+      - /opt/python/cp37-cp37m/bin/python -m pip install --upgrade awscli
```
graebm, line 10: trivial:
```suggestion
      # AWSCLI v2 not available on x86, so installing v1
      - /opt/python/cp37-cp37m/bin/python -m pip install --upgrade awscli
```

codebuild/cd/fips-jar-build.yml
TingDaoK: well, the fips shared lib will have the same structure and name as the regular one, I don't think we can assert on the files to make sure it works.

We can probably run a test instead.

codebuild/cd/fips-compat-jar-build.yml
```diff
@@ -0,0 +1,34 @@
...
+      # upload artifacts to S3
+      - export GIT_TAG=$(git describe --tags)
+      - mkdir -p target/cmake-build/lib
+      # prepare fips-compat uber jar, download the regular libs
+      - aws s3 cp --recursive s3://aws-crt-java-pipeline/${GIT_TAG}/lib $CODEBUILD_SRC_DIR/aws-crt-java/target/cmake-build/lib
+      # Override with the FIPS libs
+      - aws s3 cp --recursive s3://aws-crt-java-pipeline/${GIT_TAG}/fips_lib $CODEBUILD_SRC_DIR/aws-crt-java/target/cmake-build/lib
+      # Run a test to make sure we get the FIPS libs
+      - export CRT_FIPS=ON
+      - mvn test -Dtest=software.amazon.awssdk.crt.test.SystemInfoTest -Dshared-lib.skip=true
```
graebm, line 23: trivial: cool! but let's be a bit more explicit
```suggestion
      - mvn test -Dtest=software.amazon.awssdk.crt.test.SystemInfoTest.testIsFIPS -Dshared-lib.skip=true
```

codebuild/cd/fips-compat-jar-build.yml
```diff
@@ -0,0 +1,34 @@
...
+      - cd $CODEBUILD_SRC_DIR/aws-crt-java
+      # upload artifacts to S3
+      - export GIT_TAG=$(git describe --tags)
+      - mkdir -p target/cmake-build/lib
+      # prepare fips-compat uber jar, download the regular libs
+      - aws s3 cp --recursive s3://aws-crt-java-pipeline/${GIT_TAG}/lib $CODEBUILD_SRC_DIR/aws-crt-java/target/cmake-build/lib
+      # Override with the FIPS libs
+      - aws s3 cp --recursive s3://aws-crt-java-pipeline/${GIT_TAG}/fips_lib $CODEBUILD_SRC_DIR/aws-crt-java/target/cmake-build/lib
+      # Run a test to make sure we get the FIPS libs
+      - export CRT_FIPS=ON
```
graebm, line 22: is it worrisome that CRT_FIPS is set for all build and deployment subsequent steps in this yml file? Maybe just set it for 1 line, like: `CRT_FIPS=ON mvn test ...`

CMakeLists.txt
graebm: Maybe add a check later in the file like `if (CRT_FIPS AND NOT FIPS) fail fail fail`

since we only set `FIPS=ON` when going down the aws-lc path

src/test/java/software/amazon/awssdk/crt/test/S3ClientTest.java
```diff
@@ -852,7 +854,15 @@ public void testS3Put() throws IOException {
         skipIfAndroid();
         skipIfNetworkUnavailable();
         Assume.assumeTrue(hasAwsCredentials());
-        testS3PutHelper(false, false, null, false, 16 * 1024 * 1024);
+        testS3PutHelper(false, false, null, false, 16 * 1024 * 1024, false);
+    }
+
+    @Test
+    public void testS3PutWithMD5() throws IOException {
```
graebm, line 861: add a comment explaining how we were especially worried about FIPS and MD5 and that's one of the reasons this test is here

graebm (approved):
fix & ship

PR #792, opened 2024-06-07
Running tests for arm CI

TingDaoK:
*Issue #, if available:*

- The arm build was not running tests as expected

*Description of changes:*
- Fix the CI to run tests properly for linux-musl-arm
- Fix reading the environment variable properly
- Tried to use codebuild arm host for the ARM, referring to https://docs.aws.amazon.com/codebuild/latest/userguide/action-runner.html. But, for armv7, we still need qemu and nothing really improves
- Also, looks like the log is missing somehow when we use codebuild, eg: https://github.com/awslabs/aws-crt-java/actions/runs/9457170557/job/26060185356
- Another finding worth to mention is `multiarch/qemu-user-static` does not support arm host properly, [here](https://github.com/awslabs/aws-crt-builder/pull/285) is my attempt to use https://github.com/dbhi/qus for arm host, which works, but we suffer from the docker pull limits, and no needs as we are not using arm host now.

### TODO:
- We should probably move the armv7 test separately as an integration test. Not running those for each commit pushed.
- github runner for ARM is coming soon, https://github.com/orgs/community/discussions/73040, which may also have the same issue that for armv7, we still need qemu to run, but may work better in term of missing log or the docker pull limits?


.github/workflows/ci.yml
```diff
@@ -137,24 +118,34 @@ jobs:
...
 
-  # armv7 needs its own action due to inability to differentiate at runtime armv6 vs armv7 in the jvm:
-  #
-  # At build time we can properly figure out that we're targeting armv7.
-  # At run time we have to force armv7 (via environment variable) in order to achieve proper resource path
-  # resolution.
-  linux-musl-armv7:
-    runs-on: ubuntu-20.04 # latest
+
+  linux-musl-x86:
```
waahm7, line 122: wrong name?
```suggestion
  linux-musl-x64:
```

builder.json
```diff
@@ -87,7 +87,23 @@
                     ],
                     "!test_steps": [
                         "aws-crt-java-test"
-                    ]
+                    ],
+                    "!run_tests": true,
+                    "architectures": {
+                        "armv7": {
+                            "_comment": [
+                                "Due to inability to differentiate at runtime armv6 vs armv7 in the jvm",
```
waahm7, line 95: A comment about why we can't differentiate at runtime would be greatly appreciated.

builder.json
```diff
@@ -154,4 +170,4 @@
             "!build_env": {}
         }
     }
-}
+}
```
waahm7, line 173: trivial: blank line at the end.
```suggestion
}

```

waahm7 (approved):
Great job 👍

.github/workflows/ci.yml
```diff
@@ -137,24 +118,34 @@ jobs:
...
+    runs-on: ubuntu-22.04 # latest
```
waahm7, line 123:
```suggestion
    runs-on: ubuntu-24.04 # latest
```

.github/workflows/ci.yml
```diff
@@ -137,24 +118,34 @@ jobs:
...
+          - alpine-3.16-x64
+          - openwrt-x64-openjdk8
+    steps:
+      - name: Build ${{ env.PACKAGE_NAME }}
+        run: |
+          aws s3 cp s3://aws-crt-test-stuff/ci/${{ env.BUILDER_VERSION }}/linux-container-ci.sh ./linux-container-ci.sh && chmod a+x ./linux-container-ci.sh
+          ./linux-container-ci.sh ${{ env.BUILDER_VERSION }} aws-crt-${{ matrix.image }} build -p ${{ env.PACKAGE_NAME }}
+
+  linux-musl-arm:
+    runs-on: ubuntu-22.04 # latest
```
waahm7, line 136: trivial:
```suggestion
    runs-on: ubuntu-24.04 # latest
```

builder.json
TingDaoK: well, I also don't know... Just copy/paste from the comments in our CI before.

PR #800, opened 2024-06-21
Expose operationName for S3 requests

graebm:
**Issue:**
If an `S3MetaRequestType.PUT_OBJECT` failed, you couldn't tell exactly which S3 operation failed. Was it PutObject? StartMultipartUpload? UploadPart? CompleteMultipartUpload?

**Description of changes:**
- Expose S3 operation name in `S3FinishedResponseContext`
- Add `S3MetaRequestOptions.withOperationName(String)`, so users sending requests of type `S3MetaRequestType.DEFAULT` can pass the actual operation name through.
    - You _SHOULD_ set this for `DEFAULT` requests. It will be required in the very near future
- aws-lc: AWS-LC-FIPS-2.0.12 -> AWS-LC-FIPS-2.0.13

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/s3/S3FinishedResponseContext.java
+++ b/src/main/java/software/amazon/awssdk/crt/s3/S3FinishedResponseContext.java
@@ -10,6 +10,7 @@ public class S3FinishedResponseContext {
     private final int errorCode;
     private final int responseStatus;
     private final byte[] errorPayload;
+    private final String errorOperationName;
     private final ChecksumAlgorithm checksumAlgorithm;
     private final boolean didValidateChecksum;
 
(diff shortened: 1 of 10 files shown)


src/main/java/software/amazon/awssdk/crt/s3/S3FinishedResponseContext.java
```diff
@@ -53,6 +55,18 @@ public byte[] getErrorPayload() {
...
+     * @return the name of the S3 operation that failed, in the case of a failed HTTP response.
+     * For example, if {@link S3MetaRequestOptions.MetaRequestType#PUT_OBJECT} fails
+     * this could be "PutObject", "CreateMultipartUpload", "UploadPart",
+     * "CompleteMultipartUpload", or others. For {@link S3MetaRequestOptions.MetaRequestType#DEFAULT},
+     * this is the name passed to {@link S3MetaRequestOptions#withOperationName}.
+     * May be null.
+     */
+    public String getErrorOperationName() {
+        return this.errorOperationName;
+    }
```
waahm7, line 68: Debatable: Can we always set it for success response as well? Might make things simpler. Ignore if this requires too much effort of changing it in the C code as well.

src/native/s3_client.c
```diff
@@ -998,6 +1019,17 @@ JNIEXPORT jlong JNICALL Java_software_amazon_awssdk_crt_s3_S3Client_s3ClientMake
...
 
+    if (jni_operation_name) {
+        operation_name = aws_jni_byte_cursor_from_jbyteArray_acquire(env, jni_operation_name);
+        if (operation_name.ptr == NULL) {
+            goto done;
+        }
+        if (operation_name.len == 0) {
+            aws_jni_throw_illegal_argument_exception(env, "Operation name cannot be empty");
+            goto done;
+        }
```
waahm7, line 1030: Why are we not throwing exception in the first if? Should we just combine these if statements?

waahm7 (approved):

src/main/java/software/amazon/awssdk/crt/s3/S3FinishedResponseContext.java
graebm: Most of the members for S3FinishedResponseContext aren't set on success, since they're specifically about the properties of a failed HTTP response (headers, status-code, body, operation-name). And for something like multipart PUT_OBJECT, there are actually multiple different operations going on, so in the case of success ... which one would you be talking about?

src/native/s3_client.c
graebm: If `operation_name.ptr == NULL` then the `aws_jni_string_from_string()` call on the previous line failed, and it already raised an exception.

https://github.com/awslabs/aws-crt-java/blob/d3ab7768c426cda20f8ed33695f019e6d2af3bcc/src/native/crt.h#L162-L165

PR #802, opened 2024-06-24
S3Client breaking change: DEFAULT requests must pass operation name

graebm:
**Issue:**
see: https://github.com/awslabs/aws-c-s3/pull/439

**Description of changes:**
- When making requests with the `S3Client` of type `MetaRequestType.DEFAULT`, the user MUST specify the operation name
- Latest submodules:
    ```
    aws-c-cal          v0.6.15 -> v0.7.1
    aws-c-common       v0.9.21 -> v0.9.23
    aws-c-s3           v0.5.10 -> v0.6.0
    s2n                v1.4.16 -> v1.4.17
    ```

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/s3/S3Client.java
+++ b/src/main/java/software/amazon/awssdk/crt/s3/S3Client.java
@@ -142,11 +142,10 @@ public class S3Client extends CrtResource {
         }
 
         String operationName = options.getOperationName();
-        if (operationName == null && options.getMetaRequestType() == S3MetaRequestOptions.MetaRequestType.DEFAULT) {
-            // The upcoming release of aws-c-s3 v0.6.0 will require an operation name for DEFAULT meta-requests.
-            // Set a default value for now, in case this code gets mixed with the upcoming aws-c-s3 release.
-            // But when we take the upcoming release for real, change this code to throw IllegalArgumentException.
-            operationName = new String("DEFAULT");
+        if (options.getMetaRequestType() == S3MetaRequestOptions.MetaRequestType.DEFAULT && operationName == null) {
+            Log.log(Log.LogLevel.Error, Log.LogSubject.S3Client,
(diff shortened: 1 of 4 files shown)


src/main/java/software/amazon/awssdk/crt/s3/S3Client.java
```diff
@@ -142,11 +142,10 @@ public S3MetaRequest makeMetaRequest(S3MetaRequestOptions options) {
         }
 
         String operationName = options.getOperationName();
-        if (operationName == null && options.getMetaRequestType() == S3MetaRequestOptions.MetaRequestType.DEFAULT) {
-            // The upcoming release of aws-c-s3 v0.6.0 will require an operation name for DEFAULT meta-requests.
-            // Set a default value for now, in case this code gets mixed with the upcoming aws-c-s3 release.
-            // But when we take the upcoming release for real, change this code to throw IllegalArgumentException.
-            operationName = new String("DEFAULT");
+        if (options.getMetaRequestType() == S3MetaRequestOptions.MetaRequestType.DEFAULT && operationName == null) {
```
waahm7, line 145: Should we check for empty string?

waahm7 (approved):

src/main/java/software/amazon/awssdk/crt/s3/S3Client.java
graebm: The length is checked in the C binding: https://github.com/awslabs/aws-crt-java/blob/a90227261b09a9243532c46f877ff4476e5e5fe4/src/native/s3_client.c#L1027-L1028

Originally, I put the length check here in java, but then I saw [other strings](https://github.com/awslabs/aws-crt-java/blob/a90227261b09a9243532c46f877ff4476e5e5fe4/src/native/s3_client.c#L1038-L1039) getting their length checked in C so I made this similar

PR #816, opened 2024-08-12
GraalVM updates to extract the shared lib with native image instead of package as resource

TingDaoK:
*Issue #, if available:*

- Add a new GraalVM Feature for CRT to extract the share lib to the path of native image, instead of rely on the `extracting -> writing to temp -> load` every time
- In lambda, if you load CRT from the handler, this reduces the cold start time -> 300 ms
- if you load CRT from the constructor, this reduces the cold start time -> 30 ms. the cold-start request take 500 ms (I did a load CRT and done).
- Also reduced the binary size from **142M -> 101M** for this demo https://github.com/aws-samples/serverless-graalvm-demo

*Description of changes:*
- Extra change because of the feature added:
  - Adds the graalVM sdk as the "provided" dependency, refer to https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#dependency-scope, so that it will not be included in our jar.
  - Because of the "provided" dependency, the `javadoc` is not working properly as the dependency is not included during runtime. Switch to `mvn javadoc:javadoc` to generates the doc, which mvn can figures out the dependency during compile time.
 

src/main/java/software/amazon/awssdk/crt/CRT.java
```diff
@@ -245,64 +245,23 @@ private static List<String> runProcess(String[] cmdArray) throws IOException {
...
-
-            // Prefix the lib we'll extract to disk
-            String tempSharedLibPrefix = "AWSCRT_";
-
-            File tempSharedLib = File.createTempFile(tempSharedLibPrefix, libraryName, tmpdirFile);
-            if (!tempSharedLib.setExecutable(true, true)) {
+    /**
+     * Extract the CRT JNI library on current platform to a specific FIle.
+     */
+    public static void extractLibrary(File extractFile) {
```
DmitriyMusatkin, line 252: Can we have a note either here or in readme on how the users are supposed to use it and what the pitfalls are.

DmitriyMusatkin (approved):

graebm (changes requested):
How would one use this? There's no documentation

pom.xml
```diff
@@ -351,6 +357,13 @@
             <version>1.4</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.graalvm.sdk</groupId>
+            <artifactId>graal-sdk</artifactId>
+            <version>21.3.11</version>          <!-- The latest version supports JDK8 -->
```
graebm, line 363: at first glance, I thought you were saying 21.3.11 is the latest version, and we chose it because it supports JDK8
```suggestion
            <version>21.3.11</version> <!-- 21.3.11 is the last version to support JDK8 -->
```

src/main/java/software/amazon/awssdk/crt/CRT.java
```diff
@@ -29,7 +28,7 @@ public final class CRT {
     private static final String CRT_ARCH_OVERRIDE_SYSTEM_PROPERTY = "aws.crt.arch";
     private static final String CRT_ARCH_OVERRIDE_ENVIRONMENT_VARIABLE = "AWS_CRT_ARCH";
 
-    private static final String CRT_LIB_NAME = "aws-crt-jni";
+    public static final String CRT_LIB_NAME = "aws-crt-jni";
```
graebm, line 31: you just showe me an example where you didn't move the shared-lib along with the native image, but the error message was about how it couldn't extract the shared-lib from the JAR. But that doesn't apply at all anymore to GraalVM users

Please get the error messages to fit the new behavior and point people in the right direction

src/main/java/software/amazon/awssdk/crt/CRT.java
```diff
@@ -246,112 +245,65 @@ private static List<String> runProcess(String[] cmdArray) throws IOException {
...
-        } catch (Exception ex) {
-            System.err.println("Unable to unpack AWS CRT lib: " + ex);
+        // Prefix the lib we'll extract to disk
+        String tempSharedLibPrefix = "AWSCRT_";
+        File tempSharedLib = null;
+        try{
+            tempSharedLib = File.createTempFile(tempSharedLibPrefix, libraryName, tmpdirFile);
+        }
+        catch (IOException ex){
+            System.err.println("Unable to create temp file to extract AWS CRT library" + ex);
```
graebm, line 280: trivial: should the string end with ": " or something?
right now, `ex` is just going to get appended right after the letter "y"

src/main/java/software/amazon/awssdk/crt/internal/GraalVMNativeFeature.java
```diff
@@ -0,0 +1,22 @@
...
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+package software.amazon.awssdk.crt.internal;
+
+import org.graalvm.nativeimage.hosted.Feature;
+
+import software.amazon.awssdk.crt.CRT;
+
+/**
+ * Implementation of GraalVM feature to extract the share lib to the image path.
```
graebm, line 12:
```suggestion
 * Implementation of GraalVM feature to extract the share lib to the image path.
 * From GraalVM docs:
 * > When loading native libraries using System.loadLibrary() (and related APIs),
 * > the native image will search the directory containing the native image before searching the Java library path
 * https://www.graalvm.org/latest/reference-manual/native-image/dynamic-features/JNI/#loading-native-libraries
```

src/main/java/software/amazon/awssdk/crt/internal/GraalVMNativeFeature.java
```diff
@@ -0,0 +1,22 @@
...
+ * Internal API, not for external usage.
+ */
+public class GraalVMNativeFeature implements Feature {
+
+    @Override
+    public void afterImageWrite(AfterImageWriteAccess access) {
+        new CRT();
```
graebm, line 19: Is it OK that we're running CRT (extracting the lib to /tmp and loading it etc...) during the build process?

If it's problematic, maybe we could we shuffle code around so that whatever helper functions this needs to call, don't actually rely on the native code

javadoc.options
```diff
@@ -6,7 +6,7 @@
 -bottom 'Copyright &copy; Amazon.com, Inc. or its affiliates. All Rights Reserved.'
 -sourcepath src/main/java
 -notimestamp
--subpackages software.amazon.awssdk.crt
+-exclude software.amazon.awssdk.crt.internal
```
graebm, line 9: did you diff the new and old docs to ensure we're not accidentally pulling in tests? or accidentally omitting major classes?

TingDaoK: good point, I ran the make doc again with java22, no major different in terms of classes. I did find some of the options are not applied. Updated those option to maven configuration instead.

src/main/java/software/amazon/awssdk/crt/internal/GraalVMNativeFeature.java
TingDaoK: why it will be problematic?

src/main/java/software/amazon/awssdk/crt/CRT.java
TingDaoK: I don't think there is a decent way to tell graalvm apart from normal jvm here. :(

TingDaoK: actually `org.graalvm.nativeimage.imagecode` works

README.md
```diff
@@ -159,6 +159,13 @@ Platforms without FIPS compliance are also included in this jar, for compatibili
 > [!WARNING]
 > The classifier, and platforms with FIPS compliance are subject to change in the future.
 
+## GraalVM support
+
+Since version v0.29.20, GraalVM native image was supported. You can compile your application with AWS CRT in a GraalVM native image project without any additional configuration.
+
+Since version v0.30.12, GraalVM support was updated. Instead of packaging the JNI shared lib with native image as resource, the corresponding shared lib will be written to the same directory as the native image.
+In this way, it reduces the native image size around 30%, and avoids the extra loading time needed for extracting the JNI lib to the temporary path for load. No additional configuration needed.
+**Note**: the extracted JNI lib, `libaws-crt-jni.*` , will need to be exist in the directory containing the native image, if you move the native image to different directory.
```
graebm, line 168: Is it possible to keep the shared-lib in resources, but ONLY package the 1 shared-lib that users need?

TingDaoK: https://www.graalvm.org/sdk/javadoc/org/graalvm/nativeimage/hosted/RuntimeResourceAccess.html#addResource(java.lang.Module,java.lang.String)

This seems to be the only way available to make the resource from code. And it's introduced since 22.3.
And we are stuck at 21.3.11 for jdk8.

And I think it's fine for user to be aware of the separate shared lib to be part of the resource they need to include for graalvm.

1. GraalVM is already an advanced use case, people need to know their stuff to use GraalVM.
2. For Lambda use case, you need extra resources (boostrap script) for GraalVM to work, the shared lib is just one extra resource for user.

graebm (approved):

PR #837, opened 2024-10-16
Update CMake to 3.9

waahm7:
*Issue #, if available:*
See: https://github.com/awslabs/aws-c-common/pull/1159

*Description of changes:*
- Update minimum CMake version to 3.9.
- Fix Android build since previous version was using CMake 3.10. 


android/crt/build.gradle
```diff
@@ -53,7 +52,8 @@ android {
 
     defaultConfig {
         minSdkVersion 24  // TODO - dictated by CompletableFuture which is API 24+
-        targetSdkVersion 30
+        targetSdkVersion 33
```
waahm7, line 55: It was complaining that targetSDK 30 is not supported, and the minimum version should be 33 if you want to submit apps to Google Play. While we don't need to submit it to Google Play, 33 seems to work.

android/crt/build.gradle
```diff
@@ -43,8 +43,7 @@ ext {
 }
 
 android {
-    compileSdkVersion 30
```
waahm7, line 46: compileSdkVersion is deprecated and replaced by compileSdk

android/crt/build.gradle
```diff
@@ -43,8 +43,7 @@ ext {
...
-    buildToolsVersion "30.0.3"
```
waahm7, line 47: buildToolsVersion is deprecated and we don't need to specify it.

android/build.gradle
```diff
@@ -4,7 +4,7 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:7.0.0'
+        classpath 'com.android.tools.build:gradle:7.4.2'
```
waahm7, line 7: Gradle 7.0 had a bug with multiple object files which we were running into. Updating to 7.4 seems to work. 7.4 was chosen randomly.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
+#Wed Oct 16 16:47:56 PDT 2024
```
waahm7, line 1: This file was automatically updated by Android Studio after updating Gradle to 7.4

android/crt/build.gradle
waahm7: We are not updating the minSdkVersion, so we should be fine.

android/crt/build.gradle
```diff
@@ -103,7 +103,6 @@ android {
         cmake {
             path "../../CMakeLists.txt"
             buildStagingDirectory "../../target/cmake-build"
-            version "3.10.2"
```
waahm7, line 106: Seems to work without an explicit version. If we face any issues, we can revert back to an explicit version.

graebm (approved):
great job chasing down all those gradle issues

PR #839, opened 2024-10-31
Fix Android CI

graebm:
**Issue:**
The Android CI started failing due to:
```
Daemon will be stopped at the end of the build after running out of JVM memory
> Task :android:crt:lintAnalyzeDebug FAILED
```

This has been a recurring issue since we updated a bunch of build stuff as part of updating CMake https://github.com/awslabs/aws-crt-java/pull/837.

**Description of changes:**

Manually set -Xmx (max heap size) to something huge (tested 2g and that works, but why not go bigger) via the cmdline when invoking gradle in CI. Only in CI, gradle daemon runs out of memory during "lintAnalyzeDebug" task, unless you specify it this way. You'd think that Java's default of 25% RAM (ubuntu24 runner has 12g, so max 4g) would be sufficient, but no. You'd think setting -Xmx via gradle.properties would help, but no.

Also, make it so a new commit cancels any old in-progress builds. It's a waste of energy to continue these builds, some of which take a looooooong time.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -240,12 +240,14 @@ jobs:
         with:
           java-version: '11'
           distribution: 'temurin'
-      - name: set up Gradle
-        uses: gradle/actions/setup-gradle@v4
       # Build and publish locally for the test app to find the SNAPSHOT version
       - name: Build ${{ env.PACKAGE_NAME }}
         run: |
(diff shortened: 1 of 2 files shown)


TingDaoK (approved):
image

waahm7 (approved):

PR #851, opened 2024-11-30
Fix release pipeline

graebm:
**Issue:**
The most recent release failed for 2 different reasons:
1) Many cross-compilation jobs failed, they all [use builder like this](https://github.com/awslabs/aws-crt-java/blob/ff164f25866afbcea64d61226904eb2ab420b348/codebuild/cd/generic-unix-build.sh#L31-L33)
    - failed while doing: `mv target/cmake-build/aws-crt-java/* target/cmake-build/`
    - error message: "target/cmake-build/deps already present"
2) The `aws-crt-java-manylinux-x64-fips` job failed with error "Could not find Go"

**Description of changes:**
1) Fix 1st issue by doing `mv` on the 1 directory we care about, instead of doing `mv` on a wildcard where lots of stuff we don't care about can collide
2) Fix 2nd issue by installing golang via package manager
    - Previously, it was manually unpacking golang to a folder, and adding a relative path (`.go/go/bin`) to `$PATH` (which is weird usually you add absolute paths). And broke when the new "prebuild aws-lc" logic changed the working directory
3) Other small fixes, that I'll comment on below in the file diff...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,6 +67,14 @@ if (BUILD_DEPS)
         if (CRT_FIPS)
             list(APPEND AWSLC_CMAKE_ARGUMENTS -DFIPS=ON)
             list(APPEND AWSLC_CMAKE_ARGUMENTS -DDISABLE_PERL=OFF)
+
+            # Pick up GO_PATH env-var, set by aws-crt-builder when cross-compiling, see:
+            # https://github.com/awslabs/aws-crt-builder/blob/31307c808ed9f2ea1eb16503b25a9b582f886481/builder/imports/golang.py#L84
+            # https://github.com/awslabs/aws-crt-builder/blob/31307c808ed9f2ea1eb16503b25a9b582f886481/builder/actions/cmake.py#L110
+            if (DEFINED ENV{GO_PATH})
+                list(APPEND AWSLC_CMAKE_ARGUMENTS -DGO_EXECUTABLE=$ENV{GO_PATH}/go)
+                message(STATUS "Overriding GO_EXECUTABLE to ${GO_EXECUTABLE}")
+            endif()
         else()
             list(APPEND AWSLC_CMAKE_ARGUMENTS -DDISABLE_PERL=ON)  # Disable codegen
             list(APPEND AWSLC_CMAKE_ARGUMENTS -DDISABLE_GO=ON)  # Disable codegen
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,11 +67,6 @@ if (BUILD_DEPS)
         if (CRT_FIPS)
             list(APPEND AWSLC_CMAKE_ARGUMENTS -DFIPS=ON)
             list(APPEND AWSLC_CMAKE_ARGUMENTS -DDISABLE_PERL=OFF)
-
-            if (DEFINED ENV{GO_PATH})
-                list(APPEND AWSLC_CMAKE_ARGUMENTS -DGO_EXECUTABLE=$ENV{GO_PATH}/go)
-                message(STATUS "Overriding GO_EXECUTABLE to ${GO_EXECUTABLE}")
-            endif()
         else()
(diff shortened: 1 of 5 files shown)


codebuild/cd/generic-unix-build.sh
```diff
@@ -29,10 +29,20 @@ chmod a+x builder
...
+# Double check that shared lib is where we expect
+if ! find target/cmake-build/lib -type f -name "*.so" | grep -q .; then
+  echo "No .so files found"
+  exit 1
+fi
 
 JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64 mvn -B package -DskipTests -Dshared-lib.skip=true -Dcrt.classifier=$CLASSIFIER
 
-aws s3 cp --recursive --include "*.so" target/cmake-build/lib s3://aws-crt-java-pipeline/${GIT_TAG}/lib
+aws s3 cp --recursive --exclude "*" --include "*.so" target/cmake-build/lib s3://aws-crt-java-pipeline/${GIT_TAG}/lib
```
graebm, line 47: Adding `--exclude "*"` because:

https://docs.aws.amazon.com/cli/latest/reference/s3/#use-of-exclude-and-include-filters
> Note that, by default, all files are included. This means that providing only an --include filter will not change what files are transferred. --include will only re-include files that have been excluded from an --exclude filter. If you only want to upload files with a particular extension, you need to first exclude all files, then re-include the files with the particular extension.

This wasn't causing any bugs, but it's code that wasn't doing what it was supposed to be doing, so fixed this here (and in a few other locations)

CMakeLists.txt
```diff
@@ -67,11 +67,6 @@ if (BUILD_DEPS)
         if (CRT_FIPS)
             list(APPEND AWSLC_CMAKE_ARGUMENTS -DFIPS=ON)
             list(APPEND AWSLC_CMAKE_ARGUMENTS -DDISABLE_PERL=OFF)
-
-            if (DEFINED ENV{GO_PATH})
-                list(APPEND AWSLC_CMAKE_ARGUMENTS -DGO_EXECUTABLE=$ENV{GO_PATH}/go)
-                message(STATUS "Overriding GO_EXECUTABLE to ${GO_EXECUTABLE}")
-            endif()
```
graebm, line 74: This wasn't causing problems, but I don't know why it's here, so removing it.

[GOPATH](https://go.dev/wiki/GOPATH) is a thing in Golang, but this one has an underscore in it??? GOPATH would point at the directory containing `bin/go`, but this one would point at the directory containing `go/bin/go`???

I believe this was some way of communicating with builder.json, that is no longer in use. Here are some commits from the PR that added these lines:
- GO_PATH added to builder.json and CMakeLists.txt here: https://github.com/awslabs/aws-crt-java/pull/772/commits/ed809e43c20f7c72e0ff2c8e93db2f551427dcaf
- GO_PATH tweaked in builder.json here https://github.com/awslabs/aws-crt-java/pull/772/commits/c11cf464494ee7e64dd90711cb7ceb152e4e42a5
- GO_PATH removed from builder.json here: https://github.com/awslabs/aws-crt-java/pull/772/commits/96c5e1e059d467af743e4bd8a7d1bab9bbc1bd11

TingDaoK: https://github.com/awslabs/aws-crt-builder/pull/267/files#diff-f0fd7d35e67059aa0e19b135ded7e70e89b146b793db24528f9f8bda7c07b855R80-R84

TingDaoK: This is used for cross compile, we need to install the go executable for the target platform and override the executable path for the aws-lc to find the right go executable to use. 

so, we still need this to override the go_path if it's cross compile. We could add more comments to make it more clear.

TingDaoK (approved):
:shipit:

CMakeLists.txt
graebm: I put this back, and added comment explaining it

PR #859, opened 2025-01-14
Tweaks for codebuild jobs to run on native platforms

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/codebuild/cd/musl-linux-build.sh
+++ b/codebuild/cd/musl-linux-build.sh
@@ -23,7 +23,7 @@ export QEMU_IMAGE=123124136734.dkr.ecr.us-east-1.amazonaws.com/multiarch-qemu-us
 docker run --rm --privileged ${QEMU_IMAGE} --reset -p yes
 
 export BRANCH_TAG=$(git describe --tags)
-docker run --mount type=bind,src=$(pwd),dst=/root/aws-crt-java --env CXXFLAGS --env AWS_CRT_ARCH --platform =${ARCH} $DOCKER_IMAGE --version=${BUILDER_VERSION} build -p aws-crt-java --classifier ${CLASSIFIER} --branch ${BRANCH_TAG} run_tests=false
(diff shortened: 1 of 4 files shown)


codebuild/cd/musl-linux-build.sh
```diff
@@ -6,18 +6,24 @@ IMAGE_NAME=$1
...
+
+# on x86-64 and aarch64 we run on native images in codebuild 
+# (note: we still register mapping for those platforms, but in practice they are not used)
+# on other platforms we require emulation which is done through binfmt
 export QEMU_IMAGE=123124136734.dkr.ecr.us-east-1.amazonaws.com/multiarch-qemu-user-static:latest
 docker run --rm --privileged ${QEMU_IMAGE} --reset -p yes
 
 export BRANCH_TAG=$(git describe --tags)
-docker run --mount type=bind,src=$(pwd),dst=/root/aws-crt-java --env AWS_ACCESS_KEY_ID --env AWS_SECRET_ACCESS_KEY --env AWS_DEFAULT_REGION --env CXXFLAGS --env AWS_CRT_ARCH $DOCKER_IMAGE --version=${BUILDER_VERSION} build -p aws-crt-java --classifier ${CLASSIFIER} --branch ${BRANCH_TAG} run_tests=false
+docker run --mount type=bind,src=$(pwd),dst=/root/aws-crt-java --env AWS_DEFAULT_REGION --env CXXFLAGS --env AWS_CRT_ARCH $DOCKER_IMAGE --platform =${ARCH} --version=${BUILDER_VERSION} build -p aws-crt-java --classifier ${CLASSIFIER} --branch ${BRANCH_TAG} run_tests=false
```
TingDaoK, line 26: we are not doing cross-compiling, why do we need this `--platform`?

DmitriyMusatkin: it does not hurt to specify the platform (and its kinda additional sanity check that we are running on the expected platform). 
in future when we can restore armv7 build we will need this to get rid if warning about platform mismatch

TingDaoK: > in future when we can restore armv7 build we will need this to get rid if warning about platform mismatch

Why? I thought we will still have qemu and then load the docker image to build.

TingDaoK (approved):

PR #869, opened 2025-02-28
fix java arm ci

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
We rely on qemu to test aarch64 on x86-64 github runners.
Qemu had a bug where they had a bug with incorrectly mapping address space. That mostly worked fine until kernel patch that increased entropy in their address randomization. TLDR; qemu likes to segfault a lot against new kernels on our aarch64 test. Solution is to switch to a different github qemu user static image that forces a latest version of qemu, that has the bug fixed.
if you want to read more - https://github.com/tonistiigi/binfmt/issues/240

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -175,7 +175,7 @@ jobs:
         aws-region: ${{ env.AWS_DEFAULT_REGION }}
         role-duration-seconds: 14400 # these tests run slow and easily reach default cred expiry, hence change expiry to 4hrs
     - name: Install qemu/docker
-      run: docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
+      run: docker run --privileged --rm tonistiigi/binfmt --install all
     - name: Build ${{ env.PACKAGE_NAME }}
       run: |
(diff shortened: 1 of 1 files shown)


TingDaoK (approved):
we can update line 165 as well

.github/workflows/ci.yml
```diff
@@ -175,7 +175,7 @@ jobs:
         aws-region: ${{ env.AWS_DEFAULT_REGION }}
         role-duration-seconds: 14400 # these tests run slow and easily reach default cred expiry, hence change expiry to 4hrs
     - name: Install qemu/docker
-      run: docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
+      run: docker run --privileged --rm tonistiigi/binfmt --install all
```
waahm7, line 178: Should we have qemu for linux-arm CI as well?

waahm7 (approved):

.github/workflows/ci.yml
DmitriyMusatkin: good question. i have no idea how its currently running at all. let me check

TingDaoK: linux arm is cross-compiling, we don't need qemu for that

DmitriyMusatkin: yeah, its running dockcross through builder there, so we are fine

PR #874, opened 2025-03-14
Async cognito support

bretambrose:
Adds support for dynamically sourcing login tokens every time a Cognito credentials provider attempts to fetch credentials from Cognito.


src/native/credentials_provider.c
```diff
@@ -648,6 +659,258 @@ static int s_fill_in_logins(struct aws_array_list *logins, struct aws_byte_curso
...
+static struct aws_login_token_source_data *s_aws_login_token_source_data_acquire(
+    struct aws_login_token_source_data *login_token_source_data) {
+    if (login_token_source_data != NULL) {
+        aws_ref_count_acquire(&login_token_source_data->ref_count);
+    }
+
+    return login_token_source_data;
+}
+
+static struct aws_login_token_source_data *s_aws_login_token_source_data_release(
```
sbSteveK, line 707: Trivial: Why have this function return anything at all?

src/native/credentials_provider.c
```diff
@@ -648,6 +659,258 @@ static int s_fill_in_logins(struct aws_array_list *logins, struct aws_byte_curso
...
+
+    return NULL;
+}
+
+static struct aws_login_token_source_data *s_aws_login_token_source_data_new(
+    struct aws_allocator *allocator,
+    JNIEnv *env,
+    jobject login_token_source) {
+    if (login_token_source == NULL) {
+        return NULL;
```
sbSteveK, line 721: Trivial: There is a comment below where this is called "/* If no login token source is provided, this evaluates to NULL */". Would it be overkill to also comment here that this is called unconditionally and returning NULL is not an error?

sbSteveK (approved):
Trivial comments. Looks good!

src/native/credentials_provider.c
sbSteveK: This was for the release, but as you explained 
``` my_struct = release(my_struct); ``` 
as a pattern to assign NULL to the thing being released makes sense.

bretambrose: Oops, correct.

Issue #876, opened 2025-03-28
Long running SnapStart Lambda runs out of ephemeral storage


steven-aerts:
### Describe the bug

AWS SnapStart has a fixed [ephemeral storage size in `/tmp` of 512 Mb](https://docs.aws.amazon.com/lambda/latest/dg/snapstart.html#snapstart-runtimes).  We see long running Lambda's using the `aws-crt-java` library running out of ephemeral storage space:

```
Unable to unpack AWS CRT lib: java.io.IOException: No space left on device
java.io.IOException: No space left on device
at java.base/java.io.FileOutputStream.writeBytes(Native Method)
at java.base/java.io.FileOutputStream.write(Unknown Source)
at software.amazon.awssdk.crt.internal.ExtractLib.extractLibrary(ExtractLib.java:63)
at software.amazon.awssdk.crt.CRT.extractAndLoadLibrary(CRT.java:310)
at software.amazon.awssdk.crt.CRT.loadLibraryFromJar(CRT.java:330)
at software.amazon.awssdk.crt.CRT.<clinit>(CRT.java:50)
at software.amazon.awssdk.crt.CrtResource.<clinit>(CrtResource.java:104)
at software.amazon.awssdk.http.crt.AwsCrtHttpClientBase.<init>(AwsCrtHttpClientBase.java:77)
```

The reason for that is that SnapStart does not adhere the `deleteOnExit()` call necessary to clean up the shared objects which the `aws-crt-java` library extracts in the ephemeral storage.

Every time the AWS SnapStart re-initializes the lambda a copy of the shared library will leak chipping away 2 megabyte of the ephemeral storage.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

Old version of the library not leaking when AWS SnapStart re-initializes the function.

### Current Behavior

A typical example run showing `INIT_REPORT` or timeouts gradually filling the ephemeral storage:

```
Step 1: A typical request in snapstart takes 12 ms:

01:25:01.151 | START RequestId: 0axxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxx9e Version: 61
01:25:01.163 | END RequestId: 0axxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxx9e
[...]

PR #877, opened 2025-04-02
Cleanup SharedLib Immediately + Bump Minimum MacOS version to 10.15

waahm7:
*Issue #, if available:*
#876 

*Description of changes:*
- Clean up shared lib on non-Windows platforms immediately after loading. It will keep working for the current process and will be deleted as soon as the current process stops using it. See: https://stackoverflow.com/questions/31099347/what-happens-internally-when-deleting-an-opened-file-in-linux. I have verified it manually as well.
- Update to latest AWS-LC FIPS branch commit. Tagged release doesn't work due to CMake removing support for CMake <3.5, and AWS-LC team has decided that they will not cut a new release until there are any changes in the FIPS portion of the code. They recommended using the latest commit from the FIPS branch. 
- Update our minimum macOS version to 10.15 from 10.9 (The secItem changes in Aws-c-io (https://github.com/awslabs/aws-c-io/releases/tag/v0.18.0) require at least macOS 10.15).

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/CRT.java
+++ b/src/main/java/software/amazon/awssdk/crt/CRT.java
@@ -294,9 +294,14 @@ public final class CRT {
         // Ask Java to try and delete it on exit. We call this immediately
         // so that if anything goes wrong writing the file to disk, or
         // loading it as a shared lib, it will still get cleaned up.
+        // Once it is loaded successfully, we can safely delete it and it will keep working in the current process.
         tempSharedLib.deleteOnExit();
 
-        // Unfortunately File.deleteOnExit() won't work on Windows, where
+        ExtractLib.extractLibrary(tempSharedLib);
+        // load the shared lib from the temp path
+        System.load(tempSharedLib.getAbsolutePath());
+
+        // Unfortunately File.deleteOnExit() and File.delete() won't work on Windows, where
         // files cannot be deleted while they're in use. On Windows, once
         // our .dll is loaded, it can't be deleted by this process.
         //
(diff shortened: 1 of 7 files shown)


src/main/java/software/amazon/awssdk/crt/CRT.java
```diff
@@ -306,11 +311,10 @@ private static void extractAndLoadLibrary(String path) {
         String os = getOSIdentifier();
         if (os.equals("windows")) {
             tryDeleteOldLibrariesFromTempDir(tmpdirFile, tempSharedLibPrefix, libraryName);
+        } else {
+            // Once it is loaded successfully, we can safely delete it and it will keep working in the current process.
```
graebm, line 315: trivial: Move comments so platform-specific comments go along with the platform-specific actions.
Also, give the reason why we're doing early-delete on !windows
```
// Unfortunately, we can't rely solely on File.deleteOnExit() to clean things up...
if (windows)
    // File.deleteOnExit() won't work on Windows, where
    // ...blah blah rest of windows specific comments
    tryDeleteOldLibrariesFromTempDir();
else
    // On other platforms, File.deleteOnExit() might not run if <INSERT DETAILS>
    // Therefore, we delete the library file immediately after it was successfully loaded.
    // This is safe to do on non-Windows platforms, everything will keep working in the current process.
    tempSharedLib.delete();
```

src/main/java/software/amazon/awssdk/crt/CRT.java
```diff
@@ -294,9 +294,14 @@ private static void extractAndLoadLibrary(String path) {
         // Ask Java to try and delete it on exit. We call this immediately
         // so that if anything goes wrong writing the file to disk, or
         // loading it as a shared lib, it will still get cleaned up.
+        // Once it is loaded successfully, we can safely delete it and it will keep working in the current process.
```
graebm, line 297: details are platform-specific ... so save them for later
```suggestion
```

graebm (approved):

TingDaoK (approved):
maybe add to our readme as well about the minimal support version?

DmitriyMusatkin (approved):

PR #878, opened 2025-04-04
switch to critical mem access for checksums

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
GetByteArrayElements is proving to be really slow in some cases due to excessive copying. Switch over to GetPrimitiveArrayCritical for checksums.
This approach mimics what OpenJDK does for zlib crcs (minus the intrinsics annotations)

### DIFF
--- a/src/native/checksums.c
+++ b/src/native/checksums.c
@@ -9,6 +9,12 @@
 #include "crt.h"
 #include "java_class_ids.h"
 
+/**
+ * Note: we use critical mem access in below functions to speed up checksums.
+ * This approach is the same as what OpenJDK uses for their checksum implementation.
+ * Think twice before using similar approach elsewhere as it can lead to stalling.
+ */
+
 jint crc32_common(
     JNIEnv *env,
     jbyteArray input,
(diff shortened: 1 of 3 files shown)


src/native/crt.h
```diff
@@ -180,11 +185,32 @@ void aws_jni_byte_cursor_from_jstring_release(JNIEnv *env, jstring str, struct a
  ******************************************************************************/
 struct aws_byte_cursor aws_jni_byte_cursor_from_jbyteArray_acquire(JNIEnv *env, jbyteArray str);
 
+/*******************************************************************************
+ * aws_jni_byte_cursor_from_jbyteArray_critical_acquire - Creates an aws_byte_cursor from the
+ * bytes extracted from the supplied jbyteArray.
+ * The aws_byte_cursor MUST be given to aws_jni_byte_cursor_from jstring_release() when
+ * it's no longer needed, or it will leak.
```
TingDaoK, line 192:
```suggestion
 * The EXACT SAME aws_byte_cursor MUST be given to aws_jni_byte_cursor_from_jbyteArray_critical_release() when
 * it's no longer needed, or it will leak.
```

similar typo in the doc for `aws_jni_byte_cursor_from_jbyteArray_acquire`

Also, I feel like it's tricky that we return the byte cursor, while we need the same ptr to release.
But the cursor may change the ptr under the hood...

However, it was the same for the original function and getting used everywhere...

TingDaoK (approved):
LGTM

src/native/checksums.c
```diff
@@ -9,19 +9,29 @@
...
     jbyteArray input,
     jint previous,
     const size_t start,
     size_t length,
     uint32_t (*checksum_fn)(const uint8_t *, size_t, uint32_t)) {
-    struct aws_byte_cursor c_byte_array = aws_jni_byte_cursor_from_jbyteArray_acquire(env, input);
+    struct aws_byte_cursor c_byte_array = aws_jni_byte_cursor_from_jbyteArray_critical_acquire(env, input);
+    if (AWS_UNLIKELY(c_byte_array.ptr == NULL)) {
+        return previous;
+    }
```
waahm7, line 28: Why are we returning previous here?

waahm7 (approved):

src/native/checksums.c
DmitriyMusatkin: we need to return something. what does not really matter as java should find exception on the stack

PR #891, opened 2025-05-29
Migrate CD to maven central

TingDaoK:
*Issue #, if available:*

- migrate to maven central. since Publishing via OSSRH will be sunset on June 30th, 2025

*Description of changes:*
Details about the new Maven Central API can be found:
1. Maven Plugin :  https://central.sonatype.org/publish/publish-portal-maven/, which we used to publish snapshot here
2. Publisher API: https://central.sonatype.org/publish/publish-portal-api/, which we used to publish our classified JARs.
3. OSSRH - gradle: https://central.sonatype.org/publish/publish-portal-ossrh-staging-api/, which we used to publish android.

There are couple major change:
  1.  SNAPSHOT version is very limited now. I cannot figure out how to use the publiser API with snapshot. So, the snapshot only uses `central-publishing-maven-plugin` to test the uber jar.
  4. The ossrh API for maven central is now different with limitations as this PR written, see [here](https://central.sonatype.org/publish/publish-portal-ossrh-staging-api/#known-limitations) for details. It also doesn't support `deploy-file` plugin.
     - the major limitation affects us is that the APIs they provide currently automatically "close"  the staging repositories, so that it cannot be changed afterward.
     - The previous way we used is to open the staging repo, and upload different classified JAR and then close it will not work now.
   5. The `central-publishing-maven-plugin` doesn't seems to support publish multiple jar with different classifier,so that I worked around it by bundle all those jars together locally and then use the publisher API to upload the whole bundle.
   6. The gradle plugin `io.github.gradle-nexus.publish-plugin` also affected by the limitation mentioned above, so that the publish and release has to be in the same call now. https://github.com/gradle-nexus/publish-plugin/issues/379
   7. GPG API is a bit different now. https://github.com/keybase/keybase-issues/issues/2798 and https://maven.apache.org/plugins/maven-gpg-plugin/usage.html as they updated the needed environment variables for GPG@v2. 

As a result:
- Starts to use the latest Ubuntu version, since the maven version required is not available from the old Ubuntu version 
- Side affect of the new ubuntu is that it comes with new GPG, and updated the API to adapt the change from the new GPG as mentioned above.


szpak:
Btw, one limitation should be "fixed": https://github.com/gradle-nexus/publish-plugin/issues/379#issuecomment-2940868753

PR #914, opened 2025-09-05
Do not require `coreLibraryDesugaring` on Android

gmazzotta-bit:
The `java.time` package is only available natively on Android API 28+. Using it forces consuming apps to enable `coreLibraryDesugaring` to support older API levels.

This library currently uses `java.time.Instant` only to log the current date, which can be done easily using the old `java.util.Date`.

### DIFF
--- a/android/crt/build.gradle
+++ b/android/crt/build.gradle
@@ -138,17 +138,12 @@ android {
     compileOptions {
         sourceCompatibility = 1.8
         targetCompatibility = 1.8
-        // Enable desugaring so that Android lint doesn't flag `java.time` usage. Downstream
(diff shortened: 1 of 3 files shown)


gmazzotta-bit:
This is currently impacting `aws-iot-device-sdk-java-v2`, which requires [`coreLibraryDesugaring`](https://github.com/aws/aws-iot-device-sdk-java-v2/blob/d38dc94bfce5be49d8ee16217d4277f4737cc50b/android/iotdevicesdk/build.gradle#L104-L107) despite not using anything from `java.time` directly.

I'm all for using `java.time` when dealing with dates, but adding an extra dependency just to log a timestamp seems a bit overkill to me.

PR #920, opened 2025-10-07
Add TLS_CIPHER_PREF_TLSv1_0_2023 cipher pref for opt-ing out of PQ

WillChilds-Klein:
*Issue #, if available:* n/a

*Description of changes:* PR #921 incorporates CRT changes to prefer PQ TLS by default. This PR adds a cipher preference value allowing consumers to opt-out of that default if needed.


TingDaoK (commented):
Looks like the aws-lc update still gets to here.

PR #929, opened 2025-10-31
Dispatch Queue for macOS

sbSteveK:
When dealing with Dispatch Queue threads we need to attach and detatch them to and from the JVM during the operational period of the thread. For non Dispatch Queue threads we add a callback that invokes and handles the detatch from JVM when the aws_thread exits. For Dispatch Queue we have no control over which threads are used, in what order, or their lifetimes so we detach as soon as we are done using them in callbacks.

Added additional logging to `CrtResource.decRef` to help with tracking of what is decreffing what.

The CI Jobs we have listed as non-required due to their extensive length runs and android device farm have been split into their own separate workflows.

### DIFF
--- a/src/native/crt.c
+++ b/src/native/crt.c
@@ -58,9 +58,9 @@ struct aws_allocator *aws_jni_get_allocator(void) {
 Dispatch queue threads are handled differently than other types as we do not create and manage them. This is set during
 creation of an Event Loop Group to control attach/detach to/from JVM behavior of threads.
 */
-static bool s_disptach_queue_threads = false;
+static bool s_dispatch_queue_threads = false;
 void aws_jni_set_dispatch_queue_threads(bool is_dispatch_queue) {
-    s_disptach_queue_threads = is_dispatch_queue;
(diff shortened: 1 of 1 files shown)


src/native/crt.c
```diff
@@ -54,16 +54,22 @@ struct aws_allocator *aws_jni_get_allocator(void) {
...
+
 static void s_detach_jvm_from_thread(void *user_data) {
-    AWS_LOGF_DEBUG(AWS_LS_COMMON_GENERAL, "s_detach_jvm_from_thread invoked");
     JavaVM *jvm = user_data;
 
     /* we don't need this JNIEnv, but this is an easy way to verify the JVM is still valid to use */
     /********** JNI ENV ACQUIRE **********/
     JNIEnv *env = aws_jni_acquire_thread_env(jvm);
     if (env != NULL) {
-        (*jvm)->DetachCurrentThread(jvm);
```
sfod: why? :)

sbSteveK: I must have removed it during testing and didn't restore. Weird.

sfod (approved):

PR #938, opened 2025-11-18
Debug Improvements

sbSteveK:
Various debugging improvements

Log test names when they run
More explicit logging to track refCount manipulation in CrtResource
decRef logging descriptions for overwritten close() functions and for staticClose

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/eventstream/ClientConnectionContinuationHandler.java
+++ b/src/main/java/software/amazon/awssdk/crt/eventstream/ClientConnectionContinuationHandler.java
@@ -67,7 +67,7 @@ public abstract class ClientConnectionContinuationHandler implements AutoCloseab
     @Override
     public void close() {
         if (continuation != null) {
-            continuation.decRef();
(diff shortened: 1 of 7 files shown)


src/test/java/software/amazon/awssdk/crt/test/CrtTestFixture.java
```diff
@@ -230,6 +236,7 @@ public static void setupOnce() {
     /* The setup function will be run before every test */
     @Before
     public void setup() {
+        System.out.println("[TEST START] " + testName.getMethodName());
```
TingDaoK: https://github.com/awslabs/aws-crt-java/actions/runs/19476392289/job/55739151489?pr=938#step:4:2041
Do we really want all of those print out in the log? Is there an example about this helping debugging?

sbSteveK: There are times when tests fail but there is no record of which test from a class of tests was the one that failed. I've seen it most when tests time out and logs need to be truncated.

PR #940, opened 2025-11-20
Detach only external threads

sfod:
*Issue #, if available:*

**The issue this PR fixes appears on Apple platforms with dispatch queue event loop enabled and using GraalVM.**

Currently, when a native callback is executed, crt-java assumes it always runs on a dispatch queue thread. So, at the beginning of a callback, the thread attaches to JVM, and at the end - detaches from JVM. However, sometimes callbacks can be actually executed in the main thread, like in the [bootstrap create-destroy test](https://github.com/awslabs/aws-crt-java/blob/v0.39.4/src/test/java/software/amazon/awssdk/crt/test/ClientBootstrapTest.java#L29). This leads to callbacks attempting to detach the main JVM thread from JVM.
When executing in JVM mode, it seems such requests are simply being ignored, otherwise we should have seen errors. But in GraalVM this leads to the following error:
```
[TEST START] software.amazon.awssdk.crt.test.ClientBootstrapTest#testCreateDestroy
Fatal error: Must either be at a safepoint or in native mode
```
and then crash.

*Description of changes:*

This PR does for dispatch queue threads what is basically already done for CRT threads: detachment will happen only if attachment actually happened at the beginning of a callback.

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/s3/S3ClientOptions.java
+++ b/src/main/java/software/amazon/awssdk/crt/s3/S3ClientOptions.java
@@ -204,7 +204,7 @@ public class S3ClientOptions {
      * <p>
      * WARNING: This feature is experimental.
      * Currently, backpressure is only applied to GetObject requests which are split into multiple parts,
-     * and you may still receive some data after the window reaches zero.
+     * You will not receive any data after the window reaches zero until the window incremented.
      *
      * @param enable whether to enable or disable backpressure
      * @return this
--- a/src/test/java/software/amazon/awssdk/crt/test/S3ClientTest.java
+++ b/src/test/java/software/amazon/awssdk/crt/test/S3ClientTest.java
@@ -620,11 +620,12 @@ public class S3ClientTest extends CrtTestFixture {
         skipIfNetworkUnavailable();
         Assume.assumeTrue(hasAwsCredentials());
         final long fileSize = 1 * 1024 * 1024;
+        final long initialReadWindowSize = 1024;
         S3ClientOptions clientOptions = new S3ClientOptions()
 
                 .withRegion(REGION)
                 .withReadBackpressureEnabled(true)
-                .withInitialReadWindowSize(1024)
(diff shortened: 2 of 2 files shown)


sbSteveK (approved):

bretambrose (commented):
As annoying as it will be to do (given how often it's used), I'd prefer to see a slight change to how this is done:

Add a aws_jvm_thread_attachment_context struct (or similar name) that contains the env and the bool, ala:

```
struct aws_jvm_thread_attachment_context {
   JNIEnv *env;
   bool should_detach;
};
```

and have `aws_jni_acquire_thread_env()` return the struct by value.  Then the pattern becomes:

```
   struct aws_jvm_thread_attachment_context attachment_context = aws_jni_acquire_thread_env(jvm);
   JNIEnv *env = attachment_context.env;

   ...

   aws_jni_release(thread_env(jvm, &attachment_context);

```

I prefer encapsulating the attachment state in one type rather than splitting into two (and maybe more in the future?) separate fields.

bretambrose:
> As annoying as it will be to do (given how often it's used), I'd prefer to see a slight change to how this is done:
> 
> Add a aws_jvm_thread_attachment_context struct (or similar name) that contains the env and the bool, ala:
> 
> ```
> struct aws_jvm_thread_attachment_context {
>    JNIEnv *env;
>    bool should_detach;
> };
> ```
> 
> and have `aws_jni_acquire_thread_env()` return the struct by value. Then the pattern becomes:
> 
> ```
>    struct aws_jvm_thread_attachment_context attachment_context = aws_jni_acquire_thread_env(jvm);
>    JNIEnv *env = attachment_context.env;
> 
>    ...
> 
>    aws_jni_release(thread_env(jvm, &attachment_context);
> ```
> 
> I prefer encapsulating the attachment state in one type rather than splitting into two (and maybe more in the future?) separate fields.

Or maybe `struct aws_jvm_env` or similar to make it more general along the lines of "this holds all the state we need to perform JNI calls from native.

PR #981, opened 2026-04-16
Bind out the write data API

azkrishpy:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/http/HttpClientConnection.java
+++ b/src/main/java/software/amazon/awssdk/crt/http/HttpClientConnection.java
@@ -49,8 +49,22 @@ public class HttpClientConnection extends CrtResource {
      *
      * @param request The Request to make to the Server.
      * @param streamHandler The Stream Handler to be called from the Native EventLoop
-     * @param useManualDataWrites When true, request body data will be provided via
-     *        {@link HttpStreamBase#writeData} instead of from the request's body stream.
+     * @param useManualDataWrites When {@code true}, request body data is provided via
+     *        {@link HttpStreamBase#writeData} instead of from the request's
+     *        {@link HttpRequestBodyStream}.
+     *
+     *        <p>By design, CRT does not support setting both a body stream and enabling
+     *        manual writes for HTTP/1.1. Body streams are intended for requests whose
+     *        payload is available in full at the time of sending. If the stream does not
+     *        signal end-of-stream promptly, the event loop will busy-wait (hot-loop) for
+     *        more data, wasting CPU time. Manual writes avoid this by letting the caller
+     *        control when data is sent; the event loop only processes the request when
+     *        {@link HttpStreamBase#writeData} is called and is free to service other
+     *        requests in the meantime.
+     *
+     *        <p>If the request was created with an {@link HttpRequestBodyStream} and this
+     *        parameter is {@code true}, the first call to {@link HttpStreamBase#writeData}
+     *        will throw an {@link IllegalStateException}.
      * @throws CrtRuntimeException if stream creation fails
      * @return The HttpStream that represents this Request/Response Pair. It can be closed at any time during the
      *          request/response, but must be closed by the user thread making this request when it's done.
--- a/src/main/java/software/amazon/awssdk/crt/http/HttpClientConnection.java
+++ b/src/main/java/software/amazon/awssdk/crt/http/HttpClientConnection.java
@@ -41,6 +41,22 @@ public class HttpClientConnection extends CrtResource {
      */
     public HttpStream makeRequest(HttpRequest request, HttpStreamResponseHandler streamHandler)
             throws CrtRuntimeException {
+        return makeRequest(request, streamHandler, false);
+    }
+
+    /**
+     * Schedules an HttpRequest on the Native EventLoop for this HttpClientConnection specific to HTTP/1.1 connection.
+     *
+     * @param request The Request to make to the Server.
+     * @param streamHandler The Stream Handler to be called from the Native EventLoop
+     * @param useManualDataWrites When true, request body data will be provided via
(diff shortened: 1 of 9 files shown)


src/main/java/software/amazon/awssdk/crt/http/HttpStreamBase.java
```diff
@@ -112,6 +116,63 @@ public void cancel() {
...
+     * Write data to an HTTP stream. Works for both HTTP/1.1 and HTTP/2.
+     * The stream must have been created with {@code useManualDataWrites = true}.
+     * You must call activate() before using this function.
+     *
+     * @param data       data to send, or null to write zero bytes. Pass null with
+     *                   endStream=true to signal end-of-body without sending additional data.
+     * @param endStream  if true, this is the last data to be sent on this stream.
+     * @param completionCallback invoked when the data has been flushed or an error occurs.
+     */
+    public void writeData(final byte[] data, boolean endStream,
```
TingDaoK, line 136: I feel like we can also support taking `HttpRequestBodyStream`, which allows java impl to write directly into C buffer. (SDK probably doesn't use it, but leave the door open) Or we can do that later when we need it. 

Meanwhile mention the difference of this and the existing `writeChunk` method in HttpStream. Looks like we should just mark the writeChunk deprecate?

src/main/java/software/amazon/awssdk/crt/http/HttpClientConnection.java
```diff
@@ -65,13 +82,30 @@ public HttpStream makeRequest(HttpRequest request, HttpStreamResponseHandler str
...
+     * @param request The Request to make to the Server.
+     * @param streamHandler The Stream Handler to be called from the Native EventLoop
+     * @param useManualDataWrites When true, request body data will be provided via
+     *        {@link HttpStreamBase#writeData} instead of from the request's body stream.
+     * @throws CrtRuntimeException if stream creation fails
+     * @return The HttpStream that represents this Request/Response Pair. It can be closed at any time during the
+     *          request/response, but must be closed by the user thread making this request when it's done.
+     */
+    public HttpStreamBase makeRequest(HttpRequestBase request, HttpStreamBaseResponseHandler streamHandler,
+            boolean useManualDataWrites) throws CrtRuntimeException {
```
TingDaoK, line 121: IIRC, we have some restriction on with manual write, the request doesn't allow to have an input stream with it. Is that true?

azkrishpy: Do you mean the body stream? then yes, when manual writes is on, having a body stream with h1 would fail but h2 is supposed to send them one after the other and is handled by the H2 protocol..

TingDaoK: yeah, that's a very inconsistent behavior, either we document it out clearly, or try to just handle it. 

And if we decided to error out for the case, I think we should error out early in the binding, so that it's more clear comparing to our C error handling with one error code and need to look deep into the log to figure out the error

src/main/java/software/amazon/awssdk/crt/http/HttpStream.java
```diff
@@ -24,6 +32,52 @@ protected HttpStream(long ptr) {
...
+     * calling this method will throw {@link IllegalStateException}.
+     * <p>
+     * HTTP/2 does not have this restriction. HTTP/2 uses its own DATA frame
+     * framing (RFC 9113, Section 8.1), where the body stream and manual writes
+     * both append DATA frames to the same outgoing queue.
+     * <p>
+     * <b>Migration from writeChunk:</b> This method supersedes the deprecated
+     * {@link #writeChunk} methods. Use {@code writeData} for all manual body data
+     * writes on both HTTP/1.1 and HTTP/2 streams.
+     */
```
TingDaoK, line 51: I think this should be moved to the make_request instead, since it should target on the error case where manual_write is set with a http message with the body, right? 

Also, I am not sure I follow the reason why we cannot we have the body stream and manual write.

src/test/java/software/amazon/awssdk/crt/test/WriteDataTest.java
```diff
@@ -278,4 +279,71 @@ public void onResponseComplete(HttpStreamBase stream, int errorCode) {
         shutdownComplete.get(60, TimeUnit.SECONDS);
         CrtResource.waitForNoResources();
     }
+
+    /**
+     * Tests that makeRequest throws IllegalStateException when called with both
+     * a body stream and useManualDataWrites=true on an HTTP/1.1 connection.
+     */
+    @Test(expected = IllegalStateException.class)
```
TingDaoK: I am a bit leaning towards to use try catch the exception from the exact method we invoke, so that it will be assert on the exact exception raised from where we think they should be raised. 

from my understanding, this will be assert on the whole test raise the exception?

src/main/java/software/amazon/awssdk/crt/http/HttpClientConnection.java
```diff
@@ -49,27 +49,48 @@ public HttpStream makeRequest(HttpRequest request, HttpStreamResponseHandler str
...
+        }
         HttpStreamBase stream = httpClientConnectionMakeRequest(getNativeHandle(),
                 request.marshalForJni(),
                 request.getBodyStream(),
                 new HttpStreamResponseHandlerNativeAdapter(streamHandler),
                 useManualDataWrites);
 
-        return (HttpStream)stream;
+        HttpStream h1Stream = (HttpStream)stream;
+        h1Stream.setHasBodyStream(request.getBodyStream() != null);
```
TingDaoK: do we still need this?

TingDaoK (approved):
fix & ship

PR #987, opened 2026-05-12
Expose async writes to HTTP Stream Manager

azkrishpy:
*Issue #, if available:*

*Description of changes:*
- https://github.com/awslabs/aws-crt-java/pull/981 exposes write data to connections but stream manager's acquire API automatically calls connections with no boolean for useManualDataWrites. This change adds an overloaded implementation of acquire stream to add the manual writes as a boolean.
- This also changes native write data to avoid a copy by setting up a new global reference to user's byte array that will get cleaned up after callback is invoked or if the API errors out.

### DIFF
--- a/src/native/http_request_response.c
+++ b/src/native/http_request_response.c
@@ -611,7 +611,15 @@ JNIEXPORT jint JNICALL Java_software_amazon_awssdk_crt_http_HttpStream_httpStrea
 
 struct http_stream_write_data_callback_data {
     struct http_stream_binding *stream_cb_data;
-    struct aws_byte_buf data_buf;
+    /*
+     * Global ref to the Java byte[] passed in by the caller, and the aws_byte_cursor acquired
+     * from it via aws_jni_byte_cursor_from_jbyteArray_acquire. Held on callback_data so the
+     * JNI-side bytes stay valid for the full async write, and released from
+     * s_cleanup_write_data_callback_data once the write completes (or fails synchronously).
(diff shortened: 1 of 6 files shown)


src/native/http_request_response.c
```diff
@@ -676,12 +688,19 @@ JNIEXPORT jint JNICALL Java_software_amazon_awssdk_crt_http_HttpStreamBase_httpS
     };
 
     if (data != NULL) {
-        struct aws_byte_cursor data_cur = aws_jni_byte_cursor_from_jbyteArray_acquire(env, data);
-        aws_byte_buf_init_copy_from_cursor(&callback_data->data_buf, aws_jni_get_allocator(), data_cur);
-        aws_jni_byte_cursor_from_jbyteArray_release(env, data, data_cur);
-
-        data_cur = aws_byte_cursor_from_buf(&callback_data->data_buf);
-        callback_data->data_stream = aws_input_stream_new_from_cursor(aws_jni_get_allocator(), &data_cur);
+        /*
```
TingDaoK, line 691: let's add a test for this, if we don't already have one.
invoke the write data API with a local variable (Or somehow remove the reference to the object from java side). We should sill keep the data alive and works as expected since we keep the object referenced in C here.

azkrishpy: modified the existing happy path tests to create weak references and remove object, forced a gc and checked for the the weak reference to not be null.

TingDaoK (approved):

PR #988, opened 2026-05-13
Fail to Load CRT safely

azkrishpy:
*Issue #, if available:*

*Description of changes:*
When CRT is attempted to load on the wrong platform, currently it throws a runtime exception from a static class, causing jvm to automatically throw an [ExceptionInInitializerError](https://docs.oracle.com/javase/8/docs/api/java/lang/ExceptionInInitializerError.html) which is never caught and might crash users. This change attempts set a static throwable instead of throwing it and use a helper to check whether CRT was loaded correctly, avoiding an error being thrown while throwing the Exception.


sbSteveK (approved):
Everything looks good. It'd be nice if we had a test to make sure when `s_loadFailure` is set that we are getting notified the way we expect.

azkrishpy:
> Everything looks good. It'd be nice if we had a test to make sure when `s_loadFailure` is set that we are getting notified the way we expect.

It is hard to test internally since it would require for the native to not exist and adding it to CI such that this gets tested before native is built is a lot more complex. I tested this locally using a reproducer and Java SDK has already verified the this fix.

zoewangg (approved):

PR #990, opened 2026-05-22
Support s2n-tls on macOS

sfod:
Add support for s2n-tls as a TLS backend on macOS.
Apple Secure Transport remains the default for FIPS compliance.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,7 +29,7 @@ if (BUILD_DEPS)
     set(IN_SOURCE_BUILD ON)
     set(BUILD_TESTING OFF)
     add_subdirectory(crt/aws-c-common)
-    if (UNIX AND NOT APPLE)
+    if (UNIX)
         include(AwsPrebuildDependency)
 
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ if (BUILD_DEPS)
     set(IN_SOURCE_BUILD ON)
     set(BUILD_TESTING OFF)
     add_subdirectory(crt/aws-c-common)
-    if (UNIX AND NOT APPLE)
+    if (UNIX)
```
xiazhvera: In cpp we use the following to exclude the iOS and tvOS, would we need to do the same here? 
`if(UNIX AND NOT BYO_CRYPTO AND (NOT APPLE OR CMAKE_SYSTEM_NAME STREQUAL "Darwin"))`

xiazhvera (commented):
Not sure why we only use `if (UNIX)` in CMake. Except to that, the PR looks good to me.

xiazhvera (approved):

PR #994, opened 2026-06-05
 Add IoT SDK Metrics V2 with feature encoding and metadata support

rakshil14-2:
*Issue #, if available:*

*Description of changes:*
  * Added metadata entries support to `IoTDeviceSDKMetrics` for passing key-value metrics through JNI to aws-c-mqtt
  * Added encoder for CRT-level features (protocol, socket, TLS, proxy, jitter, session, queue, topic alias) into compact "ID/Value" format
  * Added metrics creation/merge logic : auto-sets CRTVersion, validates IoTSDKMetricsVersion before merging SDK features
  * Added user-configurable metrics API (`withMetrics()` / `setMetrics()`) for IoT SDK layer
  * Renamed `metricsEnabled` → `disableMetrics` (opt-out pattern, default false)
  * Stored TLS cipher preference and min TLS version on `TlsContext` for feature reporting
  * Updated JNI layer to parse metadata entries list and pass to C
  * Added unit tests for encoding, merging, and version validation
  
  **Breaking changes:**
  * `withMetricsEnabled(boolean)` → `withDisableMetrics(boolean)`
  * `getMetricsEnabled()` → `getDisableMetrics()`


src/main/java/software/amazon/awssdk/crt/internal/IoTDeviceSDKMetrics.java
```diff
@@ -4,18 +4,40 @@
  */
 package software.amazon.awssdk.crt.internal;
 
+import java.util.List;
+
 /**
  * @internal
```
xiazhvera: `IoTDeviceSDKMetrics` should be public accessible now. The class should be moved from internal to external folders, update the comments, and removed the internal notation.

src/main/java/software/amazon/awssdk/crt/internal/IoTMetricEncoder.java
```diff
@@ -0,0 +1,457 @@
...
+import java.util.List;
+import java.util.Map;
+
+/**
+ * @internal
+ * Encodes IoT SDK metrics features and creates final metrics object
+ *
+ */
+
+public class IoTMetricEncoder{
```
xiazhvera: `IoTMetricEncoder` is a private util class, we should make it as inaccessible as possible.
There are couple of approaches we could do:
1. Move it as a `private static` inner class of `IoTDeviceSDKMetrics`, preserving the logical separation while making it completely invisible outside this class.
2. Inline the encoding logic directly as `private static` methods of `IoTDeviceSDKMetrics`.

src/main/java/software/amazon/awssdk/crt/internal/IoTMetricEncoder.java
```diff
@@ -0,0 +1,457 @@
...
+     * A (retry_jitter_mode), B (session_behavior), C (offline_queue_behavior),
+     * D (outbound_topic_alias), E (inbound_topic_alias), H (http_proxy_type),
+     * J (tls_cipher_preference), K (minimum_tls_version)
+     *
+     * Feature I (certificate_source) is set at the IoT SDK level, not here.
+     *
+     * @param clientOptions MQTT5 client options containing connection configuration
+     * @return the encoded feature list string
+     */
+    public static String getEncodedFeatureListMqtt5(Mqtt5ClientOptions clientOptions) {
```
xiazhvera: The function is only used in `createMetricsMqtt5` and tests, doesn't seems need to be public. The same for Mqtt3 one.

src/main/java/software/amazon/awssdk/crt/internal/IoTMetricEncoder.java
```diff
@@ -0,0 +1,457 @@
...
+    /**
+     * Merges CRT-generated features with user-provided (IoT SDK) features.
+     * When both lists contain the same feature ID, the user-provided value
+     * takes precedence. Insertion order is preserved.
+     *
+     * @param crtFeatures CRT-generated feature list string (e.g., "F/5,G/A")
+     * @param userFeatures user-provided feature list from the IoT SDK (may be null or empty)
+     * @return the merged feature list string
+     */
+    public static String mergeFeatureLists(String crtFeatures, String userFeatures) {
```
xiazhvera: Should not be public.

src/main/java/software/amazon/awssdk/crt/iot/IoTMetricsMetadata.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+package software.amazon.awssdk.crt.internal;
+
+/**
+ * @internal
+ * A key-value pair for IoT SDK metrics metadata.
+ * Metadata entries are appended to the MQTT CONNECT packet username field
+ * as part of the Metadata query parameter.
+ */
+public class IoTMetricsMetadata {
```
xiazhvera, line 12: `IoTMetricsMetadata` should be public API.

src/main/java/software/amazon/awssdk/crt/io/TlsContext.java
```diff
@@ -13,6 +13,8 @@
  */
 public class TlsContext extends CrtResource {
 
+    private int cipherPreference;
+    private int minimumTlsVersion;
```
xiazhvera: Why use `int` for `minimumTlsVersion` and `cipherPreference`?

Using raw `int` values loses the readability benefits of enums. The current decoder has magic numbers like `case 6: return "A"` which are hard to understand without looking up the enum definitions.

src/main/java/software/amazon/awssdk/crt/internal/IoTMetricEncoder.java
```diff
@@ -0,0 +1,457 @@
...
+    // ======================== Metrics Creation ========================
+
+    /**
+     * Creates the final IoTDeviceSDKMetrics object for an MQTT5 client.
+     *
+     * @param clientOptions MQTT5 client options containing connection configuration
+     * @param userMetrics optional metrics configuration from the IoT SDK (may be null)
+     * @return the final metrics object ready for JNI passing
+     */
+    public static IoTDeviceSDKMetrics createMetricsMqtt5(Mqtt5ClientOptions clientOptions, IoTDeviceSDKMetrics userMetrics) {
```
xiazhvera: Is `userMetrics` a member of the `Mqtt5ClientOptions`? Why do we need pass it in separately?

src/main/java/software/amazon/awssdk/crt/internal/IoTMetricEncoder.java
```diff
@@ -0,0 +1,457 @@
...
+
+    /**
+     * Creates the final IoTDeviceSDKMetrics object for an MQTT3 connection.
+     *
+     * @param userMetrics optional metrics configuration from the IoT SDK (may be null)
+     * @param proxyOptions optional HTTP proxy options from the connection
+     * @param tlsCtx optional TLS context used by the connection
+     * @return the final metrics object ready for JNI passing
+     */
+    public static IoTDeviceSDKMetrics createMetricsMqtt3(IoTDeviceSDKMetrics userMetrics, HttpProxyOptions proxyOptions, TlsContext tlsCtx) {
```
xiazhvera: All three parameters (`userMetrics`, `proxyOptions`, `tlsContext`) can be obtained from `MqttConnectionConfig`. Passing them separately is unnecessary and creates API fragility.

In Python, you can add `**kwargs` or default parameters without breaking existing callers. In Java, adding a new parameter to a method signature is a breaking change. Therefore in Java API design, we tends to use structure/class as parameter for future-proofing.

src/main/java/software/amazon/awssdk/crt/mqtt/MqttConnectionConfig.java
```diff
@@ -542,23 +544,44 @@ public Consumer<WebsocketHandshakeTransformArgs> getWebsocketHandshakeTransform(
...
-     * Enables or disables IoT Device SDK metrics collection. The metrics includes SDK name, version, and platform.
+     * Disables IoT Device SDK metrics collection. The metrics includes SDK name, version, and platform.
+     * Default is false (metrics enabled).
      *
-     * @param enabled true to enable metrics, false to disable
+     * @param disable true to disable metrics, false to enable (default)
      */
-    public void setMetricsEnabled(boolean enabled) {
-        this.metricsEnabled = enabled;
+    public void setDisableMetrics(boolean disable) {
```
xiazhvera: Trivial: disabled or disableMetrics

xiazhvera (changes requested):

src/native/iot_device_sdk_metrics.c
```diff
@@ -57,10 +60,96 @@ struct aws_mqtt_iot_metrics_java_jni *aws_mqtt_iot_metrics_java_jni_create_from_
...
+        if (count > 0 && !aws_jni_check_and_clear_exception(env)) {
+            java_metrics->metadata_entries =
+                aws_mem_calloc(allocator, (size_t)count, sizeof(struct aws_mqtt_metadata_entry));
+
+            /* First pass: compute exact buffer size needed */
+            size_t total_size = 0;
+            for (jint i = 0; i < count; i++) {
+                jobject entry = (*env)->CallObjectMethod(env, metadata_list, boxed_list_properties.list_get_id, i);
+                if (entry == NULL || aws_jni_check_and_clear_exception(env)) {
+                    continue;
```
xiazhvera: Why do we continue instead of failing/break when `CallObjectMethod` returns NULL or throws an exception?

src/native/iot_device_sdk_metrics.c
```diff
@@ -57,10 +60,96 @@ struct aws_mqtt_iot_metrics_java_jni *aws_mqtt_iot_metrics_java_jni_create_from_
...
+                }
+                jstring key_jstr =
+                    (jstring)(*env)->GetObjectField(env, entry, iot_metrics_metadata_properties.key_field_id);
+                jstring value_jstr =
+                    (jstring)(*env)->GetObjectField(env, entry, iot_metrics_metadata_properties.value_field_id);
+                if (key_jstr != NULL && value_jstr != NULL) {
```
xiazhvera: As mentioned in Python PR, empty key or empty value is valid case. While I'm not sure how `GetObjectField` handled undefined fields though.

src/native/iot_device_sdk_metrics.c
```diff
@@ -57,10 +60,96 @@ struct aws_mqtt_iot_metrics_java_jni *aws_mqtt_iot_metrics_java_jni_create_from_
...
+                    (jstring)(*env)->GetObjectField(env, entry, iot_metrics_metadata_properties.value_field_id);
+
+                if (key_jstr != NULL && value_jstr != NULL) {
+                    const char *key_chars = (*env)->GetStringUTFChars(env, key_jstr, NULL);
+                    size_t key_len = (size_t)(*env)->GetStringUTFLength(env, key_jstr);
+                    const char *value_chars = (*env)->GetStringUTFChars(env, value_jstr, NULL);
+                    size_t value_len = (size_t)(*env)->GetStringUTFLength(env, value_jstr);
+
+                    /* Append key to storage buffer */
+                    size_t key_offset = java_metrics->metadata_storage.len;
```
xiazhvera: Use existing help functions instead of manually do string manipulate and memory copy. https://github.com/awslabs/aws-crt-java/blob/bacc2d17269887b234727f5645dc53005810681c/src/native/crt.c#L446, can also refer to [s_populate_user_properties](https://github.com/awslabs/aws-crt-java/blob/iot_metrics_improvement/src/native/mqtt5_packets.c#L126)

src/native/iot_device_sdk_metrics.c
```diff
@@ -57,10 +60,96 @@ struct aws_mqtt_iot_metrics_java_jni *aws_mqtt_iot_metrics_java_jni_create_from_
...
+                (*env)->DeleteLocalRef(env, value_jstr);
+                (*env)->DeleteLocalRef(env, entry);
+            }
+
+            /* Set the metrics struct fields */
+            java_metrics->metrics.metadata_entries = java_metrics->metadata_entries;
+            java_metrics->metrics.metadata_count = java_metrics->metadata_count;
+        }
+
+        (*env)->DeleteLocalRef(env, metadata_list);
```
xiazhvera, line 110: I'm not familiar with JNI memory management. Could you confirm that delete the jni object would not release the memory buffer? Could we have full crt logS that shows that metadata fields are set properly?

rakshil14-2: [DeleteLocalRef](https://docs.oracle.com/en/java/javase/17/docs/specs/jni/functions.html#deletelocalref): It will just release the local reference without affecting the memory buffer

src/main/java/software/amazon/awssdk/crt/iot/IoTDeviceSDKMetrics.java
```diff
@@ -0,0 +1,473 @@
...
+     * @param pref the configured TLS cipher preference
+     * @return {@code "A"} for {@code TLS_CIPHER_PREF_PQ_TLSv1_0_2021_05},
+     *         {@code "B"} for {@code TLS_CIPHER_PQ_DEFAULT},
+     *         {@code "C"} for {@code TLS_CIPHER_PREF_TLSv1_2_2025},
+     *         or {@code null} for system default/unknown
+     */
+    private static String tlsCipherPreferenceValue(TlsCipherPreference pref) {
+        if (pref == null) return null;
+        switch (pref) {
+            case TLS_CIPHER_PREF_PQ_TLSv1_0_2021_05: return "A"; // TLS_CIPHER_PREF_PQ_TLSv1_0_2021_05
```
xiazhvera: Why the encoder didn't cover all TlsCipherPreference? Thinking about it, let's have a full encoding list and update the value in Python as well.

src/main/java/software/amazon/awssdk/crt/io/TlsContextOptions.java
```diff
@@ -577,6 +604,14 @@ public TlsContextOptions withNoCertificateRevocation() {
         return this;
     }
 
+    /** 
+     * @return the {@link CertificateSource} of the configured mTLS, or {@code null}
+     *         if no mTLS source has been set (or the source has no defined metrics mapping).
+     */
+    public CertificateSource getCertificateSource() {
```
xiazhvera: The API could be package private.

src/main/java/software/amazon/awssdk/crt/iot/IoTDeviceSDKMetrics.java
```diff
@@ -0,0 +1,532 @@
...
+    /**
+     * Builds the final metrics object for an MQTT3 connection by encoding the CRT
+     * feature list from the connection configuration and merging it
+     * with any user-supplied metadata.
+     *
+     * @param config the MQTT3 connection configuration containing proxy, TLS, and user metrics
+     * @return the merged metrics object ready to be passed to JNI
+     */
+    public static IoTDeviceSDKMetrics createMetricsMqtt3(MqttConnectionConfig config) {
+        String crtFeatureList = getEncodedFeatureListMqtt3(config.getHttpProxyOptions(), config.getTlsContext());
```
xiazhvera: Minor: `getEncodedFeatureListMqtt3` would be more extendable if we just passed config as parameter. On the other hand, we are unlikely to extend mqtt3 config/options though.

src/main/java/software/amazon/awssdk/crt/iot/IoTDeviceSDKMetrics.java
```diff
@@ -0,0 +1,532 @@
...
+        return createMetrics(config.getMetrics(), crtFeatureList);
+    }
+
+    /**
+     * Generates the encoded feature list string for metrics from MQTT5 client options.
+     *
+     * <p>
```
xiazhvera, line 95: trivial: clean up.

src/main/java/software/amazon/awssdk/crt/iot/IoTDeviceSDKMetrics.java
```diff
@@ -0,0 +1,532 @@
...
+
+    /**
+     * Builds the final metrics object for an MQTT5 client by encoding the CRT
+     * feature list from {@code clientOptions} and merging it with any
+     * user-supplied metadata.
+     *
+     * @param clientOptions MQTT5 client options containing connection configuration and user metrics
+     * @return the merged metrics object ready to be passed to JNI
+     */
+    public static IoTDeviceSDKMetrics createMetricsMqtt5(Mqtt5ClientOptions clientOptions) {
```
xiazhvera, line 74: If the API has to be public, would we have a way to hide API from the doc? The same for `createMetricsMqtt3`

src/native/iot_device_sdk_metrics.c
```diff
@@ -11,7 +20,12 @@
 #include <java_class_ids.h>
 
 static char s_iot_device_sdk_metrics_string[] = "IoTDeviceSDKMetrics";
+struct aws_metadata_buf_holder {
```
xiazhvera: We could just use existing struct `buffer_and_cursor_array_holder_struct`.

xiazhvera (commented):
First pas, I still need go through the tests.

src/test/java/software/amazon/awssdk/crt/test/IoTDeviceSDKMetricsTest.java
```diff
@@ -0,0 +1,347 @@
...
+        assertTrue(feature.contains("A/B")); // Full
+        assertTrue(feature.contains("B/A")); // CLEAN
+        assertTrue(feature.contains("C/C")); // FAIL_ALL
+        assertTrue(feature.contains("D/B")); // LRU
+        assertTrue(feature.contains("E/A")); // Enabled
+        assertTrue(feature.contains("F/5")); // MQTT5
+    }
+
+    @Test
+    public void testDefaultEnumValuesOmitted() {
```
xiazhvera: I'm not sure if we need this. This test seems redundant to the minimal one, which also uses the default values.

src/main/java/software/amazon/awssdk/crt/iot/IoTDeviceSDKMetrics.java
rakshil14-2: added @hidden annotation

xiazhvera (approved):

PR #1002, opened 2026-06-30
AbortStream method to Http1StreamManager and ensure streams are released to connection pool

TingDaoK:
*Issue #, if available:*
- replicate https://github.com/awslabs/aws-crt-java/pull/1001

*Description of changes:*
Otherside of the original PR, which already be well described in the original description. 
- fixed couple test issue from the original PR
- Added comments about the requirement to activate the http/1.1 stream before using it.


alextwoods (approved):

src/main/java/software/amazon/awssdk/crt/http/Http1StreamManager.java
```diff
@@ -118,27 +137,34 @@ public int onResponseBody(HttpStreamBase stream, byte[] bodyBytesIn) {
...
-                     * `activate` is idempotent.
-                     **/
                     try {
                         stream.activate();
                     } catch (CrtRuntimeException e) {
                         /* If activate failed, complete callback will not be invoked */
                         streamHandler.onResponseComplete(stream, e.errorCode);
-                        /* Release the connection back */
-                        connManager.releaseConnection(conn);
+                        if (connectionReleased.compareAndSet(false, true)) {
```
azkrishpy, line 158: should we allow multiple release of connections natively?

azkrishpy: Looks like we do [allow it](https://github.com/awslabs/aws-c-http/blob/8acf604e12137d8e033c57108eecf3c352d4e0cc/source/connection.c#L412). Why should connection be released exactly once?

azkrishpy (approved):

src/main/java/software/amazon/awssdk/crt/http/Http1StreamManager.java
```diff
@@ -81,11 +82,24 @@ public CompletableFuture<HttpStream> acquireStream(HttpRequestBase request,
     /**
      * Request an HTTP/1.1 HttpStream from StreamManager.
      *
+     * <p>The returned future completes with an {@link HttpStream} that will be activated immediately
+     * after the future is completed. The stream's response callbacks (onResponseHeaders, etc.) may
+     * begin firing as soon as activate is called.
+     *
+     * <p><b>Important:</b> Operations on the stream (such as {@code cancel()}) must only be invoked
+     * after {@code activate()} has been called. If {@code cancel()} is called before activate, it
+     * is a no-op — the stream will still proceed normally once activate runs, which may lead to
```
DmitriyMusatkin, line 91: a bit more of a philosophical question, but should activate on cancelled stream really be a noop? should it not activate the stream at all?

TingDaoK: well, yeah, the activate on the cancelled stream should result in no-op.  
But, we kind designed the cancel on an unactivated stream to be a no-op. And without altering any state of the stream. https://github.com/awslabs/aws-c-http/blob/main/include/aws/http/request_response.h#L1222-L1232

If we keep that behavior to basically ignore the cancel, then it still makes more sense to let activate do its thing.

src/main/java/software/amazon/awssdk/crt/http/Http1StreamManager.java
TingDaoK: it actually ends up with `httpClientConnectionReleaseManaged` in the binding, it avoids the race between multiple threads trying to release the same connection back to the pool.

`aws_http_connection_manager_release_connection` cannot be invoked twice.
And the binding level is not thread safe.

src/main/java/software/amazon/awssdk/crt/http/Http1StreamManager.java
DmitriyMusatkin: in retrospect was it the correct choice to design it this way? not saying we should change it now. 
cancel being a noop for non-active connection seems to bring more problems that we have to work around.

DmitriyMusatkin (approved):

PR #1003, opened 2026-07-06
Add test for auto-PUBACK on QoS 1 publish received

sfod:
Verify that the client sends PUBACK automatically when acquirePublishAcknowledgementControl() is not called. Confirmed by the absence of broker re-delivery.

Bump aws-crt-builder to v0.9.96.

### DIFF
--- a/src/test/java/software/amazon/awssdk/crt/test/Mqtt5ClientTest.java
+++ b/src/test/java/software/amazon/awssdk/crt/test/Mqtt5ClientTest.java
@@ -2644,23 +2644,16 @@ public class Mqtt5ClientTest extends Mqtt5ClientTestFixture {
             CompletableFuture<byte[]> firstDeliveryFuture = new CompletableFuture<>();
             CompletableFuture<byte[]> unexpectedRedeliveryFuture = new CompletableFuture<>();
 
-            // Publisher
-            Mqtt5ClientOptionsBuilder publisherBuilder = new Mqtt5ClientOptionsBuilder(AWS_TEST_MQTT5_IOT_CORE_HOST, 8883l);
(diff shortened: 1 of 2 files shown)


src/test/java/software/amazon/awssdk/crt/test/Mqtt5ClientTest.java
```diff
@@ -2624,6 +2624,108 @@ public void QoS1_UC1() throws Exception {
...
+                 Mqtt5Client subscriber = new Mqtt5Client(subscriberBuilder.build())) {
+                publisher.start();
+                publisherEvents.connectedFuture.get(OPERATION_TIMEOUT_TIME, TimeUnit.SECONDS);
+                subscriber.start();
+                subscriberEvents.connectedFuture.get(OPERATION_TIMEOUT_TIME, TimeUnit.SECONDS);
+
+                // Subscribe to the topic with QoS 1
+                SubscribePacketBuilder subscribeBuilder = new SubscribePacketBuilder(testTopic, QOS.AT_LEAST_ONCE);
+                subscriber.subscribe(subscribeBuilder.build()).get(OPERATION_TIMEOUT_TIME, TimeUnit.SECONDS);
+
```
sbSteveK, line 2677: maybe we want to await the suback before sending the publish?

sbSteveK (approved):
Trivial addition just for sanity but should be fine since it's a pattern we use everywhere in these tests.

src/test/java/software/amazon/awssdk/crt/test/Mqtt5ClientTest.java
sbSteveK: nvm~~~~~

PR #1011, opened 2026-08-17
feat: expose more metrics for S3

azkrishpy:
*Issue #, if available:*

*Description of changes:*
Adds/fixes S3 request telemetry exposed via S3RequestMetrics:

- getServiceId() — now returns "S3" (was "s3"), matching SDK convention
- getServiceEndpoint() — now returns a full scheme-qualified URL (e.g. https://bucket.s3.region.amazonaws.com) instead of a bare host, so it can be parsed with URI.create()
- getTimeToFirstByte() / getTimeToLastByte() — now return durations (delta from request start), not raw timestamps
- getConnectionAcquisitionDurationNs() — new, time spent acquiring a connection for the request
- getHttpManagerMetrics() — new, snapshot of the endpoint's connection pool state (available/pending/leased concurrency) taken just before this request asks for a connection
- S3Client.getMaxActiveConnections() — new, the connection concurrency cap of the client.
- Updates submodules to latest versions.

Requires https://github.com/awslabs/aws-c-s3/pull/673 for newly bound client level API.

### DIFF
--- a/src/main/java/software/amazon/awssdk/crt/s3/S3Client.java
+++ b/src/main/java/software/amazon/awssdk/crt/s3/S3Client.java
@@ -262,6 +262,15 @@ public class S3Client extends CrtResource {
         return shutdownComplete;
     }
 
+    /**
+     * @return the maximum number of connections this client will establish to a single endpoint. Fixed for
+     * the lifetime of the client - derived from its configured throughput target, or overridden directly via
+     * {@link S3ClientOptions#withMaxConnections}.
+     */
+    public int getMaxActiveConnections() {
+        return s3ClientGetMaxActiveConnections(getNativeHandle());
+    }
+
     /*******************************************************************************
      * native methods
      ******************************************************************************/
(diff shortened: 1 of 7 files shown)


DmitriyMusatkin:
description of changes? title of pr does not seem to match with whats in the diff?

src/main/java/software/amazon/awssdk/crt/s3/S3Client.java
```diff
@@ -262,6 +262,15 @@ public CompletableFuture<Void> getShutdownCompleteFuture() {
         return shutdownComplete;
     }
 
+    /**
+     * @return the maximum number of connections this client will establish to a single endpoint. Fixed for
```
TingDaoK: >the maximum number of connections this client will establish to a single endpoint

? why it's the max per single endpoint?

azkrishpy: ah, good catch! I seem to have confused the connection manager's max connections with client's max active connections. fixed now.

TingDaoK (approved):

DmitriyMusatkin (approved):

